PR #3, opened 2019-10-22
Add support for GreenWaves-Technologies GAP8 RI5CY core

sivasivarajah:
GreenWaves-Technologies GAP8 port of FreeRTOS kernel.

### DIFF
--- a/portable/GCC/RI5CY-GAP8/chip_specific_extensions/gap8/freertos_risc_v_chip_specific_extensions.h
+++ b/portable/GCC/RI5CY-GAP8/chip_specific_extensions/gap8/freertos_risc_v_chip_specific_extensions.h
@@ -1,0 +1,67 @@
+/*
+ * FreeRTOS Kernel V10.2.1
+ * Copyright (C) 2019 Amazon.com, Inc. or its affiliates.  All Rights Reserved.
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy of
+ * this software and associated documentation files (the "Software"), to deal in
+ * the Software without restriction, including without limitation the rights to
+ * use, copy, modify, merge, publish, distribute, sublicense, and/or sell copies of
+ * the Software, and to permit persons to whom the Software is furnished to do so,
+ * subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in all
+ * copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY, FITNESS
+ * FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR
+ * COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER
+ * IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN
+ * CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+ *
+ * http://www.FreeRTOS.org
+ * http://aws.amazon.com/freertos
+ *
+ * 1 tab == 4 spaces!
+ */
+
+#if !defined(ASSEMBLY_LANGUAGE)
+/* Control and Status Registers reset values. */
+/* Machine mode with IRQ disabled, after MRET stay in MM. */
+#define portINITIAL_MSTATUS ( 0x1880 )
+/* Retrieve MTVEC address from linker script. */
+extern uint8_t __irq_vector_base_m__;
+#define portINITIAL_MEPC    ( &__irq_vector_base_m__ )
+#define portINITIAL_MCAUSE  ( 0x00000000 )
+#endif  /* ASSEMBLY_LANGUAGE */
+
+/* Size of a word, in bytes. */
+#define portWORD_SIZE                    ( 4 )
+/* Number of chip specific additional extensions. */
+#define portGAP8_ADDITIONAL_EXTENSIONS   ( 6 )
+
+/* a0 - a7 */
+#define portGAP8_ARGS_REGS               ( 8 )
+/* t0 - t6 */
+#define portGAP8_TEMP_REGS               ( 7 )
+/* s0 - s11 */
+#define portGAP8_SAVE_REGS               ( 12 )
+/* Minimal context size to save for irq handler. */
+/* a0-a7 + t0-t6 + ra */
+#define portGAP8_MINIMAL_CONTEXT_SIZE    ( portGAP8_ARGS_REGS + portGAP8_TEMP_REGS + 1 )
+/* General context size. */
+/* ra + a0-a7 + t0-t6 + s0-s11 + mstatus + mepc */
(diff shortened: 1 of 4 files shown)


RichardBarry:
Hi - thanks for your contribution.

Our intent with RISC-V ports is to have a single implementation that works on all processors, as far as possible, with the freertos_risc_v_chip_specific_extensions.h header file being used to enable chip specific registers to be handled by the port.  Currently your GAP8 port is a separate implementation so I would like to learn what we need to to to our upstream port to make it suitable for your hardware too - otherwise it will be very difficult to keep feature and testing parity across different RISC-V chips.

The main line code is already running on a RI5CY core so the first thing to do is determine the differences between that port and the needs of the GAP8.  Then to look at the features in your port to see if you have introduced functionality that doesn't exist in the upstream port so we can close any gaps that may exist.

In the upstream currently:
  + The existing supported RI5CY core has six additional registers - two sets of three loop optimisation registers.
  + The following link contains a description of the existing upstream port https://www.freertos.org/Using-FreeRTOS-on-RISC-V.html

Given those two bullets - any additional information you can provide will be greatly appreciated.  In the mean time I will also look further into the code you posted.

sivasivarajah:
Hi,
When FreeRTOS was released with a a RISC-V port, I took a look and adapted our code to follow RISC-V architecture organization. But the port implementation previously done has not changed.

The existing RISC-V port and ours differ in stack initialization, which is problematic.
As you said, the two chips are very similar, I can try to adjust the drivers and libs to reuse the existing port. In the end we end up with only one RISC-V implementation.

sivasivarajah:
I have tried to run the port on our board, I changed few things here and there, but could not make it work.
There are few points I would like talk about.

- First is about the function to pop first stack, _xPortStartFirstTask_:
```c
	load_x  sp, pxCurrentTCB			/* Load pxCurrentTCB. */
	load_x  sp, 0( sp )				 	/* Read sp from first TCB member. */

	load_x  x1, 0( sp ) /* Note for starting the scheduler the exception return address is used as the function return address. */

	portasmRESTORE_ADDITIONAL_REGISTERS	/* Defined in freertos_risc_v_chip_specific_extensions.h to restore any registers unique to the RISC-V implementation. */

	load_x  t0, 29 * portWORD_SIZE( sp )	/* mstatus */
	csrrw  x0, mstatus, t0					/* Interrupts enabled from here! */

	load_x  x5, 2 * portWORD_SIZE( sp )		/* t0 */
	load_x  x6, 3 * portWORD_SIZE( sp )		/* t1 */
	load_x  x7, 4 * portWORD_SIZE( sp )		/* t2 */
	load_x  x8, 5 * portWORD_SIZE( sp )		/* s0/fp */
	load_x  x9, 6 * portWORD_SIZE( sp )		/* s1 */
	load_x  x10, 7 * portWORD_SIZE( sp )	/* a0 */
	load_x  x11, 8 * portWORD_SIZE( sp )	/* a1 */
	load_x  x12, 9 * portWORD_SIZE( sp )	/* a2 */
	load_x  x13, 10 * portWORD_SIZE( sp )	/* a3 */
	load_x  x14, 11 * portWORD_SIZE( sp )	/* a4 */
	load_x  x15, 12 * portWORD_SIZE( sp )	/* a5 */
	load_x  x16, 13 * portWORD_SIZE( sp )	/* a6 */
	load_x  x17, 14 * portWORD_SIZE( sp )	/* a7 */
	load_x  x18, 15 * portWORD_SIZE( sp )	/* s2 */
	load_x  x19, 16 * portWORD_SIZE( sp )	/* s3 */
	load_x  x20, 17 * portWORD_SIZE( sp )	/* s4 */
	load_x  x21, 18 * portWORD_SIZE( sp )	/* s5 */
	load_x  x22, 19 * portWORD_SIZE( sp )	/* s6 */
	load_x  x23, 20 * portWORD_SIZE( sp )	/* s7 */
	load_x  x24, 21 * portWORD_SIZE( sp )	/* s8 */
	load_x  x25, 22 * portWORD_SIZE( sp )	/* s9 */
	load_x  x26, 23 * portWORD_SIZE( sp )	/* s10 */
	load_x  x27, 24 * portWORD_SIZE( sp )	/* s11 */
	load_x  x28, 25 * portWORD_SIZE( sp )	/* t3 */
	load_x  x29, 26 * portWORD_SIZE( sp )	/* t4 */
	load_x  x30, 27 * portWORD_SIZE( sp )	/* t5 */
	load_x  x31, 28 * portWORD_SIZE( sp )	/* t6 */
	addi	sp, sp, portCONTEXT_SIZE
	ret
```
The first task should start executing after its context is restored and a call to mret which will setup MSTATUS register and jump to task's function.
But here there does not seem to be any call to mret, and MEPC is not set either. I don't undestand how it could work, especially without IRQ enabled.

- There are also other suggestions.
Is it possible to have chip/boards specific macros? In our port, we have functions to set and clear interrupts, call to yield, and functions to enter and exit critical sections using a global variable for the counter instead of a single one per task, to save few bytes(not much but still).
An improvement could be to have the file **portmacro.h** have as many common macros as possible and make it include a file, say **freertos_risc_v_chip_macro_extensions.h** or whatever, where macros functions are defined.

- Last one concerns the way to save a task's stack. Isn't it easier to use ABI names when manipulating registers since they are more explicit(ra, t0, a0,...) and in this order?
```c
/*
 * Task's stack view.
 *     LOW
 * *************  <------ pxTopOfStack
 * *    ra     *
 * *    a0     *  <------ Parameters sent to task
 * *    a1     *
 * *    --     *
 * *    a7     *
 * *    t0     *
 * *    --     *
 * *    t6     *
 * *    s0     *
 * *    --     *
 * *    s11    *
 * *-----------*
 * *  MSTATUS  *  <------ Initial mstatus for the task
 * *   MEPC    *  <------ MEPC contains address of task's function, jump to it after mret.
 * *-----------*
 * *  HW loop  *
 * *===========*
 * * deedfeed  *
 * *************
 *    HIGH
 */
```
General purpose registers and CSR should be common to all RISC-V chips and above them we can save additional registers.

RichardBarry:
- the first code you pasted in has been updated slightly but only to ensure interrupts are enabled before the ret is called.  In the version you pasted interrupts did not become enabled until the first task to run entered the blocked state.  The change is very minor and at the time of writing can be seen on line 303 here: https://sourceforge.net/p/freertos/code/HEAD/tree/trunk/FreeRTOS/Source/portable/GCC/RISC-V/portASM.S#l303   It can be seen that interrupts are then enabled before the registers are popped from the stack, and it would be better to enable interrupts after the registers are popped from the the stack in case an interrupt comes in immediately (an immediate interrupt would not be an issue other than the additional stack space used).

- ret is used in place of mret to start the first as if xPortStartFirstTask() was a sub routine called by the first task, so returning from the called subroutine returns to the task that called it.  MSTATUS has already been configured by that point as per the referenced line 303 above.

- I don't think it is valid to have a single variable to keep track of critical section nesting depth if an environment call is used to yield when entering the blocked state.  The reason being that the environment call will not be prevented because you are inside a critical section so each task needs its own critical section nesting depth.  The internal implementation of the kernel copes with both scenarios - synchronous yields such as an environment call on a RISC-V, and asynchronous yields such as the setting of a PendSV bit on a Cortex-M - in the former case each task needs its own critical section nesting count, in the latter it doesn't as a yield will only actually occur when the critical section has exited (so the nesting count can only be zero at that point).

....not looked at the stack frame you posted yet to see if it differs from the one we use barring the use of the ABI names.  If the stack frames are different it is probably something can can be harmonised without breaking backward compatibility though.

yuhui-zheng:
@sivasivarajah  I believe we have other comm channel available on this topic, so that I'm closing this PR. Please reach out again, happy to chat.

PR #8, opened 2020-01-20
Synopsys ARC port, adding support for ARC EM and HS cores

YuguoWH:
Synopsys ARC port of FreeRTOS kernel.

### DIFF
--- a/portable/Synopsys/ARC/arc_freertos_exceptions.c
+++ b/portable/Synopsys/ARC/arc_freertos_exceptions.c
@@ -1,0 +1,52 @@
+/*
+ * FreeRTOS Kernel V10.2.1
+ * Copyright (C) 2019 Amazon.com, Inc. or its affiliates.  All Rights Reserved.
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy of
+ * this software and associated documentation files (the "Software"), to deal in
+ * the Software without restriction, including without limitation the rights to
+ * use, copy, modify, merge, publish, distribute, sublicense, and/or sell copies of
+ * the Software, and to permit persons to whom the Software is furnished to do so,
+ * subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in all
+ * copies or substantial portions of the Software.
+ *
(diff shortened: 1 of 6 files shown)


RichardBarry:
Thanks for your submission.  Are you from Synopsys?  I note there are subtle differences between this pull request and the code here: https://github.com/foss-for-synopsys-dwc-arc-processors/embarc_osp/releases including an update to the MIT license for the port layer hence the question.

YuguoWH:
Hi @RichardBarry.  Yes I'm from Synopsys. After the OSP release you mentioned, we had some modifications to the embARC code and put it in embARC BSP, which you could find here: https://github.com/foss-for-synopsys-dwc-arc-processors/embarc_bsp/tree/upstream
But basically not much has been changed in port layer. May I ask which license is required in FreeRTOS kernel?

RichardBarry:
Thanks for the confirmation.  FreeRTOS is now MIT licensed - I note your website still has the old license.   You can of course license your port bits anyway you like, but we strongly encourage MIT to make the port layers license compatible with the rest of FreeRTOS, and to make it as easy for users to use in production as possible.

I've checked the code out and will look through it to put together some comments (such as moving to the third party directory, and also publishing a demo, etc.).  Then we can work to incorporating in the release.

Currently SourceForge/SVN is still used as the source of truth, so merges will be done there.  This repo is synced from there.  After the next release the intention is to switch, so this git repo becomes the source of truth, and code gets synced back to SVN.

yuhui-zheng:
Hi @YuguoWH, 

It seems this PR has "allow edits from maintainers" disabled, so that I cannot update your PR directly. Instead, I'm proposing changes in this draft PR -- https://github.com/FreeRTOS/FreeRTOS-Kernel/pull/28. 

Changes:
- Rebase 
- Move ```./portable/Synopsys/ARC``` to ```./portable/ThirdParty/GCC/ARC_EM_HS```, since our folder structure as of now is ```./portable/<compiler>/<architecture>```. (The "GCC" is concluded from your Travis CI config -- https://travis-ci.org/github/foss-for-synopsys-dwc-arc-processors/embarc_osp/builds/659363653/config.)

There's no source code change. Please let me know your thoughts.

yuhui-zheng:
Moving further review to PR 28, so that I'm closing this PR.

PR #9, opened 2020-01-26
Added LoadStore Exception handlers

sachin0x18:
Handles `LoadStoreErrorCause` and `LoadStoreAlignmentCause` allowing to use
32-bit memory region (IRAM) as 8-bit or 16-bit memory region

### DIFF
--- a/portable/ThirdParty/GCC/Xtensa_ESP32/xtensa_loadstore_handler.S
+++ b/portable/ThirdParty/GCC/Xtensa_ESP32/xtensa_loadstore_handler.S
@@ -1,0 +1,559 @@
+/*
+ Copyright 2019 Espressif Systems (Shanghai) PTE LTD
+
+ Licensed under the Apache License, Version 2.0 (the "License");
+ you may not use this file except in compliance with the License.
+ You may obtain a copy of the License at
+
+     http://www.apache.org/licenses/LICENSE-2.0
+
+ Unless required by applicable law or agreed to in writing, software
+ distributed under the License is distributed on an "AS IS" BASIS,
+ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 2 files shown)


RichardBarry:
Thanks for this - are you from Espressif?

sachin0x18:
Yes

yuhui-zheng:
I will shortly close this PR and merge the exact change into our kernel upstream. The change will be included in the coming release.

yuhui-zheng:
The changes could be reviewed here -- https://sourceforge.net/p/freertos/code/2816/. We are still in the middle of the migration, should be able to merge PR normally afterwards.

mahavirj:
@yuhui-zheng 

> I will shortly close this PR and merge the exact change into our kernel upstream. The change will be included in the coming release.

Thanks. We have dependency of this change on PR at https://github.com/aws/amazon-freertos/pull/1713. Do we need to wait for next minor/major FreeRTOS kernel release or we can immediately update submodule pointer (for FreeRTOS kernel) there?

> Minor question for @sachin0x18 -- what was the "historical" reason that this port is under /ThirdPaty?

No specific reason. From our (Espressif) side, when we had first submitted ESP32 port, we did not include in `ThirdParty` prefix. Probably we can change that now. 

CC @sachin0x18

yuhui-zheng:
@mahavirj 

> We have dependency of this change on PR at aws/amazon-freertos#1713. Do we need to wait for next minor/major FreeRTOS kernel release or we can immediately update submodule pointer (for FreeRTOS kernel) there?

Kernel team has prepared a release candidate (which has included this PR), and we will soon have a kernel release. After this release, everyone should be able to update their submodule pointer.

As for the ```ThirdParty``` directory, let me check with team and see what needs to be done to move Espressif port out of ```ThirdParty``` directory.

PR #15, opened 2020-02-19
projdefs.h - resolve TickType_t overflow

feilipu:
Where `TickType_t` is `uint16_t` the calculation overflows.
This proposal was used to resolve [`feilipu/Arduino_FreeRTOS_Library` #61](https://github.com/feilipu/Arduino_FreeRTOS_Library/issues/61)

### DIFF
--- a/include/projdefs.h
+++ b/include/projdefs.h
@@ -38,7 +38,7 @@ typedef void (*TaskFunction_t)( void * );
 overridden by a macro of the same name defined in FreeRTOSConfig.h in case the
 definition here is not suitable for your application. */
 #ifndef pdMS_TO_TICKS
-	#define pdMS_TO_TICKS( xTimeInMs ) ( ( TickType_t ) ( ( ( TickType_t ) ( xTimeInMs ) * ( TickType_t ) configTICK_RATE_HZ ) / ( TickType_t ) 1000 ) )
+	#define pdMS_TO_TICKS( xTimeInMs ) ( ( TickType_t ) ( ( ( xTimeInMs ) * (uint32_t) configTICK_RATE_HZ ) / 1000 ) )
 #endif
 
 #define pdFALSE			( ( BaseType_t ) 0 )


yuhui-zheng:
Hi @feilipu,

Thank you for submitting this PR. Yet, I cannot accept this PR, since it won't be the correct fix in all scenarios. 

One example, if we are on Cortex M4 port, depending on whether ```configUSE_16_BIT_TICKS``` is set, ```TickType_t``` can be either ```uint16_t``` or ```uint32_t```.  https://github.com/FreeRTOS/FreeRTOS-Kernel/blob/master/portable/GCC/ARM_CM4F/portmacro.h#L59-L69

And, another example for RISC-V, ```TickType_t``` can be either ```uint32_t``` or ```uint64_t```. https://github.com/FreeRTOS/FreeRTOS-Kernel/blob/master/portable/GCC/RISC-V/portmacro.h#L47-L58

When consuming FreeRTOS, users could provide their own implementation of ```pdMS_TO_TICKS( xTimeInMs )``` in ```FreeRTOSConfig.h``` (as commented here -- https://github.com/FreeRTOS/FreeRTOS-Kernel/blob/master/include/projdefs.h#L37-L39). On a specific port, users have slightly more options -- could divide first (if dividend is larger than divisor), could assert or #error if assumption is not met, could check overflow...

feilipu:
Hi @yuhui-zheng,

I'm not at all sensitive to how you fix this issue, but afaik it is not sensible to leave obviously broken code in your solution. You should think about how to make a solution that works generally.

The mainline code is broken if `TickType_t` is set to a 16 bit type. Please, work out a solution that works for this supported option.

RichardBarry:
There have been a few other discussions on this topic, and, as per the above, there isn't one solution that is optimised for all use cases - hence the code is written so if you provide your own implementation in FreeRTOSConfig.h then the default implementation is ignored.  The macro is really just a helper macro, the kernel doesn't use it itself, so you have the choice whether to use it or create any other method that suites your use case.

feilipu:
@RichardBarry noted thanks. 👍 

Assuming the ATmega architecture will remain using 16 bit ticks for the foreseeable future, I'll retain my solution as it stands. As you say it is only a helper macro.

cobusve:
I think the design and documentation here acknowledges that this implementation may not be ideal for all situations. It suggests  "This macro can be overridden by a macro of the same name defined in FreeRTOSConfig.h in case the overridden by a macro of the same name defined in FreeRTOSConfig.h in case the definition here is not suitable for your application.."

I would generally encourage you to follow this design and override the macro in FreeRTOSConfig.h as per the documentation for your implementation instead of updating the Kernel files as doing the latter will cause you a lot of future effort and unnecessary risks.

PR #19, opened 2020-02-26
Xtensa_ESP32: Change _iram_end to _iram_text_end

sachin0x18:
xtensa_loadstore_handler.S uses _iram_end to prevent modification of IRAM
code. With the LoadStore exception handler in place, IRAM can also be
used for .bss and .data section. Hence the sanity check should be based
upon _iram_text_end and not _iram_end


sachin0x18:
While supporting some AFR customers, we came across a use-case where it was required to place certain library's .bss and .data section in IRAM thus freeing up memory in DRAM. This lead to change _iram_end to _iram_text_end

yuhui-zheng:
Thank you for submitting this PR. We'll merge it here first, and the change will be reflected in AFR in the next release.

PR #21, opened 2020-03-05
Cortex-A9 port: Adding stack alignment directive to assembly code

VladimirUmek:
Hello,
Cortex-A9 port should have eabi_attribute TAG_ABI_Align_preserved set in order to enable smooth object linking between different toolchains.

I believe that [explanation written on this link](http://www.support.code-red-tech.com/CodeRedWiki/Preserve8) summaries the problem well enough. In most cases linker only generates a warning if this attribute is not present but there can also be link errors when using strict project settings.

Assembly code from portASM.S maintains alignment so this attribute is safe to add.

Vladimir

### DIFF
--- a/portable/GCC/ARM_CA9/portASM.S
+++ b/portable/GCC/ARM_CA9/portASM.S
@@ -24,7 +24,7 @@
  *
  * 1 tab == 4 spaces!
  */
-
+	.eabi_attribute Tag_ABI_align_preserved, 1
 	.text
 	.arm
 


RichardBarry:
This is great information - thanks.  I've not come across this before, but just looked it up in the GNU Assembler manual.  My only slight concern would be if it were not supported in all the GCC distributions - however we will test it in our office and assuming all is good merge into the mainline.

yuhui-zheng:
Thank you for this PR. I've checked spec and run a couple tests. Putting down information below, and will merge after a quick round of review.  

The most recent spec available -- [Addenda to, and Erranta in, the ABI for the ARM architecture ABI r2.10, version e](https://developer.arm.com/docs/ihi0045/e).

objdump elf files before and after, and the difference: 
```
// before
 136e80:	ba000c02 	blt	139e90 <__ARM.attributes_end+0x39e>
// after 
136e80:	ba000c02 	blt	139e90 <__ARM.attributes_end+0x39c>
```

Also tried attribute ```Tag_ABI_align8_preserved```, which does the same thing.

PR #22, opened 2020-03-12
update tasks.c

KiLuYa:
Is this a more efficient way to update "xNextTaskUnblockTime" in prvResetNextTaskUnblockTime( ) ?


RichardBarry:
Thanks for your suggestion - on first look it seems this could be a nice optimisation - we will test and report back.

PR #23, opened 2020-03-14
Update tasks.c

KiLuYa:
This is the same question as a PR I used to make.
In this case, we can call "portRESET_READY_PRIORITY" directly.


KiLuYa:
Thanks for your correction. I will reopen this PR.

yuhui-zheng:
I'm closing this PR, since we have PR #26 in place.

PR #25, opened 2020-03-14
Remove obsolete variable from Win32 port.

RichardBarry:
The Windows port layer is built with both MSVC and GCC.  GCC generated a warning relating to the variable lWaitForYield being set but not used.  This change removes the variable.


yuhui-zheng:
I'm closing this PR since the same commit has been merged in PR #27. 

(Commit ID 23a05b5817661ff99ca0834eb2269ee4ec82ae66 is in both this PR and PR #27. When creating that branch for PR #27, it's likely that you've branched from this branch instead of master. We are fine though.)

PR #38, opened 2020-03-27
Use CMSIS Intrinsic Functions and Core Register Access Functions

paperchalice:
Description
-----------

Just replace some functions same from CMSIS
Test Steps
-----------

Related Issue
-----------

### DIFF
--- a/portable/GCC/ARM_CM7/r0p1/port.c
+++ b/portable/GCC/ARM_CM7/r0p1/port.c
@@ -709,7 +709,7 @@ static void vPortEnableVFP( void )
 	uint8_t ucCurrentPriority;
 
 		/* Obtain the number of the currently executing interrupt. */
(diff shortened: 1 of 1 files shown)


yuhui-zheng:
@paperchalice Thank you for submitting this PR. 

We may not be able to accept this PR though, given:
- Nothing in the FreeRTOS port layers depend on third party libraries as that would require us to re-test everything each time the third party library was updated. (This PR introduces dependency on CMSIS, and may need to have device.h in include path.)
- We have many other ARM Cortex-M ports and this PR only covers one of them. This seems to break the consistency. And without obvious runtime benefit, I think it may not be worth the effort to go through all ports to make the same updates.  

I'm closing this PR for now. Please reopen if you believe we have other aspects not covered. Thanks again.

PR #40, opened 2020-03-28
[RFC] tasks.c: disable interrupts in critical section only if scheduler is started

mahavirj:
Description
-----------

Behaviour of `vTaskEnterCritical` and `vTaskExitCritical` was asymmetrical,
in the sense, if these APIs were invoked prior to scheduler start condition
then one would disable interrupts but other would not re-enable again.
This commit ensure that behaviour is symmetrical for pre-scheduler start
condition.

Test Steps
-----------

Ensured that prior to scheduler start situation, calling `vTaskEnterCritical` followed by `vTaskExitCritical` does not affect interrupt enable status.

Related Issue
-----------

### DIFF
--- a/tasks.c
+++ b/tasks.c
@@ -4291,10 +4291,10 @@ static void prvResetNextTaskUnblockTime( void )
 
 	void vTaskEnterCritical( void )
 	{
-		portDISABLE_INTERRUPTS();
-
 		if( xSchedulerRunning != pdFALSE )
 		{
+			portDISABLE_INTERRUPTS();
+
 			( pxCurrentTCB->uxCriticalNesting )++;
 
 			/* This is not the interrupt safe version of the enter critical


RichardBarry:
Thanks for this, but I think it needs to be an FAQ if it is not already, as this is intended behaviour.  Even if it were not, changing the behaviour in one place would also require changing all the ports (and there are a LOT of them) that have port specific implementations of this function.

Explanation:  Many years ago many bug reports were tracked down to interrupt service routines attempting to use the FreeRTOS API, or perform a context switch, before the scheduler had started.  These were bug reports that were, themselves, very hard to debug but turned out to be application, rather than RTOS, issues.  To help users avoid these tricky to debug situations we made a decision to keep interrupts disabled until the scheduler is started.  So from the time FreeRTOS API functions are first called to the time the first task starts running interrupts are intentionally kept masked.  As soon as the first task is running, they are unmasked.

PR #48, opened 2020-04-07
ATmegaxxxx - generalised support

feilipu:
<h2>ATmegaxxxx</h2>

__Port for generalised Microchip ATmega architecture__

<h3>Description</h3>

This port provides a basis for supporting all modern ATmega devices using either the Enhanced Watchdog Timer, or Timer0 (an 8-bit Timer generally available across the whole range).

This initial commit contains the information required to build with System Tick being generated by either the:
- Watchdog Timer, or
- Timer0 - an 8-bit Timer

Further commits can add support for 16-bit Timers available on many relevant devices. The availability of these 16-bit Timers is somewhat device specific, and these complex and highly configurable Timers are often used to generate phase correct PWM timing (for example) and they would be wasted as a simple System Tick.

The port also provides support for the 3 byte program counter devices __ATmega2560__ and __ATmega2561__. Specific to these two devices the `RAMPZ` and `EIND` registers need to be preserved during a context switch. Also, due to a limitation in GCC, the scheduler needs to reside in the lower 128kB of flash for both of these devices. This is achieved by adding the `.lowtext` section attribute to the function prototype.

The ATmega808/809/1608/1609/3208/3209/4808/4809 megaAVR family is a substantial revision of the ATmega platform, and is currently not supported by avr-libc. Therefore the __megaAVR 0-series__ is not supported by this port.


feilipu:
To build generic Microchip (AVR) ATmega support the similarities across the family must be considered, and differences respected. Some comments on the strategy.

<h3>System Tick</h3>

The Microchip (AVR) ATmega family has limited Timer and Pin capabilities, and is designed to be used in physical applications, controlling hardware with PWM and recognising level and edge voltage changes. It does this mainly through the use of 16 bit Timers (for generating phase correct PWM by up/down counting), and Pins attached to Interrupts. The 8 bit Timers are also attached to Pins, and they can be used for more simple timing tasks, requiring only a single counting direction.

The Timers not attached to Pins (and therefore not impacting the application of the device) are (occasionally, device dependent) an additional 16 bit Timer, and the Watch Dog Timer.

The Watch Dog Timer is configured identically across the entire range of ATmega devices. It comes in two variants. 1. Old style (eg ATmega32) which does not have an Interrupt capability, and hence on these old devices cannot be used as the System Tick. and 2. New style, which can generate an Interrupt, and is available on every relevant device.

Using the Watch Dog Timer (WDT) to generate the System Tick does not impact its use as a watch dog. It can be configured to generate a System Tick interrupt, and then one period later to Reset the device if the interrupt is not serviced.

Configuration and usage of the WDT is covered in `<avr/wdt.h>` which was revised in avr-libc 2.0.0.

Two additional WDT functions are provided in `port.c`, which extend avr-libc functions to enable the WDT Interrupt without enabling Reset `wdt_interrupt_enable()`, and to enable both the Interrupt and the Reset `wdt_interrupt_reset_enable()`.

<h3>3 Byte PC Devices</h3>

The ATtiny, ATmega, ATxmega families can optionally support both 3 byte PC and 3 byte RAM addresses. However, focusing on just the ATmega family only two devices have a large Flash requiring them to use 3 byte PC. These are the __ATmega2560__ and __ATmega2561__. This PR provides support for these two devices in two ways.

 - providing `portSAVE_CONTEXT()` and `portRESTORE_CONTEXT` saving both the __RAMPZ__ and __EIND__ registers.
 - providing a `portTASK_FUNCTION_PROTO()` with the linker attribute `.lowtext` which is used to ensure that the scheduler and relevant functions remain in the lower 128kB of Flash.

<h3>Interrupt Nesting</h3>

The ATmega family does not support interrupt nesting, having only one interrupt priority. This means that when the Scheduler is running, interrupts are normally disabled.

When a very time critical process is running, based on microsecond timing generated by one of the Timers, it is important to re-enable interrupts as early as possible in processing a Yield. Fortunately, this is supported through the use of the `NO_BLOCK` decorator when defining the Interrupt Service Routine.

The `NO_BLOCK` decorator will enable the global interrupt early in the handling of an ISR (in this case for the Scheduler), and enable interrupts to be nested. Using this method, I've been able to successfully an [Audio Synthesiser](https://feilipu.me/2015/06/02/goldilocks-analogue-synthesizer/) with less than 83 microseconds for each cycle.

Using `NO_BLOCK` is optional, and should only be done if a critical Timer should interrupt the Scheduler.

<h3>Extension proposals for PR</h3>

From the basis of this PR, following further commits could be made. But, before doing this work I'd like confirmation to proceed.

 - Identify which "in production" ATmega devices are "old style"  devices without WDT Interrupt capability, and configure them to automatically use the 8 bit Timer 0 as the System Tick.

 - Add option to use a number of alternate 16 bit Timers with further configuration. As already noted, these 16 bit Timers are not configured consistently across all devices (or even across different Timers within one device) so providing them as a generalised solution is challenging.

Please comment and provide direction as needed.

feilipu:
<h3>Heap Management</h3>

Most users of FreeRTOS will choose to manage their own heap using one of the pre-allocated heap management algorithms, but for those that choose to use `heap_3.c`, the wrappered `malloc()` method, there is an issue that needs to be addressed.

The avr-libc library assumes that the stack will always be above the heap, and does a check for this when responding to a `malloc()` request, if it has not otherwise been configured. This is not the case when Tasks are running, as their stack is located in the early allocated heap address ranges which will be below free heap memory.

To avoid this issue causing `pvPort_Malloc()` from failing, the user needs to issue this tuning statement BEFORE they use the heap.

```
__malloc_heap_end = (char *)(RAMEND - __malloc_margin);
```
Unfortunately in the repository there is nowhere sensible to include this statement.

feilipu:
ATmega devices without enhanced __WDT__ Interrupt capability
 - ATmega8A/16A/32A/64A/128A -> 4kB RAM
 - ATmega808/809/1608/1609/3208/3209/4808/4809 -> 6kB RAM
 - ATmega165A/165PA/325A/325PA/3250A/3250PA/645A/645P/6450A/6450P -> 4kB RAM
 - ATmega169A/169PA/329A/329PA/3290A/3290PA/649A/649P/6490A/6490P -> 4kB RAM

ATmega devices with __ENHANCED WDT__ Interrupt capability
 - ATmega8U2/16U2/32U2 -> 2kB RAM
 - ATmega16U4/32U4 - Arduino Leonardo -> 2.5kB RAM
 - ATmega48PB/88PB/168PB/328PB - Arduino Uno -> 2kB RAM
 - ATmega164PA/324PA/644PA/1284P - Goldilocks -> __16kB RAM__
 - ATmega324PB -> 2kB RAM
 - ATmega640/1280/2560/1281/2561 - Arduino Mega -> __8kB RAM + XRAM__

portable/GCC/ATmegaxxxx/port.c
```diff
@@ -0,0 +1,741 @@
...
+ * so we need not worry about reading/writing to the stack pointer.
+ */
+#if defined(__AVR_ATmega2560__) || defined(__AVR_ATmega2561__)
+/* 3-Byte PC Save */
+#define portSAVE_CONTEXT()                                                              \
+        __asm__ __volatile__ (  "push   __tmp_reg__                             \n\t"   \
+                                "in     __tmp_reg__, __SREG__                   \n\t"   \
+                                "cli                                            \n\t"   \
+                                "push   __tmp_reg__                             \n\t"   \
+                                "in     __tmp_reg__, 0x3B                       \n\t"   \
```
yuhui-zheng, line 234: Just to confirm -- here we are storing RAMPZ register (0x3B) and EIND register (0x3C) in thread context. And this is to support large flash access per thread, would this be correct?

portable/GCC/ATmegaxxxx/portmacro.h
```diff
@@ -0,0 +1,139 @@
...
+ *
+ * The settings in this file configure FreeRTOS correctly for the
+ * given hardware and compiler.
+ *
+ * These settings should not be altered.
+ *-----------------------------------------------------------
+ */
+
+/* Type definitions. */
+
```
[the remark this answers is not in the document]
feilipu: Will add these back in with a separate commit.
Personally, I delete old stuff because it bites me.
But I understand the need for legacy support.

Will not add the `portSTACK_TYPE` as suggested.

portable/GCC/ATmegaxxxx/port.c
```diff
@@ -0,0 +1,741 @@
...
+
+/* Start tasks with interrupts enabled. */
+#define portFLAGS_INT_ENABLED                ( (StackType_t) 0x80 )
+
+#if defined( portUSE_WDTO)
+    #warning "Watchdog Timer used for scheduler."
+    #define    portSCHEDULER_ISR        WDT_vect
+
+#elif defined( portUSE_TIMER0 )
+/* Hardware constants for Timer0. */
```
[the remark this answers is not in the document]
feilipu: If you would like to add 16-bit Timer capability into this PR, I can do that too.

But, I'm not totally sure how to do it generally for all ATmega devices.
My experience is limited to these devices.

 - ATmega8U2/16U2/32U2 -> 2kB RAM
 - ATmega16U4/32U4 - Arduino Leonardo -> 2.5kB RAM
 - ATmega48PB/88PB/168PB/328PB - Arduino Uno -> 2kB RAM
 - ATmega164PA/324PA/644PA/1284P - Goldilocks -> __16kB RAM__
 - ATmega640/1280/2560/1281/2561 - Arduino Mega -> __8kB RAM + XRAM__

But, let's get the WDT and 8-bit Timer0 done first.

portable/GCC/ATmegaxxxx/port.c
feilipu: Yes, we're storing them because any function may be running when the `Yield()` is run by the ISR. But the Scheduler has to be in the lower 128kB of flash. More on the `.lowtext` discussion above.

portable/GCC/ATmegaxxxx/portmacro.h
```diff
@@ -0,0 +1,139 @@
...
+/* Kernel utilities. */
+extern void vPortYield( void )          __attribute__ ( ( naked ) );
+#define portYIELD()                     vPortYield()
+/*-----------------------------------------------------------*/
+
+#if defined(__AVR_ATmega2560__) || defined(__AVR_ATmega2561__)
+/* Task function macros as described on the FreeRTOS.org WEB site. */
+// Add .lowtext tag for the linker for ATmega2560 and ATmega2561. To make sure they are loaded in low memory.
+#define portTASK_FUNCTION_PROTO( vFunction, pvParameters ) void vFunction( void *pvParameters ) __attribute__ ((section (".lowtext")))
```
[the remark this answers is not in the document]
feilipu: The original conversation was in 2008, by [stu_san on AVR Freaks](https://www.avrfreaks.net/forum/tutsoftcodefreertos-atmega25601). I won't paraphrase the details here.

My [comment from 2012](https://www.avrfreaks.net/comment/779006#comment-779006) showed I addressed the issue by adding a new section into the `avr6.x` linker called `.task`. And more discussion here on [my post from 2011](https://feilipu.me/2011/09/22/freertos-and-libraries-for-avr-atmega/).

From avrlibc 1.8.0 (January 2014) there was a `.lowtext` section that was implemented in `avr6.x`, so I moved my repository to use that new section name, which also made modifying the default linker script unnecessary.

portable/GCC/ATmegaxxxx/port.c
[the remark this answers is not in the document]
yuhui-zheng: I agree. And I think 8-bit is sufficient for tick clock. Will check other's opinion on this.

portable/GCC/ATmegaxxxx/portmacro.h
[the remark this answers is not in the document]
yuhui-zheng: ```portBASE_TYPE``` got changed to ```BaseType_t``` in v8.x.x.
https://www.freertos.org/upgrading-to-FreeRTOS-V8.html

I could add some words on top of type definitions, suggesting to not use legacy types. Will double check other folks' opinion.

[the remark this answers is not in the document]
RichardBarry: Agree best to keep them in.  They are not used by the kernel any more (not for a long time) but there are old demos that will not build if they are removed.

portable/GCC/ATmegaxxxx/portmacro.h
```diff
@@ -0,0 +1,149 @@
...
+#define portBYTE_ALIGNMENT          1
+#define portNOP()                   __asm__ __volatile__ ( "nop" );
+/*-----------------------------------------------------------*/
+
+/* Kernel utilities. */
+extern void vPortYield( void )      __attribute__ ( ( naked ) );
+#define portYIELD()                 vPortYield()
+/*-----------------------------------------------------------*/
+
+#if defined(__AVR_ATmega2560__) || defined(__AVR_ATmega2561__)
```
RichardBarry, line 134: I presume these parts are singled out (doubled out?) because they have larger memory space.  In general preference is not to have chip (as opposed to architecture) specifics in a port layer, especially if it puts a burden on the user to get the compile time options right, or the target chip family is likely to change in the future.  So a couple of questions - is there a catch-all definition that can be use for all devices for which this special use case applies, or alternatively, can we just have this special use case apply to all chips no matter if they need it or not?    [either way, the code as you have it now is better than the legacy code, which was very chip family specific.]

Is .lowtext something that always exists, or something the user will potentially trip up on if they don't have the correct linker script.  If the latter is true then please update the comment to tell users how to fix any potential linker errors.  (can just be a link to a forum post somewhere that tells people how to fix).

portable/GCC/ATmegaxxxx/port.c
```diff
@@ -0,0 +1,747 @@
...
+
+/*-----------------------------------------------------------*/
+
+/* We require the address of the pxCurrentTCB variable, but don't want to know
+any details of its type. */
+typedef void TCB_t;
+extern volatile TCB_t * volatile pxCurrentTCB;
+
+/* actual number of ticks per second, after configuration. */
+TickType_t portTickRateHz;
```
RichardBarry, line 71: Is this needed, or can it be replaced by a const?  If it is needed, can it be made static?

portable/GCC/ATmegaxxxx/port.c
```diff
@@ -0,0 +1,747 @@
...
+
+/* remaining ticks in each second, decremented to enable the system_tick. */
+volatile TickType_t ticksRemainingInSec;
```
RichardBarry, line 74: Can this be static?

portable/GCC/ATmegaxxxx/port.c
```diff
@@ -0,0 +1,747 @@
...
+/*-----------------------------------------------------------*/
+
+#if defined(portUSE_WDTO)
+/*
+ * Setup WDT to generate a tick interrupt.
+ */
+void prvSetupTimerInterrupt( void )
+{
+    //reset watchdog
+    wdt_reset();
```
RichardBarry, line 654: Is this function guaranteed to be available?

portable/GCC/ATmegaxxxx/portmacro.h
```diff
@@ -0,0 +1,149 @@
...
+
+
+#define portDISABLE_INTERRUPTS()    __asm__ __volatile__ ( "cli" ::: "memory")
+#define portENABLE_INTERRUPTS()     __asm__ __volatile__ ( "sei" ::: "memory")
+
+/*-----------------------------------------------------------*/
+
+/* Architecture specifics. */
+
+// System Tick  - Scheduler timer
```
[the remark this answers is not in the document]
feilipu: Done.

portable/GCC/ATmegaxxxx/portmacro.h
```diff
@@ -0,0 +1,149 @@
...
+// Prefer to use the enhanced Watchdog Timer, but also Timer0 is ok.
+
+#if defined(WDIE) && defined(WDIF)              // Enhanced WDT with interrupt capability
```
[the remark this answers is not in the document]
feilipu: To 1. that is exactly the purpose of this path. The Enhanced Watchdog Timer (WDT) is available on all commonly used ATmega devices. It is configured simply and can be detected by the availability of the Watchdog Interrupt Enable (WDIE), and Watchdog Interrupt Flag (WDIF) provided by avr-libc. These flags are also configured consistently across all MCUs.

To 2. it becomes more challenging, as the numbering for the 16-bit Timers changes from MCU to MCU, as does their configuration and availability. It would be possible to provide a configuration for the ATmega32 (a legacy product), but this wouldn't be applicable to modern ATmega MCU. I've built a matrix of Timers vs MCU for a number of the common MCU types, but it is relatively complex and, as @yuhui-zheng pointed out, reserved bits are masked differently depending on the MCU too.

Happy to add 16-bit Timers as an example, but probably best to be left to the user to do final configuration.

To 3. I'd seek advice on how to best achieve that outcome. Perhaps reviewer edits?

portable/GCC/ATmegaxxxx/portmacro.h
feilipu: The two parts __ATmega2560__ and __ATmega2561__ are the only two parts in common usage (and for the past 12 years at least) that need the 3 Byte PC solution.

Potentially they could also be differentiated by architecture by using the [these options](https://gcc.gnu.org/onlinedocs/gcc/AVR-Options.html) and providing the below.

```
#if defined(__AVR_3_BYTE_PC__)
```
This is more generic, so I'll fix it.

The `.lowtext` section is provided by the linker script in `avr6.x`. If they have correctly nominated their device type in the compile line `-mmcu=atmegaxxxx`, then the correct linker script will be selected. It is automatic.

portable/GCC/ATmegaxxxx/port.c
feilipu: This is left over from my repo. Will be deleted.

portable/GCC/ATmegaxxxx/port.c
feilipu: Also, left over. Can be deleted.

portable/GCC/ATmegaxxxx/port.c
feilipu: The `wdt_reset()` function is [provided by `avr/wdt.h`](https://www.nongnu.org/avr-libc/user-manual/group__avr__watchdog.html). It is always available.

portable/GCC/ATmegaxxxx/portmacro.h
feilipu: There is an additional twist to this story in that one device has both 256kB of Flash and needs `__AVR_3_BYTE_PC__` and can address XRAM beyond 64kB and so has the `__AVR_HAVE_RAMPZ__` register. This is the ATmega2560.

To cater for the ATmega2561, which has a 3 Byte PC but doesn't have XRAM capability, another option for `portSAVE_CONTEXT()` needs to be provided.

__EDIT__ done.

portable/GCC/ATmegaxxxx/readme.md
```diff
@@ -0,0 +1,84 @@
...
+ - ATmega16U4/32U4 - Arduino Leonardo -> 2.5kB RAM
+ - ATmega48PB/88PB/168PB/328PB - Arduino Uno -> 2kB RAM
+ - ATmega164PA/324PA/644PA/1284P - Goldilocks -> __16kB RAM__
+ - ATmega324PB -> 2kB RAM
+ - ATmega640/1280/2560/1281/2561 - Arduino Mega -> __8kB RAM + XRAM__
+
+ATmega devices without enhanced __WDT__ Interrupt capability - will use Timer0.
+
+ - ATmega8A/16A/32A/64A/128A -> 4kB RAM
+ - ATmega808/809/1608/1609/3208/3209/4808/4809 -> 6kB RAM
```
TintuVayalattu, line 82: ATmega808/809/1608/1609/3208/3209/4808/4809 series is comparatively new. 
It doesn't have a dedicated 8 bit timer Timer0. It has only 16 bit timers named TCA and TCB. These can be used in split-mode to create 8 bit timers if required. Also, TCA and TCB have completely different set of registers and bitmaps than that of legacy mega devices such as mega48PB/mega328PB. I think, with Timer0, this PR can't serve ATmega808/809/1608/1609/3208/3209/4808/4809. What do you think?

feilipu: Grrr. Yes. It is a totally different beast.

Looking at the data sheet the best option seems to use the RTC with the internal 32kHz Oscillator in the Periodic Interrupt Timer (PIT) mode as the System Tick.

_Using the same clock source as the RTC function, the PIT can request an interrupt or trigger an output event on every nth clock period. n can be selected from {4, 8, 16,.. 32768} for interrupts, and from {64, 128, 256,... 8192} for events._

The ATmega808/809/1608/1609/3208/3209/4808/4809 WDT doesn't have any interrupt capability, and I'd picked up that it couldn't be used. But I hadn't noticed the completely reimagined Timer solution.

I don't have any of ATmega808/809/1608/1609/3208/3209/4808/4809, so I'd suggest to work this out later.

feilipu: Further, I'm quite sure that avr-libc doesn't support this series of devices, as [2.0.0 was released in 2016](https://savannah.nongnu.org/forum/forum.php?forum_id=8460). Without avr-libc, everything here is irrelevant.

The only option is to make a note that the ATmega808/809/1608/1609/3208/3209/4808/4809 series is unsupported.

TintuVayalattu: I think, megaAVR-0 series is of architecture **avr8x** - which is the same as xmega and few tiny devices. This is supported by avr-libc since the architecture was already existing.
I have one 4808. May be I can try after this got merged.

feilipu: Ok. That makes sense. I've not seen that avr8.x architecture before.

Then it would be best to handle support as part of an __ATxmega__ port.

Something to add later, or separately.

yuhui-zheng: Another solution I could think of is probably making ```prvSetupTimerInterrupt()``` in port.c external. We could provide example implementations in various projects, e.g. 8-bit/16-bit timer (as default probably), watchdog as a second clock source (for security initiatives), and RTC (for use cases concerning about clock skew).  Users could determine which one(s) to use in their application scenario. 

This may make sense, since all the timers we are talking about here are not part of the MCU core but rather peripheral. (In contrast to ARM, SysTick is provided in MCU core.) This will also make port cleaner...

yuhui-zheng:
A standalone comment -- I setup a project which runs with Atmel Studio simulator. https://github.com/yuhui-zheng/FreeRTOS/tree/atmega_port_demo/FreeRTOS/Demo/AVR_ATMega_AtmelStudio_simulator 

I'll use this one to verify build at least. And if you are interested, could play with it as well. ```AVR_ATMega_AtmelStudio_simulator/README.md``` has included some details regarding to setup. (The project is configured to use ATmega2560, one 8-bit timer, one 16-bit timer, one WDT, and etc. And if to play with RTC, could reconfigure the project. But you'll notice, RTC is not so common in ATmega family.)

portable/GCC/ATmegaxxxx/portmacro.h
```diff
@@ -0,0 +1,153 @@
...
+
+/* Architecture specifics. */
+
+/* System Tick  - Scheduler timer
+ * Prefer to use the enhanced Watchdog Timer, but also Timer0 is ok.
+ */
+
+#if defined(WDIE) && defined(WDIF)              /* If Enhanced WDT with interrupt capability is available */
+
+#define portUSE_WDTO                WDTO_15MS   /* use the Watchdog Timer for xTaskIncrementTick */
```
yuhui-zheng, line 95: I think we are missing ```#include <avr/wdt.h>``` in this file. 

In the simulator project I linked, WDTO_15MS is not recognized without ```#include <avr/wdt.h>```

portable/GCC/ATmegaxxxx/portmacro.h
```diff
@@ -0,0 +1,153 @@
...
+#define portENABLE_INTERRUPTS()     __asm__ __volatile__ ( "sei" ::: "memory")
+/*-----------------------------------------------------------*/
+
+/* Architecture specifics. */
+
+/* System Tick  - Scheduler timer
+ * Prefer to use the enhanced Watchdog Timer, but also Timer0 is ok.
+ */
+
+#if defined(WDIE) && defined(WDIF)              /* If Enhanced WDT with interrupt capability is available */
```
yuhui-zheng, line 93: Proposing to use counter as default clock source. It's more aligned with other ports.

portable/GCC/ATmegaxxxx/portmacro.h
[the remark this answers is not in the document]
feilipu: > I think we are missing `#include <avr/wdt.h>` in `portmacro.h` file.

Agreed. I usually hold all the timer configurations in a separate file called [`FreeRTOSBoardDefs.h`](https://github.com/feilipu/avrfreertos/blob/master/freeRTOS10xx/include/FreeRTOSBoardDefs.h), or [`FreeRTOSVariant.h`](https://github.com/feilipu/Arduino_FreeRTOS_Library/blob/master/src/FreeRTOSVariant.h).

This relates to your question about using a separate layer, for individual timer configuration, and in practice that is what I do.

portable/GCC/ATmegaxxxx/portmacro.h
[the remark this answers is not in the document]
feilipu: Using the TimerX for the tick is normal practice. But the issue is that every MCU has different timers available, and each of these have different reserved bits. It gets very complex.

My personal preference is to use the RTC related timer (usually Timer2), which ensures that there no drift. But this relies on having an external 32kHz watch crystal implemented. Not a general solution.

The advantage of the WDT is simple and repeatable configuration across all the common "hobbyist" MCU versions (Arduino).

portable/GCC/ATmegaxxxx/port.c
```diff
@@ -0,0 +1,822 @@
...
+    *pxTopOfStack = ( StackType_t ) ( usAddress & ( uint16_t ) 0x00ff );
+    pxTopOfStack--;
+
+#if defined(__AVR_3_BYTE_PC__)
+    /* The AVR ATmega2560/ATmega2561 have 256KBytes of program memory and a 17-bit
+     * program counter. When a code address is stored on the stack, it takes 3 bytes
+     * instead of 2 for the other ATmega* chips.
+     *
+     * Store 0 as the top byte since we force all task routines to the bottom 128K
+     * of flash. We do this by using the .lowtext label in the linker script.
```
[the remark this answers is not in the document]
feilipu: The linker script is provided by [binutils-avr](https://ubuntu.pkgs.org/20.04/ubuntu-universe-amd64/binutils-avr_2.26.20160125+Atmel3.6.2-1_amd64.deb.html) used with gcc. This PR is based on solely using the Open Source tools available for AVR. Nothing to do with Atmel Studio (although it would be doing the same thing under the covers, since it uses avr-libc too).

Look for `/usr/lib/avr/lib/ldscripts/avr6.x` in the [`binutils-avr`](https://ubuntu.pkgs.org/20.04/ubuntu-universe-amd64/binutils-avr_2.26.20160125+Atmel3.6.2-1_amd64.deb.html) package.

portable/GCC/ATmegaxxxx/readme.md
feilipu: Agreed. As noted in another comment, putting the TimerN configuration into a separate file is what I do in practice. I usually hold all the timer configurations in a separate file called [`FreeRTOSBoardDefs.h`](https://github.com/feilipu/avrfreertos/blob/master/freeRTOS10xx/include/FreeRTOSBoardDefs.h), or [`FreeRTOSVariant.h`](https://github.com/feilipu/Arduino_FreeRTOS_Library/blob/master/src/FreeRTOSVariant.h) for ArduinoFreeRTOS.

But, this is unrelated to the OP comment. Support for this new ATmega808/809/1608/1609/3208/3209/4808/4809 `avr8x` family is going to be difficult as the tool set required is not yet fully developed. Something that can be looked at as a future addition, imho.

TintuVayalattu: As far as I know, these devices are supported using a device spec file from DFP.
These devices are supproted in Atmel START and MCC for public. In fact, few IoT [examples](https://start.atmel.com/#examples/avr/iot/wg) are also available in 4808 which consumes almost 90% of memory resources. Anyway, I can check and confirm if there is any limitation.

feilipu: I have checked the `portmacro.h` and `port.c` file from the v10.0.0 files included in a ATmega4809 platform files.
 - `portmacro.h` - no relevant differences, but uses some depreciated gcc grammar.
 - `port.c` - removed optional debugging bytes in `pxPortInitialiseStack`, only standard 2-byte stack management, and some depreciated gcc grammar.

I think, if the `prvSetupTimerInterrupt()` is kept as an external or weak, along with the `portSCHEDULER_ISR` definition being weak, there are is nothing preventing support for these working.

I can add the small differences if desired. @yuhui-zheng please advise.

feilipu:
Ok. I think I'm done here. @yuhui-zheng for review, etc.

yuhui-zheng:
@feilipu  Thanks. The change for ```reti``` and ```ret``` makes sense -- the return from ```vPortYieldFromTick()``` preserves SREG I bit, and the second return from ISR sets SREG I bit to reenable interrupt. I'll see if my simulator effort can better capture this is a test maybe. 

Let's merge this PR, and make incremental changes to the port. I'll immediate follow up with a PR to move port to thirdparty directory per previous comment.

Issue #53, opened 2020-04-16
[BUG] ListItem_t and MiniListItem_t violate strict aliasing rules


Hawk777:
**Describe the bug**
ISO C’s “strict aliasing” rule says (roughly speaking) that the same memory must only be accessed either as the type it really is, or via a pointer to `char`. For example, an `int` can be accessed as an `int` or via a `char*`, but not via a `short*`. FreeRTOS breaks this rule: an object of type `List_t` contains an object of type `MiniListItem_t` (specifically `xListEnd`), but sometimes it is accessed via pointers of type `ListItem_t *` (not `MiniListItem_t *`). An example of this is in `uxListRemove` where, if removing the first item in the list, `pxItemToRemove->pxPrevious` is a `ListItem_t *` which points at `xListEnd`, and that pointer is written through in order to set `pxItemToRemove->pxPrevious->pxNext`, in violation of the strict aliasing rule.

Using aggressive optimizations in GCC 9, this generates bad code in `xTaskResumeAll` where `pxTCB` is not reloaded on each iteration of the first loop iterating over `xPendingReadyList`, resulting in the same TCB being added to the ready list more than once.

This could easily be fixed. Simply delete the `listFIRST_ITEM_INTEGRITY_CHECK_VALUE`, `xItemValue`, `pxNext`, and `pxPrevious` members from `ListItem_t` and replace them with an instance of `MiniListItem_t` (which contains exactly those members, and therefore does not change the memory layout of the structure at all), and fix up the resulting compile errors (which involves changing some `ListItem_t *` pointers to `MiniListItem_t *`). This should be totally standards-compliant. Now all iteration happens using the `MiniListItem_t` type, so the strict aliasing rule is obeyed. As long as a pointer is definitely pointing at a real item (not `xListEnd`), casting from `MiniListItem_t *` to `ListItem_t *` is also legal, because that is a cast between a pointer to a structure and a pointer to its first member, which is legal in both directions.

I would be happy to submit a PR if you agree that this should be changed.

Technically speaking I believe the `Static` types also constitute violations of the strict aliasing rule (for example, FreeRTOS accesses an object of type `struct xSTATIC_TCB` through a pointer of type `struct tskTaskControlBlock *`), but so far I have not seen this to cause actual problems, and it seems less likely to do so since the object is never actually accessed via its real type.

**Target**
- Instruction Set Architecture: ARMv7-M
- Toolchain and version: GCC 9.3.0

**Host**
[...]

PR #54, opened 2020-04-17
Obey strict aliasing rule in linked lists

Hawk777:
Description
-----------
ISO C’s “strict aliasing” rule says (roughly speaking) that the same memory must only be accessed either as the type it really is, or via a pointer to `char`. For example, an `int` can be accessed as an `int` or via a `char*`, but not via a `short*`. FreeRTOS breaks this rule: an object of type `List_t` contains an object of type `MiniListItem_t` (specifically `xListEnd`), but sometimes it is accessed via pointers of type `ListItem_t *` (not `MiniListItem_t *`). An example of this is in `uxListRemove` where, if removing the first item in the list, `pxItemToRemove->pxPrevious` is a `ListItem_t *` which points at `xListEnd`, and that pointer is written through in order to set `pxItemToRemove->pxPrevious->pxNext`, in violation of the strict aliasing rule.

Using aggressive optimizations in GCC 9, this generates bad code in `xTaskResumeAll` where `pxTCB` is not reloaded on each iteration of the first loop iterating over `xPendingReadyList`, resulting in the same TCB being added to the ready list more than once.

To fix this, replace the first four members of `ListItem_t` with an instance of `MiniListItem_t`.

Test Steps
-----------
Build with `armv7m-none-eabihf-gcc-9.3.0 -O3 -ggdb -g3 -fno-common -fno-math-errno -fstrict-aliasing -std=c99 -nostdlib -mabi=aapcs -fno-exceptions -ffreestanding -mslow-flash-data -flto=jobserver -mfloat-abi=hard -mcpu=cortex-m7 -mfpu=fpv5-sp-d16 -mthumb`. Prior to this PR, `xTaskResumeAll` would not work properly.

Related Issue
-----------
Fixes GH-53.

### DIFF
--- a/list.c
+++ b/list.c
@@ -39,7 +39,7 @@ void vListInitialise( List_t * const pxList )
 	/* The list structure contains a list item which is used to mark the
 	end of the list.  To initialise the list the list end is inserted
 	as the only list entry. */
-	pxList->pxIndex = ( ListItem_t * ) &( pxList->xListEnd );			/*lint !e826 !e740 !e9087 The mini list structure is used as the list end to save RAM.  This is checked and valid. */
+	pxList->pxIndex = &( pxList->xListEnd );			/*lint !e826 !e740 !e9087 The mini list structure is used as the list end to save RAM.  This is checked and valid. */
 
 	/* The list end value is the highest possible value in the list to
 	ensure it remains at the end of the list. */
@@ -47,8 +47,8 @@ void vListInitialise( List_t * const pxList )
 
 	/* The list end next and previous pointers point to itself so we know
 	when the list is empty. */
-	pxList->xListEnd.pxNext = ( ListItem_t * ) &( pxList->xListEnd );	/*lint !e826 !e740 !e9087 The mini list structure is used as the list end to save RAM.  This is checked and valid. */
-	pxList->xListEnd.pxPrevious = ( ListItem_t * ) &( pxList->xListEnd );/*lint !e826 !e740 !e9087 The mini list structure is used as the list end to save RAM.  This is checked and valid. */
+	pxList->xListEnd.pxNext = &( pxList->xListEnd );	/*lint !e826 !e740 !e9087 The mini list structure is used as the list end to save RAM.  This is checked and valid. */
+	pxList->xListEnd.pxPrevious = &( pxList->xListEnd );/*lint !e826 !e740 !e9087 The mini list structure is used as the list end to save RAM.  This is checked and valid. */
 
 	pxList->uxNumberOfItems = ( UBaseType_t ) 0U;
 
@@ -73,7 +73,7 @@ void vListInitialiseItem( ListItem_t * const pxItem )
 
 void vListInsertEnd( List_t * const pxList, ListItem_t * const pxNewListItem )
 {
-ListItem_t * const pxIndex = pxList->pxIndex;
+MiniListItem_t * const pxIndex = pxList->pxIndex;
 
 	/* Only effective when configASSERT() is also defined, these tests may catch
 	the list data structures being overwritten in memory.  They will not catch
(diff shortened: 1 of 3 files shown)


list.c
```diff
@@ -39,16 +39,16 @@ void vListInitialise( List_t * const pxList )
 	/* The list structure contains a list item which is used to mark the
 	end of the list.  To initialise the list the list end is inserted
 	as the only list entry. */
-	pxList->pxIndex = ( ListItem_t * ) &( pxList->xListEnd );			/*lint !e826 !e740 !e9087 The mini list structure is used as the list end to save RAM.  This is checked and valid. */
+	pxList->pxIndex = &( pxList->xListEnd );			/*lint !e826 !e740 !e9087 The mini list structure is used as the list end to save RAM.  This is checked and valid. */
```
Hawk777, line 42: A cast was removed here, which suggests that one of the lint error codes should probably be removed. But I don’t know what linting tool FreeRTOS uses or where to find a list of the codes, so I don’t know which one to remove. Could someone in the know suggest please?

list.c
```diff
@@ -39,16 +39,16 @@ void vListInitialise( List_t * const pxList )
...
 	/* The list end value is the highest possible value in the list to
 	ensure it remains at the end of the list. */
 	pxList->xListEnd.xItemValue = portMAX_DELAY;
 
 	/* The list end next and previous pointers point to itself so we know
 	when the list is empty. */
-	pxList->xListEnd.pxNext = ( ListItem_t * ) &( pxList->xListEnd );	/*lint !e826 !e740 !e9087 The mini list structure is used as the list end to save RAM.  This is checked and valid. */
-	pxList->xListEnd.pxPrevious = ( ListItem_t * ) &( pxList->xListEnd );/*lint !e826 !e740 !e9087 The mini list structure is used as the list end to save RAM.  This is checked and valid. */
+	pxList->xListEnd.pxNext = &( pxList->xListEnd );	/*lint !e826 !e740 !e9087 The mini list structure is used as the list end to save RAM.  This is checked and valid. */
+	pxList->xListEnd.pxPrevious = &( pxList->xListEnd );/*lint !e826 !e740 !e9087 The mini list structure is used as the list end to save RAM.  This is checked and valid. */
```
Hawk777, line 51: Same question here with the lint error codes.

list.c
```diff
@@ -147,17 +147,17 @@ const TickType_t xValueOfInsertion = pxNewListItem->xItemValue;
 			   before vTaskStartScheduler() has been called?).
 		**********************************************************************/
 
-		for( pxIterator = ( ListItem_t * ) &( pxList->xListEnd ); pxIterator->pxNext->xItemValue <= xValueOfInsertion; pxIterator = pxIterator->pxNext ) /*lint !e826 !e740 !e9087 The mini list structure is used as the list end to save RAM.  This is checked and valid. *//*lint !e440 The iterator moves to a different value, not xValueOfInsertion. */
+		for( pxIterator = &( pxList->xListEnd ); pxIterator->pxNext->xItemValue <= xValueOfInsertion; pxIterator = pxIterator->pxNext ) /*lint !e826 !e740 !e9087 The mini list structure is used as the list end to save RAM.  This is checked and valid. *//*lint !e440 The iterator moves to a different value, not xValueOfInsertion. */
```
Hawk777, line 150: Same question here with the lint error codes.

list.c
RichardBarry: I can take care of the lint comment (we are actually switching to Coverity anyway).

Hawk777: Great! In that case I have no other concerns myself, this just needs reviewing.

RichardBarry:
Regrettably, while I like this change, during testing it was noticed that it breaks all three of the FreeRTOS kernel aware debugger plug-ins it was tested with, so presumably will break the others too.  That means we will have to co-ordinate with partners (such as IAR) that ship FreeRTOS kernel awareness and release at a future date, or otherwise find a different way of fixing the GCC issue.  For now I'm going to close the PR.

Hawk777:
Is that coordination something that you would expect to happen at some point? What kind of timeframe would it involve?

I admit I never tried any such things (at such high optimization levels I’ve never had very good results with them anyway; required symbols tend to be optimized out of existence). However, OpenOCD, the only one I’m familiar with, appears to hardcode the memory layout of a list element, so since this patch doesn’t change the layout, it should be unaffected?

belchercw:
I just spent an entire week trying to track down a crash in my project which turns out to be this bug.  This is a very insidious issue that creates intermittent instability and is very difficult (read expensive) to track down.  As the correct fix has been rejected, what is the proposed resolution, and how do we avoid others continuing to encounter the problem?

PR #56, opened 2020-04-26
Fix Valgrind errors

alfred2g:
Fix Valgrind uninitialized variable errors

Description
-----------
==7996== Syscall param setitimer(&value->it_interval) points to uninitialised byte(s)
and
==7996== Syscall param rt_sigaction(act->sa_mask) points to uninitialised byte(s)
Test Steps
-----------
Run the port from FreeRTOS Demo similar to the one described in issue 1876

Related Issue
-----------
https://github.com/aws/amazon-freertos/issues/1876

### DIFF
--- a/portable/ThirdParty/GCC/Posix/port.c
+++ b/portable/ThirdParty/GCC/Posix/port.c
@@ -281,13 +281,13 @@ struct sigaction sigtick;
 	itimer.it_value.tv_sec = 0;
 	itimer.it_value.tv_usec = 0;
 
-        itimer.it_interval.tv_sec = 0;
-        itimer.it_interval.tv_usec = 0;  
+	itimer.it_interval.tv_sec = 0;
+	itimer.it_interval.tv_usec = 0;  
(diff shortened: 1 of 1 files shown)


portable/ThirdParty/GCC/Posix/port.c
```diff
@@ -280,10 +280,14 @@ struct sigaction sigtick;
...
 	itimer.it_value.tv_sec = 0;
 	itimer.it_value.tv_usec = 0;
+
+	itimer.it_interval.tv_sec = 0;
+	itimer.it_interval.tv_usec = 0;  
 	(void)setitimer( ITIMER_REAL, &itimer, NULL );
 
 	sigtick.sa_flags = 0;
 	sigtick.sa_handler = SIG_IGN;
+	sigemptyset( &sigtick.sa_mask );
```
yuhui-zheng, line 290: Shall we check return values here, or maybe ```(void)``` to mark as we don't care about the return?

(Same comment for the line below, but out of scope of this PR.)

PR #59, opened 2020-05-11
Tickless idle fixes/improvement

jefftenney:
Fixes/Improvement to Tickless Idle

Description
-----------
The default implementation of tickless idle on Cortex M has two minor edge-case bugs as identified in this forum thread: https://forums.freertos.org/t/ultasknotifytake-timeout-accuracy/9629.

Both bugs induce a "jump ahead" in xTickCount compared to time as measured by the SysTick.  The first bug (when stopping SysTick on zero) jumps ahead one full tick.  The second bug (when rescheduling an imminent tick after an early wakeup) jumps ahead *nearly* one full tick.

Note that the proposed fix for the first bug also includes an enhancement for tickless idle not to stop SysTick at all in the case `eTaskConfirmSleepModeStatus()` returns eAbortSleep.

Configurations that use the SysTick's alternate clocking have additional, related issues.  In these configurations, xTickCount might jump ahead by the entire expected idle time if the alternate clock is slow.  See https://forums.freertos.org/t/xeventgroupwaitbits-premature-timeout-cause/10432.  This pull request fixes those issues too.

Test Steps
-----------
These bugs are difficult to reproduce due to critical timing.  Could use some help here.


portable/GCC/ARM_CM4F/port.c
```diff
@@ -665,13 +669,24 @@ void xPortSysTickHandler( void )
...
 			portNVIC_SYSTICK_CURRENT_VALUE_REG = 0UL;
-			portNVIC_SYSTICK_CTRL_REG |= portNVIC_SYSTICK_ENABLE_BIT;
-			vTaskStepTick( ulCompleteTickPeriods );
+			portNVIC_SYSTICK_CTRL_REG = portNVIC_SYSTICK_CLK_BIT | portNVIC_SYSTICK_INT_BIT | portNVIC_SYSTICK_ENABLE_BIT;
+			#if( portNVIC_SYSTICK_CLK_BIT_SETTING != portNVIC_SYSTICK_CLK_BIT )
+			{
+				portNVIC_SYSTICK_CTRL_REG = portNVIC_SYSTICK_CLK_BIT_SETTING | portNVIC_SYSTICK_INT_BIT | portNVIC_SYSTICK_ENABLE_BIT;
+			}
+			#endif /* portNVIC_SYSTICK_CLK_BIT_SETTING */
 			portNVIC_SYSTICK_LOAD_REG = ulTimerCountsForOneTick - 1UL;
```
jefftenney, line 686: Should probably swap these two statements.  If the reload value is tiny (eg, 2) and the "other" clock is much slower than the core clock, you could get the exact same problem we're trying to fix here -- namely starting a second tiny period that doesn't end until after we unmask interrupts below.  Swapping the statements eliminates that possibility.  After swapping these two statements, the temporary use of the core clock lasts longer, unfortunately.  The effect will be added drift in kernel time, but the amount of added drift should be typical for tickless idle.

portable/GCC/ARM_CM4F/port.c
```diff
@@ -543,23 +530,44 @@ void xPortSysTickHandler( void )
...
+			if( ( portNVIC_ICSR_REG & portNVIC_PEND_SYSTICK_SET_BIT ) != 0 )
+			{
+				portNVIC_ICSR_REG = portNVIC_PEND_SYSTICK_CLEAR_BIT;
+				
+				/* Skip the correction described above if SysTick happened to
+				stop on zero.  In that case, there are still
+				ulTimerCountsForOneTick ticks left in the second tick period,
+				not zero, so the value in portNVIC_SYSTICK_CURRENT_VALUE_REG
+				already includes the correction normally done here. */
+				if( portNVIC_SYSTICK_CURRENT_VALUE_REG != 0 )
```
jefftenney, line 565: Another concern here about configurations that use the "other" SysTick clock.  If the other clock is very slow compared to the core clock, then the SysTick could be zero here *without* the IRQ asserted.  If the SysTick were zero when the idle task calculated xExpectedIdleTime for the final time, then potentially it could still be zero here.  The code proposed here does not properly handle that case because it checks SysTick for zero only if the IRQ is asserted.  Does FreeRTOS not place any restrictions on this "other" clock?  The other clock seems to be MCU specific.  I think EFM32 uses 32kHz for the other clock.  It wouldn't be too hard to support these very slow SysTick clocks, but at some runtime cost to all configurations.

yuhui-zheng:
Thank you for the submission. I followed the forum post, will look into this PR.

jefftenney:
Awesome thanks Yuhui.  I am happy to duplicate to the rest of the ports/compilers when we get to that point, or do whatever is most helpful.

portable/GCC/ARM_CM4F/port.c
```diff
@@ -607,26 +620,22 @@ void xPortSysTickHandler( void )
 			be, but using the tickless mode will inevitably result in some tiny
 			drift of the time maintained by the kernel with respect to calendar
 			time*/
-			portNVIC_SYSTICK_CTRL_REG = ( portNVIC_SYSTICK_CLK_BIT | portNVIC_SYSTICK_INT_BIT );
+			portNVIC_SYSTICK_CTRL_REG = ( portNVIC_SYSTICK_CLK_BIT_SETTING | portNVIC_SYSTICK_INT_BIT );
```
yuhui-zheng, line 623: Same here. When setting portNVIC_SYSTICK_CTRL_REG with this assignment, portNVIC_SYSTICK_COUNT_FLAG_BIT (bit-16) is cleared. Looks like the code always ends up in the else branch. 

```
portNVIC_SYSTICK_CTRL_REG &= ~portNVIC_SYSTICK_ENABLE_BIT;
```

portable/GCC/ARM_CM4F/port.c
```diff
@@ -656,13 +683,24 @@ void xPortSysTickHandler( void )
...
+			then set portNVIC_SYSTICK_LOAD_REG back to its standard value.  If
+			the SysTick is not using the core clock, temporarily configure it to
+			use the core clock.  This configuration forces the SysTick to load
+			from portNVIC_SYSTICK_LOAD_REG immediately instead of at the next
+			cycle of the other clock.  Then portNVIC_SYSTICK_LOAD_REG is ready
+			to receive the standard value immediately. */
 			portNVIC_SYSTICK_CURRENT_VALUE_REG = 0UL;
-			portNVIC_SYSTICK_CTRL_REG |= portNVIC_SYSTICK_ENABLE_BIT;
-			vTaskStepTick( ulCompleteTickPeriods );
+			portNVIC_SYSTICK_CTRL_REG = portNVIC_SYSTICK_CLK_BIT | portNVIC_SYSTICK_INT_BIT | portNVIC_SYSTICK_ENABLE_BIT;
```
yuhui-zheng, line 694: Same here, would prefer 
```
portNVIC_SYSTICK_CTRL_REG |= portNVIC_SYSTICK_ENABLE_BIT;
```

portable/GCC/ARM_CM4F/port.c
```diff
@@ -656,13 +683,24 @@ void xPortSysTickHandler( void )
...
 			portNVIC_SYSTICK_LOAD_REG = ulTimerCountsForOneTick - 1UL;
+			#if( portNVIC_SYSTICK_CLK_BIT_SETTING != portNVIC_SYSTICK_CLK_BIT )
```
yuhui-zheng, line 696: May need some explanation on this #if-#endif. 

My understanding of this #if-#endif --
if using external clock (bit-2 is 0), it seems like we are trying to re-enable SysTick. 
Would it be necessary?

portable/GCC/ARM_CM4F/port.c
```diff
@@ -543,23 +534,45 @@ void xPortSysTickHandler( void )
 		to be unsuspended then abandon the low power entry. */
 		if( eTaskConfirmSleepModeStatus() == eAbortSleep )
```
yuhui-zheng, line 535: I think this is a good optimization -- confirming sleep mode first, and only disable SysTick temporarily in the branch where tickless idle is needed. 👍

jefftenney: Thank you.  It takes a little extra work to make it happen and some consideration of a race condition the code had previously avoided.  But I think it's worth doing especially when bug fixes are going into the same function.

portable/GCC/ARM_CM4F/port.c
```diff
@@ -543,23 +534,45 @@ void xPortSysTickHandler( void )
...
 			above. */
 			__asm volatile( "cpsie i" ::: "memory" );
 		}
 		else
 		{
+			/* Stop the SysTick momentarily.  The time the SysTick is stopped for
+			is accounted for as best it can be, but using the tickless mode will
+			inevitably result in some tiny drift of the time maintained by the
+			kernel with respect to calendar time. */
+			portNVIC_SYSTICK_CTRL_REG = ( portNVIC_SYSTICK_CLK_BIT_SETTING | portNVIC_SYSTICK_INT_BIT );
```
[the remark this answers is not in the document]
jefftenney: Note that portNVIC_SYSTICK_CLK_BIT_SETTING (with the _SETTING suffix) is configuration dependent and holds the correct value for the CLK bit (bit 2), whether set or clear.  So in reality the instruction as written really is just clearing the enable bit without any other side effects.  Just like the original code [here](https://github.com/FreeRTOS/FreeRTOS-Kernel/blob/4b353bfd7aa009d553f31b4e47dfb49a9f3c75b1/portable/GCC/ARM_CM4F/port.c#L610).  It's more efficient since it avoids have to do read-modify-write.

This method of stopping systick (a simple write instead of RMW) is required the second time we stop systick (below).  Because this method is already required in this function and is more efficient than RMW, I used it for the first stopping of systick here too.  Thoughts?

portable/GCC/ARM_CM4F/port.c
jefftenney: Note that portNVIC_SYSTICK_COUNT_FLAG_BIT is a read-only bit.  It is cleared by any *read* of the CTRL register and by any write to the current-value register.  The original code [here](https://github.com/FreeRTOS/FreeRTOS-Kernel/blob/4b353bfd7aa009d553f31b4e47dfb49a9f3c75b1/portable/GCC/ARM_CM4F/port.c#L610) relies on this behavior to stop the systick (the second time) without modifying the count flag.  Also note perhaps more importantly that there is no change here in the compiled code from the original; the new symbol portNVIC_SYSTICK_CLK_BIT_SETTING has the same value as the old symbol portNVIC_SYSTICK_CLK_BIT.

portable/GCC/ARM_CM4F/port.c
jefftenney: This one is special and is a critical part of supporting the systick's "alternate" clock.  For configurations that use the alternate clock (portNVIC_SYSTICK_CLK_BIT_SETTING == 0), the systick is started here using the core clock (portNVIC_SYSTICK_CLK_BIT -- notice the _SETTING suffix is not used here), but only temporarily.  Please let me know if the new (wordy) comment above this code isn't clear enough.

portable/GCC/ARM_CM4F/port.c
jefftenney: The new code comments attempt to explain.  This #if - #endif block changes the clocking configuration of the systick from using the core clock to using the alternate clock but only in configurations that use the alternate clock.  We detect configurations that use the alternate clock by checking portNVIC_SYSTICK_CLK_BIT_SETTING.

jefftenney:
Thanks for the review.

It may help (or hurt!) your meditation efforts regarding stopping systick on zero to read the monster commit message on the first commit.  Then when your meditation is yielding some fruit, read the (resolved) comments when I reviewed my own changes [here](https://github.com/FreeRTOS/FreeRTOS-Kernel/pull/59#discussion_r426169054).  That self review led to the third commit.

portable/GCC/ARM_CM4F/port.c
```diff
@@ -101,6 +94,19 @@ occurred while the SysTick counter is stopped during tickless idle
...
+#ifndef configSYSTICK_CLOCK_HZ
+	#define configSYSTICK_CLOCK_HZ configCPU_CLOCK_HZ
+	/* Ensure the SysTick is clocked at the same frequency as the core. */
+	#define portNVIC_SYSTICK_CLK_BIT_SETTING	( portNVIC_SYSTICK_CLK_BIT )
+#else
+	/* Select the option to clock SysTick not at the same frequency as the core.
+	The clock used is often a divided version of the core clock. */
+	#define portNVIC_SYSTICK_CLK_BIT_SETTING	( 0 )
+#endif
+
```
jefftenney, line 109: I may have caused a readability issue here with the `_SETTING` suffix.  It's not an established naming convention in FreeRTOS code.  The intention of the `_SETTING` suffix is to allow `portNVIC_SYSTICK_CLK_BIT` to have a fixed, expected value, regardless of build configuration.  The new code at line 694 uses this bit regardless of build configuration.  Would a `_CONFIG` suffix be more readable?  Is there some other convention already established in FreeRTOS code that we could use instead?

portable/GCC/ARM_CM4F/port.c
yuhui-zheng: I understand your PR did not change the line you are referring to. But I doubt there was a bug at [the line](https://github.com/FreeRTOS/FreeRTOS-Kernel/blob/4b353bfd7aa009d553f31b4e47dfb49a9f3c75b1/portable/GCC/ARM_CM4F/port.c#L610), so that the if path (line 619-640) is never taken. 

I'm referring to this Cortex M4 datasheet -- https://developer.arm.com/docs/dui0553/a/cortex-m4-peripherals/system-timer-systick
```SYST_CSR``` is RW, and ```COUNTFLAG``` is at bit 16. 

I setup a quick demo to prove this line  
```
portNVIC_SYSTICK_CTRL_REG = ( portNVIC_SYSTICK_CLK_BIT | portNVIC_SYSTICK_INT_BIT );
```
overwrites ```COUNTFLAG``` bit. And it seems to be the case. In below snapshot, I added line 603-608 and set breakpoint at 607. So if we ever trap at 607, we should also trap at line 630, if  ```SYST_CSR``` is read-only. Not quite the case. (I'm using Keil uVision v5.30. And a minor note is that the System Tick Timer window also in the snapshot is incorrectly reading ```COUNTFLAG``` bit. In this case we trapped at 607, but bit 16 is not set. I'm not sure if that contributed to "read-only" observation in your comment.. )


Screen Shot 2020-06-01 at 11 39 00 PM

jefftenney: No, there's no bug here in the original code or in the proposed code.  And the "minor note" you mentioned about your test code incorrectly reading `COUNTFLAG` is not actually happening either.

Here is the explanation:  The `COUNTFLAG` is cleared by any **read** of the `SYS_CSR`.  So as when your new "if" statement executes, regardless of the outcome of that statement, it clears the COUNTFLAG.

See here:
Capture

yuhui-zheng: I'm good now, apologies for the detour. Also break at if statement instead of inside the if, and occasionally I do see the bit got set and cleared by read. Going to resolve the this conversation and ```COUNTFLAG``` related.

jefftenney: No worries.  It's a strange bit.  I appreciate the thorough review.

n9wxu (dismissed):

jefftenney:
Thanks for the review Joseph.  Do you want me to resolve the merge conflicts?  Do you want me to replicate the changes to the other Cortex M ports -- I think there are 23 of them.

n9wxu:
We will need to replicate and validate the changes across the ports.  We will not be able to merge this until the validation has happened.  If you can replicate that will help.

jefftenney:
@n9wxu I believe this is ready for validation now.  My own testing was on CM4 with gcc and iar only.

n9wxu (approved):

jefftenney:
@n9wxu FYI this repo https://github.com/jefftenney/LPTIM-Tick demonstrates some of the issues addressed by this PR.  It tests tick timing (with or without tickless) against a reference timebase.  It is platform specific (Nucleo-L476RG) and tool specific (GCC via CubeIDE) but might be helpful to whoever is doing the validation work for this PR.

jefflongo:
Posting this thread for context
https://forums.freertos.org/t/vportsuppressticksandsleep-systick-implementation-wrong/13521

jefftenney:
@n9wxu I came back around to this PR today while adapting [LPTIM-Tick for STM32U5](https://github.com/jefftenney/LPTIM-Tick-U5).

I tested with `configSYSTICK_CLOCK_HZ` set to 32kHz, and the drift/slippage was awful, with or without this PR.  (This PR is an improvement, but still bad.)

Here's a brief explanation of the root cause.  With a slow SysTick clock, the correct compensation for stopping the timer briefly in `vPortSuppressTicksAndSleep()` is a _partial_ timer count.  Code however can only apply zero or one timer counts of compensation in that case.  The error resulting from applying either zero or one instead of the fraction is (1) large in absolute time for a slow timer and is (2) multiplied by the number of interrupts causing early wake-ups.  Drift is out of control.  For example during a barrage of interrupts the tick count could freeze for long periods of time or could race ahead at 10x its proper rate.

If the team is considering merging this PR, then maybe this PR is also the time to consider dropping support for `configSYSTICK_CLOCK_HZ` and the use of the "other" clock to drive SysTick.  It is of very limited value in the first place.  It adds complexity to the code.  And it can cause uncontrolled drift as explained above.  My suggestion would be to drop support.

If you decide to keep support, I would still like to make a small change to this PR to reduce the amount of time spent [temporarily on the CPU clock](https://github.com/FreeRTOS/FreeRTOS-Kernel/blob/493b3cf96e4c2614a5a920771e94019a763dc426/portable/CCS/ARM_CM3/port.c#L546-L552).  It will reduce the set of cases that encounter uncontrolled drift.

RichardBarry:
The default "suppress ticks and sleep" function is just that - a default.  I think with such a slow clock replacing the whole function might be the best path.  It is a [weakly defined](https://github.com/FreeRTOS/FreeRTOS-Kernel/blob/main/portable/GCC/ARM_CM4F/port.c#L518) symbol for that purpose.

jefftenney:
Also I've just realized there is a benefit to using `configSYSTICK_CLOCK_HZ`, regardless of tickless idle.  If the "other" clock has a fixed speed, using that clock to drive SysTick allows the application to change CPU clock speeds at will, without impacting the system tick.  There may be applications out there now doing this very thing.  STM32U5, EFM32, RP2040, and i.MX are some MCUs that provide a fixed-speed SysTick clocking option.  So nevermind on dropping support for `configSYSTICK_CLOCK_HZ`.

It's probably best to make the default tickless implementation as good as we can do, within reason, and then simply be aware of its limitations with slow SysTick clocking.  To that end I will plan to make a small change to this PR.

jefftenney:
@RichardBarry @n9wxu I made the necessary fixes for slow SysTick.  This PR is ready to go.

Here are some test results with and without this PR.

Platform: CM33 (STM32U585)
Core Clock: 24MHz
Tick Rate: 1000Hz
Test: [LPTIM-Tick-U5](https://github.com/jefftenney/LPTIM-Tick-U5), Test 3

This table shows the actual time of 10 ticks: `avg (min/max)`
image

I let each row in the PR59 column soak for >24 hrs.  And I have run similar tests on CM4 (STM32L476) with the same results.  (No slow SysTick option on that MCU though.)

RichardBarry:
👍 Thanks.  It will take a little time to process as we will need to try each location the code tickless code got updated.

[as an aside:  The way the ARMv7-M ports grew organically over time resulted in replication of the tickless and other functions in each port.  As you know, the core of the ARMv8-M port is centralised in the hope of preventing that replication happening again).

jefftenney:
No worries on this PR's timeline.  I made this PR in response to forum posts and not to benefit any of my own projects.

For this PR I did a manual copy/paste among the ARMv6-M and ARMv7-M ports.  All ports that use the same compiler have identical tickless code, which simplified the manual replication.  Replicating these last changes took only about 10 minutes because the changes were compiler independent.  So any feedback you have about this PR is most welcome, with no worries about additional replication effort.

jefftenney:
@n9wxu @aggarg There's a little bit more motivation now to merge this PR.  PR #484 was just merged to provide support for the "other" systick clocking option on CM0.  A popular target will be RP2040 which has a 1 MHz option for the "other" systick clock (as noted in the [forum post](https://forums.freertos.org/t/is-there-a-reason-m0-port-doesnt-respect-configsystick-clk-hz/14810)).  However, anyone that actually tries using the 1 MHz systick clock on RP2040 may run into an issue resolved by this PR.  The issue causes the tick count to jump ahead by the entire expected idle time when in reality no time elapsed.  The issue doesn't just affect RP2040, and not just CM0 either, but any Cortex M that offers a slow systick clock.  Slow here being anything less than about 10% of the CPU clock (guess based on code review).

n9wxu:
Verified on RP2040.  Tickless idle is working correctly.

n9wxu:
Compiles clean w/ GCC

n9wxu:
Compiles clean with IAR and ARMCLANG

n9wxu (approved):

kstribrnAmzn (approved):

binliu-philips:
I tested V11.0.0 on my setup which can easily reproduce the issue before, and am happy to say my setup could no longer see the issue!

PR #63, opened 2020-05-26
Feature/multiple direct to task notifications

RichardBarry:
Description
-----------
Before this change each task had a single direct to task notification value and state as described here: https://www.FreeRTOS.org/RTOS-task-notifications.html.  After this change each task has an array of task notifications, so more than one task notification value and state per task.  The new FreeRTOSConfig.h compile time constant configTASK_NOTIFICATION_ARRAY_ENTRIES sets the number of indexes in the array.

Each notification within the array operates independently - a task can only block on one notification within the array at a time and will not be unblocked by a notification sent to any other array index.

Task notifications were introduced as a light weight method for peripheral drivers to pass data and events to tasks without the need for an intermediary object such as a semaphore - for example, to unblock a task from an ISR when the operation of a peripheral completed.  That use case only requires a single notification value.  Their popularity and resultant expanded use cases have since made the single value a limitation - especially as FreeRTOS stream and message buffers themselves use the notification mechanism.  This change resolves that limitation.  Stream and message buffers still use the task notification at array index 0, but now application writers can avoid any conflict that might have with their own use of task notifications by using notifications at array indexes other than 0.

The pre-existing task notification API functions work in a backward compatible way by always using the task notification at array index 0.  For each such function there is now an equivalent that is postfixed "Indexed" and takes an additional parameter to specify which index within the array it should operate upon.  For example, xTaskNotify() is the original that only operates on array index 0.  xTaskNotifyIndexed() is the new function that can operate on any array index.

Test Steps
-----------
The update is tested using the Win32 demo (PR to be created in the FreeRTOS/FreeRTOS github repo), which has been updated to build and run a new test file FreeRTOS/Demo/Common/Minimal/TaskNotifyArray.c.  The tests in that file are in addition to, not a replacement for those in FreeRTOS/Demo/Common/Minimal/TaskNotify.c.


syntroniks:
I had a bit of issues with this latest addition.
My `./include/FreeRTOS.h` had 
```
#ifndef traceTASK_NOTIFY_TAKE
	#define traceTASK_NOTIFY_TAKE()
#endif

#ifndef traceTASK_NOTIFY_WAIT_BLOCK
	#define traceTASK_NOTIFY_WAIT_BLOCK()
#endif

#ifndef traceTASK_NOTIFY_WAIT
	#define traceTASK_NOTIFY_WAIT()
#endif
```
Under arm-none-eabi-gcc 9.2.1, I was met with 
```
...
[build] In file included from ../extern/FreeRTOS-Kernel/tasks.c:38:
[build] ../extern/FreeRTOS-Kernel/include/FreeRTOS.h:648: note: macro "traceTASK_NOTIFY_WAIT" defined here
[build]   648 |  #define traceTASK_NOTIFY_WAIT()
[build]       | 
[build] ../extern/FreeRTOS-Kernel/tasks.c:4762:4: error: 'traceTASK_NOTIFY_WAIT' undeclared (first use in this function)
[build]  4762 |    traceTASK_NOTIFY_WAIT( uxIndexToWait );
...
```
Changing FreeRTOS.h to 
```
#ifndef traceTASK_NOTIFY_TAKE
	#define traceTASK_NOTIFY_TAKE( uxIndexToWait )
#endif

#ifndef traceTASK_NOTIFY_WAIT_BLOCK
	#define traceTASK_NOTIFY_WAIT_BLOCK()
#endif

#ifndef traceTASK_NOTIFY_WAIT
	#define traceTASK_NOTIFY_WAIT( uxIndexToWait )
#endif
```
Solves this issue.

PR #73, opened 2020-06-10
Fix support for stepping tick by xExpectedIdleTime

jefftenney:
Description
-----------
This PR fixes support for user implementations of tickless idle that call `vTaskStepTick()` with `xExpectedIdleTime` ticks to step.  The new code ensures `xTickCount` reaches `xNextTaskUnblockTime` inside `xTaskIncrementTick()` instead of inside `vTaskStepTick()`.

Without this PR
--------
**Typical Case: Task wakes up one tick late**
* At `xTickCount` = 10, app task calls `vTaskDelay(5)`
* Idle task calls `vPortSuppressTicksAndSleep(5)` (`xExpectedIdleTime` = 5)
* `vPortSuppressTicksAndSleep()` sleeps 5 ticks and calls `vTaskStepTick(5)` [custom implementation -- not FreeRTOS code]
* [App task remains on delayed list, even though `xNextTaskUnblockTime` == `xTickCount`.]
* One tick period elapses (idle task running the idle loop)
* Tick interrupt occurs, and `xTaskIncrementTick()` sets `xTickCount` to 16 and moves app task to ready list.
* App task wakes up at `xTickCount` = 16 but should have woken at `xTickCount` = 15.

**Rare Case: Assertion failure when xTickCount rolls over**
* Same as example above except starting value for xTickCount is 0xFFFFFFFA instead of 10.
* App task should wake at 0xFFFFFFFF but doesn't (same as above).
* Next tick occurs, and the rollover logic for the tick counter asserts due to the delayed-task list not being empty.

Test Steps
-----------
Choose any low-power demo, and use the `vPortSuppressTicksAndSleep()` version below.  (If there is already a custom version of `vPortSuppressTicksAndSleep()` in the demo, remove it.)  Strip the demo down to just one task, with no active timers, then insert a call to function `test()` from that task.  Without this PR, the `configASSERT()` statement in `test()` fails.  Then remove that statement to see the next failure at `vTaskDelay( xDelta );` -- another assertion failure.  After the PR, there are no assertion failures.  Note the actual demo won't function normally with this test code even after the PR because this demo version of  `vPortSuppressTicksAndSleep()` doesn't actually sleep.  It just illustrates the API issue.
 
```
void vPortSuppressTicksAndSleep( TickType_t xExpectedIdleTime )
{
    /* This version of vPortSuppressTicksAndSleep() exercises the tickless API
     * only.  The specific case being simulated is a sleep that completes fully
     * with no early wake-up.  The sleep ends via timer other than the tick
     * timer, so no tick is pending. */
    vTaskStepTick( xExpectedIdleTime );
}

void test()
{
    /* Get to the beginning of a tick period. */
    vTaskDelay( 1 );

    /* Sleep (tickless idle) for ten ticks. */
    TickType_t xTickCount = xTaskGetTickCount();
    vTaskDelay( 10 );

    /* Make sure we slept only ten ticks. */    
    TickType_t xDelta = xTaskGetTickCount() - xTickCount;
    configASSERT( xDelta == 10 );  /* fails without PR 73 */

    /* Determine how many ticks until just before the tick rolls over. */
    xTickCount += xDelta;
    xDelta = portMAX_DELAY - xTickCount;

    /* Sleep (tickless idle) until just before the tick rolls over. */
    vTaskDelay( xDelta );  /* causes assert in tasks.c without PR 73 */
}
```

Related Issue
-----------
https://forums.freertos.org/t/assert-in-tasks-c-2611-on-xtickcount-wrap-around/9984


tasks.c
```diff
@@ -2616,7 +2631,9 @@ BaseType_t xTaskCatchUpTicks( TickType_t xTicksToCatchUp )
     /* Use xPendedTicks to mimic xTicksToCatchUp number of ticks occurring when
      * the scheduler is suspended so the ticks are executed in xTaskResumeAll(). */
     vTaskSuspendAll();
+    taskENTER_CRITICAL();
```
n9wxu, line 2634: vTaskSuspendAll has already shut the scheduler down so we should not need taskENTER_CRITICAL.

tasks.c
```diff
@@ -2616,7 +2631,9 @@ BaseType_t xTaskCatchUpTicks( TickType_t xTicksToCatchUp )
     /* Use xPendedTicks to mimic xTicksToCatchUp number of ticks occurring when
      * the scheduler is suspended so the ticks are executed in xTaskResumeAll(). */
     vTaskSuspendAll();
+    taskENTER_CRITICAL();
     xPendedTicks += xTicksToCatchUp;
+    taskEXIT_CRITICAL();
```
n9wxu, line 2636: Same as above

n9wxu:
I have been testing this PR in QEMU by modifying the CORTEX_MPS2_QEMU_IAR_GCC demo as described in your testing section.  I have 2 observations:
1. Your test without the PR applied fails in `vPortEnterCritical()`.
2. Your test with the PR applied fails after a few hundred iterations as the xDelta becomes 11.  If I widen the assert to allow 11's and print the xDelta value, I see mostly 10's and the occasional 11.
 
Can you provide additional details on how you tested this PR?

tasks.c
jefftenney: I think both critical sections are required in case a tick interrupt occurs while this code is trying to modify `xPendedTicks`.  When the scheduler is suspended, the tick interrupt increments `xPendedTicks`.

jefftenney:
> 1. Your test without the PR applied fails in `vPortEnterCritical()`.

Yes I would expect this result without the PR, assuming your configASSERT() is written to enter a critical section and hang (typical).  So far so good it seems.

> 2. Your test with the PR applied fails after a few hundred iterations as the xDelta becomes 11. If I widen the assert to allow 11's and print the xDelta value, I see mostly 10's and the occasional 11.

I would not expect this.  It should be 10's all the time, reliably, assuming a truly monotonic tick rate and no other tasks and no active timers.  Your observation indicates that a full tick period elapses while QEMU executes just a handful of instructions.  Would you be able to repeat on real silicon?  I don't know enough about QEMU and its tick generation to help there.

For my own testing, to the best of my memory I modified one of my low-power projects and ran it on real silicon.

RichardBarry:
I have a question about this line:  "support for user implementations of tickless idle that call vTaskStepTick() with xExpectedIdleTime ticks to step".  Is that a valid thing to do?  Maybe I am repeating back something you are trying to report because I don't understand it fully, but I'm wondering if the problem is an invalid parameter to vTaskStepTick().  I will try and be granular so we can discuss whichever bit I'm misunderstanding or you think is not correct:

There is the following sequence if I run your test() function:

- Task calls vTaskDelay( 10 ) when the tick count is 1.
- The scheduler suspends itself and calls vPortSuppressTicksAndSleep().
- Would would have been the 10th tick brings the system out of sleep.
- Interrupts get enabled, resulting in the tick interrupt executing.
- The interrupt handler pends incrementing the tick because the scheduler is suspended.
- As per [this comment](https://github.com/FreeRTOS/FreeRTOS-Kernel/blob/V10.4.6/portable/GCC/ARM_CM3/port.c#L586),  vTaskStepTick() is called to step 9 ticks, not 10.
- The scheduler unsuspends itself - during the process the pended tick is noted, and the tick count gets incremented for real. 
- Incrementing the tick results in the task being moved from the blocked to the ready list, and there is a yield to the task after 10 ticks.

Whereas in the test code, using:
```
void vPortSuppressTicksAndSleep( TickType_t xExpectedIdleTime )
{
    /* This version of vPortSuppressTicksAndSleep() exercises the tickless API
     * only.  The specific case being simulated is a sleep that completes fully
     * with no early wake-up.  The sleep ends via timer other than the tick
     * timer, so no tick is pending. */
    vTaskStepTick( xExpectedIdleTime );
}
```

The tick is stepped 10, but the task is never moved from the ready list to the blocked list until there is another tick interrupt - which would [effectively] be the 11th, and the assert():

`configASSERT( xDelta == 10 );`

fails, because the delta is actually 11 due to the delay in context switching to the next tick.

jefftenney:
Yes I agree with your description above.  And you describe exactly why `configASSERT( xDelta == 10 );` fails without this PR.  My opinion is we should improve vTaskStepTick() so that assertion would *not* fail.  That's what this PR does.

> Is it valid to pass `xExpectedIdleTime` to `vTaskStepTick()`?  

No, not with the official tickless implementation.  The official implementation uses the tick timer to wake the system, which pends a tick.  So the official implementation would never pass `xExpectedIdleTime ` but `xExpectedIdleTime - 1` instead.

The key question for this PR is this -- in a *custom* tickless idle implementation, is it valid to pass `xExpectedIdleTime` to `vTaskStepTick()`?  Consider an implementation that does not use the tick timer to wake the system but perhaps uses a different timer that operates in ultra-low-power modes.  In that timer's ISR there is no call xTaskIncrementTick().  That implementation would then be correct in passing `xExpectedIdleTime` to `vTaskStepTick()`.  Without this PR, this custom implementation appears to work but tasks wake up one tick late.  Not too big of a deal unless we're trying to wake up just before tick rollover (rare case but possible).  In that case, the rollover logic for the tick counter [asserts](https://github.com/FreeRTOS/FreeRTOS-Kernel/blob/abd887c687fb990014f38a3227f4ef1796eadf7b/tasks.c#L206) due to the delayed-task list not being empty.

As a "fix", the custom implementation could merely add a call to xTaskIncrementTick() to the other timer's ISR.  In that case, this PR is not needed.  But the "fix" in this PR is so I easy I figured we might as well add it.  And it saves the person who is adventurous enough to write their own tickless implementation from falling into the assertion failure during tick rollover.  I probably wouldn't have created this PR had [this assertion statement](https://github.com/FreeRTOS/FreeRTOS-Kernel/blob/abd887c687fb990014f38a3227f4ef1796eadf7b/tasks.c#L2600) prohibited passing `xExpectedIdleTime`.  That would have been enough to save the developer of the custom tickless implementation.

jefftenney:
@n9wxu Let me take a closer look at this test code.  I found my old testing code and changed it to call test() from a loop and I do get failures just like you do.  I'll get back to you.

jefftenney:
@n9wxu Nevermind -- false alarm.  I had to reapply the PR to my old test code.  I just ran it in a loop 100,000 times successfully.

RichardBarry:
[I made some minor edits, just to add a comment and indent the code within the critical section]

So I suppose what this does is move the subtraction of the one tick from the port specific code into the common code - although we can't change the port specific code now as its in too wide use.

Probably the real thing that is missing is much deeper documentation on authoring these tick suppression functions.

RichardBarry (approved):

RichardBarry:
Oops - looks like we will have to uncrustify before actually merging.

jefftenney:
> Probably the real thing that is missing is much deeper documentation on authoring these tick suppression functions.

Agreed.  There might even be an empty chapter in a book somewhere.  :wink:

tasks.c
```diff
@@ -2592,12 +2592,30 @@ char * pcTaskGetName( TaskHandle_t xTaskToQuery ) /*lint !e971 Unqualified char
...
+            configASSERT( xTicksToJump != 0 );
+
+            /* Prevent the tick interrupt modifying xPendedTicks simultaneously. */
+            taskENTER_CRITICAL();
+            {
+                xPendedTicks++;
+            }
+            taskEXIT_CRITICAL();
+            xTicksToJump--;
+        }
```
jefftenney, line 2617: @n9wxu @RichardBarry Do we need to add:

```
else
{
    mtCOVERAGE_TEST_MARKER();
}
```

RichardBarry: Yes - good call.

n9wxu (approved):

aggarg (approved):

PR #77, opened 2020-06-23
vTaskDelayUntil improvement

### LINKED ISSUE #76: [Inquiry] Suggestions to improve vTaskDelayUntil
Hello,

I would like to suggest some possible improvements for vTaskDelayUntil.

For our application, we would like to check whether a deadline for a periodic task was missed. 
For us, this is generally the case if vTaskDelayUntil() does not delay the task. Unfortunately, `vTaskDelayUntil` does not have a returnvalue, which tells me whether the task was delayed or not ( could also be implemented by passing a pointer to a boolean), so we needed to perform the check before calling `vTaskDelayUntil` (and we also had to implement the overflow checking again) because the last wake time will be updated internally by it. What do you think about this?

Basically, the interface for `vTaskDelayUntil` would change from

```cpp
void vTaskDelayUntil( TickType_t * const pxPreviousWakeTime, const TickType_t xTimeIncrement)
``` 

to

```cpp
void vTaskDelayUntil( TickType_t * const pxPreviousWakeTime, const TickType_t xTimeIncrement, BaseType_t* xWasDelayed)
``` 

And the local variable `xShouldDelay` is assigned too `*xWasDelayed` at the function end.
Unfortunately, this is an API change but I think it would be a nice feature to implement the checking of missed deadlines.

rmspacefish:
Description
This is the pull request related to issue #76. Fixes #76 
It adds a way to find out whether a task was actually delayed by vTaskDelayUntil,
which can be helpful for deadline checking.

This is an API change because vTaskDelayUntil has a third function parameter now.


RichardBarry:
Thanks for taking the time to create this PR.  Unfortunately this is a breaking change so we are unable to accept it in its current format.

What do you think to changing the name of the function to xTaskDelayUntil() (the x showing there is a BaseType_t return value), then returning pdTRUE/pdFALSE instead of introducing a new parameter for that value to be returned.  Then we can #define vTaskDelay() to xTaskDelay() to ensure existing projects continue working without modification.

rmspacefish:
Good idea. I will implement it like you suggested.

rmspacefish:
Alright, I integrated the changes and tested this in our project as well, appeared to work without issues (and I was able to determine missed deadlines).
I also adapted the documentation and the example.

rmspacefish:
If you are still interested in the proposal, I can merge the upstream master, there appears to be new conflicts..

n9wxu (commented):
I apologize for the tardy review.  If you can rebase and sort out the recent changes I will approve this PR.

rmspacefish:
Branch was rebased.

n9wxu (approved):

n9wxu:
I would like to approve it but the new words you added to the API need to be added to the lexicon so it will pass a spell check.
can you add the new words to lexicon.txt?

lundinc2 (approved):

n9wxu (approved):

PR #82, opened 2020-07-07
Fix: 2 1-line fixes, xTaskCheckForTimeOut() and ulTaskGenericNotifyStateClear()

RichardBarry:
First 1-line fix:
ulTaskGenericNotifyValueClear() returned the notification value of the
currently running task, not the target task.  Now it returns the
notification value of the target task.

second 1-line fix:
Some users expected xTaskCheckForTimeOut() to clear the 'last wake time'
value each time a timeout occurred, whereas it only did that in one path.
It now clears the last wake time in all paths that return that a timeout
occurred.

### DIFF
--- a/tasks.c
+++ b/tasks.c
@@ -144,7 +144,7 @@
         }                                                                     \
                                                                               \
         /* listGET_OWNER_OF_NEXT_ENTRY indexes through the list, so the tasks of \
-         * the	same priority get an equal share of the processor time. */                    \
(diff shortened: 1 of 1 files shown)


yuhui-zheng:
Rebased. Please enable "hide whitespace changes" when reviewing, and you'll see the exact three of changes. 

\* The rebased was needed due to Style change. I formatted manually, please run this with CI if available.

PR #83, opened 2020-07-09
Error in ARM_CM0 port compiled with IAR 8.50.5.

DavidJurajdaNXP:
IAR EW in release 8.50.5 modified behaviour of weak declaration "#pragma weak".

Description
-----------
IAR EW release 8.50.5 causing build error in case of FreeRTOS  IAR ARM_CM0 port.

Error[Li006]: duplicate definitions for "vPortSetupTimerInterrupt"; in "c:test\boards\<someM0board>\rtos_examples\freertos_tickless\cm0plus\iar\debug\obj\fsl_tickless_lptmr.o", and "c:test\boards\someM0board\rtos_examples\freertos_tickless\cm0plus\iar\debug\obj\port.o"

To fix the issue use __weak in function definition instead.

Test Steps
-----------
Build FreeRTOS with ARM_CM0 port in IAR 8.50.5 with custom non-weak declaration of vPortSetupTimerInterrupt in application.

### DIFF
--- a/portable/IAR/ARM_CM0/port.c
+++ b/portable/IAR/ARM_CM0/port.c
@@ -96,7 +96,6 @@ static UBaseType_t uxCriticalNesting = 0xaaaaaaaa;
  * file is weak to allow application writers to change the timer used to
  * generate the tick interrupt.
  */
-#pragma weak vPortSetupTimerInterrupt
 void vPortSetupTimerInterrupt( void );
 
 /*
@@ -242,7 +241,7 @@ void xPortSysTickHandler( void )
  * Setup the systick timer to generate the tick interrupts at the required
  * frequency.
  */
-void vPortSetupTimerInterrupt( void )
+__weak void vPortSetupTimerInterrupt( void )
 {
     /* Calculate the constants required to configure the tick interrupt. */
     #if ( configUSE_TICKLESS_IDLE == 1 )


DavidJurajdaNXP:
I would like to point out there are currently several ways how weak functions are implemented in IAR ports:
CM0
#pragma weak vPortSetupTimerInterrupt
void vPortSetupTimerInterrupt( void );

CM3
__weak void vPortSetupTimerInterrupt( void )

CM33
_attribute_(( weak )) void vPortSetupTimerInterrupt( void ) /*

CM4
__weak void vPortSetupTimerInterrupt( void )

It would be good to select one method.

DavidJurajdaNXP:
Here are comments provided by IAR:

COMMENT 1:
This is a known fact.
"_weak ", and I assume "#pragma weak", is not equivalent to "attribute_((weak))", which it used to be. This was fixed in 8.50.5.
According to linker expert  "_weak" can mean "weak definition" or "weak reference". "attribute_((weak))" mean weak definition, which is the wanted behavior here, I assume.
I have pushed a change to CMSIS_5 where the behavior of the macro "_WEAK" changes from "Pragma("object_attribute = __weak")" to "attribute_((weak))".
I ran into this issue with TF-M after upgrading to 8.50.5.
I guess FreeRTOS may need to be updated.

COMMENT 2:

We are not finished with the investigation on our side but I got some more info.
It is interesting since it may mean that the fix can be very simple on your side.

Example:
main.c:
—
#include <stdio.h>

#pragma weak foo
int foo(void);

int foo(void)
{   return 17; }

int main()
{   printf("foo %d\n", foo());   return 0; }

—

foo.c:
—
int foo(void)
{   return 42; }

—

Building this project results in:
—
Error[Li006]: duplicate definitions for "foo"; in "C:\Users\thomasto\Projects\8.50\weak\Debug\Obj\foo.o", and "C:\Users\thomasto\Projects\8.50\weak\Debug\Obj\main.o"
Error while running Linker
 
Total number of errors: 1
Total number of warnings: 0
 
Build failed.
—

If line 4 of main.c is removed/commented out it works as expected, so in the NXP example, removing line 100 would likely solve the issue.

(line:100 is referenced to file: https://github.com/FreeRTOS/FreeRTOS-Kernel/blob/210b1ffcc87bcff93871a37fbf0ad2033870ecaf/portable/IAR/ARM_CM0/port.c)

yuhui-zheng:
I'm using this M0 demo https://github.com/FreeRTOS/FreeRTOS/tree/master/FreeRTOS/Demo/CORTEX_M0%2B_LPC51U68_GCC_IAR_KEIL, and I can get ```Error[Li006]: duplicate definitions for "vPortSetupTimerInterrupt";``` if to override ```vPortSetupTimerInterrupt``` in application code. 

@aggarg We may need to pick a style.

PR #92, opened 2020-07-17
Update kernel files for ESP32 port

shubhamkulkarni97:
Description
-----------
Update kernel files with latest changes in ESP-IDF.

These changes are required for updating ESP-IDF to v4.2 release in Amazon FreeRTOS.

Changes required to add support for ESP32-S2 in Amazon FreeRTOS are also added.

### DIFF
--- a/portable/ThirdParty/GCC/Xtensa_ESP32/include/portmacro.h
+++ b/portable/ThirdParty/GCC/Xtensa_ESP32/include/portmacro.h
@@ -305,12 +305,32 @@
                                              uint32_t compare,
                                              uint32_t * set )
         {
-            __asm__ __volatile__ (
-                "WSR 	    %2,SCOMPARE1 \n"
-                "S32C1I     %0, %1, 0	 \n"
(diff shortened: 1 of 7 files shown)


mahavirj:
@yuhui-zheng Can you please take a look at this PR? Also per our earlier discussion at https://github.com/FreeRTOS/FreeRTOS-Kernel/pull/9#issuecomment-582513244, should we consider moving `port` files to `portable/Espressif/Xtensa` directory (since with this PR we are also adding support for ESP32-S2 silicon, corresponding PR https://github.com/aws/amazon-freertos/pull/2285)?

yuhui-zheng:
@mahavirj  Noted. Will either do it myself or get some help from my colleagues. 

Regarding to directory change, some thoughts:
We may need to sync with Amazon FreeRTOS. Since the directory change here may break the builds in Amazon FreeRTOS, and the release cycle is slightly different. I will also check what's our criteria to move a port out of third party.

PR #94, opened 2020-07-19
Fix deadlock when writing more bytes to stream buffer than it's size

simonbeaudoin0935:
xStreamBufferSend deadlock avoidance when sending too much data

Description
-----------
Add necessary checks when sending data to the stream/message buffer in order to avoid a task deadlock when attempting to write a longer stream/message than the underlying buffer can write.

In the case of the stream buffer, if the user tries to write more bytes than supported, the added code will truncate the stream and adapt the 'xRequiredSpace' variable to the size of the buffer. The user will be able to deal with this scenario by wrapping the call to 'xStreamBufferSend' in appropriate looping code. 

In the case of the message buffer, as opposed to stream buffer, there is no choice but to write the whole message (plus a uint32_t for it's length). In this case, the added code will do that 0 will be returned. 

A possible touch up to this pull request would be an added error return code for that scenario. 

Nevertheless, the deadlock is avoided in both case.

### DIFF
--- a/stream_buffer.c
+++ b/stream_buffer.c
@@ -518,7 +518,11 @@ size_t xStreamBufferSend( StreamBufferHandle_t xStreamBuffer,
     size_t xReturn, xSpace = 0;
     size_t xRequiredSpace = xDataLengthBytes;
     TimeOut_t xTimeOut;
-
+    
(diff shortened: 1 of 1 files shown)


simonbeaudoin0935:
> This change only affects the situation where too much is written to the buffer. All current working code should be unaffected by this change.

Exactly

PR #99, opened 2020-07-21
Synopsys ARC update, with updated BSP support

YuguoWH:
Description
-----------
embARC BSP has been updated, so the code here need a update too. 
Also remove an unused macro which could have caused warnings.

Something new to discuss
-----------------------------
- We have prepared two demos for customers to get hands on but not sure where to put the code. Please kindly advise. 

- We also have new supports for ARC v1 series processors, which are not in this PR. 
----------------

### DIFF
--- a/portable/ThirdParty/GCC/ARC_EM_HS/freertos_tls.c
+++ b/portable/ThirdParty/GCC/ARC_EM_HS/freertos_tls.c
@@ -54,6 +54,7 @@
     extern void __mw_run_tls_dtor();
     extern void __mw_run_tls_ctor();
 
+extern uint32_t exc_nest_count;
 /*
  * Linker generated symbols to mark .tls section addresses
  * first byte .. last byte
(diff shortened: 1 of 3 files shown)


yuhui-zheng:
@YuguoWH Could submit PR to https://github.com/FreeRTOS/FreeRTOS repo any moment you are ready. 

For port demos, the good directory to use is [```./FreeRTOS/FreeRTOS/Demo/```](https://github.com/FreeRTOS/FreeRTOS/tree/master/FreeRTOS/Demo), and folder naming convention would be somewhat like ```<architecture>_<board>_<IDE/compiler>```. A readme.md file associated with demo project would be great. Things like board info, what the demo does, how to configure demo, how to run the demo, and how to reach out to you for support... these are good to go into the readme.md. 


Kindly note that we may not have the board to evaluate your demo. Simulator setup would help us without hardware... Or we could get board from you if they are in production. 


Thank you!

PR #121, opened 2020-08-14
added a warning concerning the incomplete testing of the RX700v3_DFPU…

n9wxu:
RX700v3_DFPU port warning

Description
-----------
We have a new port submission for the RX700v3 with the DFPU.  New ports are fantastic but this port has not been tested to the same rigor as other ports.  In order to accept the port more quickly, we are adding a warning to this port to ensure developers understand that this port is not as tested as other ports.

As soon as the testing can be completed, this warning will be removed.


NoMaY-jp:
Thank you for your taking care. By the way, are there tasks what I can help you? I think that you are very busy, on the other hand I'm not so busy. That is the reason why I think so.

n9wxu:
We need register context switch test support.  If you look at 
[regtest](https://github.com/FreeRTOS/FreeRTOS/blob/master/FreeRTOS/Demo/RX600_RX64M_RSK_GCC_e2studio/src/RegTest.S
)
you will seen an example of some of what we need.  We are trying to get a more comprehensive set of tests supports in the demo's for each port.  This will be restructured and formalized in the coming months but that lack was why we marked the DFPU port as experimental.

PR #148, opened 2020-08-28
Revert "Fix: Add Parenthesis around if-statement in macro"

alfred2g:
Reverts FreeRTOS/FreeRTOS-Kernel#138


cobusve:
We are rolling back this one because it was causing an inconsistency throughout the code base, we will revisit it and see if it is feasible to make this a global change instead of just this one port.

PR #162, opened 2020-09-09
Fix: Comment - xTaskIncrementTick loop - to adhere to demo requirement

alfred2g:
Comment - xTaskIncrementTick loop - to adhere to demo requirement

Description
-----------
Commenting xTaskIncrementTick loop to adhere to the demo requirement

Test Steps
-----------
full demo run
```
[Posix_GCC]$ ./build/posix_demo

Trace started.
The trace will be dumped to disk if a call to configASSERT() fails.

The trace will be dumped to disk if Enter is hit.
Starting full demo
OK: No errors - tick count 10000
OK: No errors - tick count 20000
OK: No errors - tick count 30000
```

### DIFF
--- a/portable/ThirdParty/GCC/Posix/port.c
+++ b/portable/ThirdParty/GCC/Posix/port.c
@@ -366,12 +366,15 @@ uint64_t xExpectedTicks;
 
 	/* Tick Increment, accounting for any lost signals or drift in
 	 * the timer. */
-	xExpectedTicks = (prvGetTimeNs() - prvStartTimeNs)
-		/ (portTICK_RATE_MICROSECONDS * 1000);
-	do {
+/*
+ *      Comment code to adjust timing according to full demo requirements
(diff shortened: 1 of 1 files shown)


cobusve (approved):

0xjakob:
@alfred2g @cobusve What is the problem with the commented code, or what particular demo requirement is met by commenting the code out? In what situation would you un-comment it, i.e., use it again?

We're currently integrating the FreeRTOS simulation into our project ([ESP IDF](https://github.com/espressif/esp-idf)), but we don't know if we should use the commented code or not. In our own simple demo, everything seems to work well with and without the commented code.

alfred2g:
@0xjakob, I would use first the commented code, but, you can probably use it both ways.
The code was commented to satisfy the tests specific timings we have(Full Demo) , which use time to pass or fail the test.

0xjakob:
@alfred2g Thanks a lot for the information!

PR #181, opened 2020-09-22
Posix: Free the allocated memory after deleting a task or ending the scheduler

RedaMaher:
Description
-----------

In Posix Simulator on Linux, the memory allocated by idle and first task is not freed properly after ending the scheduler. Also the memory allocated for the condition variable is not freed in case of deleting a task from another task. This is needed to avoid memory leaks errors when memory checkers (Valgrind, address sanitizer, ..) are used. This pull request is solving this issue.

Test Steps
-----------

Use the FreeRTOS Posix Simulator on Linux and run it under Valgrind or use address sanitizer. Make sure no memory leaks errors after deleting a task from another task (not just a task deletes itself) and after ending the scheduler.

Related Issue
-----------

### DIFF
--- a/portable/ThirdParty/GCC/Posix/port.c
+++ b/portable/ThirdParty/GCC/Posix/port.c
@@ -178,12 +178,15 @@ portBASE_TYPE xPortStartScheduler( void )
 {
 int iSignal;
 sigset_t xSignals;
+TaskHandle_t pxIdleThread;
 
 	hMainThread = pthread_self();
 
 	/* Start the timer that generates the tick ISR(SIGALRM).
 	   Interrupts are disabled here already. */
 	prvSetupTimerInterrupt();
+
+	pxIdleThread = xTaskGetIdleTaskHandle();
 
 	/* Start the first task. */
 	vPortStartFirstTask();
@@ -196,6 +199,9 @@ sigset_t xSignals;
 	{
 		sigwait( &xSignals, &iSignal );
 	}
+
+	/* Cancel the Idle task and free its resources */
+	vPortCancelThread(pxIdleThread);
 
 	/* Restore original signal mask. */
 	(void)pthread_sigmask( SIG_SETMASK, &xSchedulerOriginalSignalMask,  NULL );
--- a/portable/ThirdParty/GCC/Posix/port.c
+++ b/portable/ThirdParty/GCC/Posix/port.c
@@ -62,6 +62,7 @@
 /* Scheduler includes. */
 #include "FreeRTOS.h"
 #include "task.h"
+#include "timers.h"
 #include "utils/wait_for_event.h"
 /*-----------------------------------------------------------*/
 
@@ -178,15 +179,12 @@ portBASE_TYPE xPortStartScheduler( void )
 {
 int iSignal;
 sigset_t xSignals;
-TaskHandle_t pxIdleThread;
 
 	hMainThread = pthread_self();
 
 	/* Start the timer that generates the tick ISR(SIGALRM).
 	   Interrupts are disabled here already. */
 	prvSetupTimerInterrupt();
-
-	pxIdleThread = xTaskGetIdleTaskHandle();
 
 	/* Start the first task. */
 	vPortStartFirstTask();
@@ -201,7 +199,11 @@ TaskHandle_t pxIdleThread;
 	}
 
 	/* Cancel the Idle task and free its resources */
-	vPortCancelThread(pxIdleThread);
+	vPortCancelThread( xTaskGetIdleTaskHandle() );
+#if ( configUSE_TIMERS == 1 )
+	/* Cancel the Timer task and free its resources */
+	vPortCancelThread( xTimerGetTimerDaemonTaskHandle() );
+#endif /* configUSE_TIMERS */
 
 	/* Restore original signal mask. */
 	(void)pthread_sigmask( SIG_SETMASK, &xSchedulerOriginalSignalMask,  NULL );
--- a/portable/ThirdParty/GCC/Posix/port.c
+++ b/portable/ThirdParty/GCC/Posix/port.c
@@ -411,6 +411,7 @@ Thread_t *pxThreadToCancel = prvGetThreadFromTask( pxTaskToDelete );
 	 */
 	pthread_cancel( pxThreadToCancel->pthread );
 	pthread_join( pxThreadToCancel->pthread, NULL );
+	event_delete( pxThreadToCancel->ev );
 }
 /*-----------------------------------------------------------*/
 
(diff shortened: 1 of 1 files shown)


portable/ThirdParty/GCC/Posix/port.c
```diff
@@ -197,6 +202,14 @@ sigset_t xSignals;
...
 	}
 
+	/* Cancel the Idle task and free its resources */
+	vPortCancelThread(pxIdleThread);
+
+	/* Cancel the first task and free its resources */
+	pthread_cancel( pxFirstThread->pthread );
+	pthread_join( pxFirstThread->pthread, NULL );
+	event_delete(pxFirstThread->ev);
+
```
alfred2g, line 212: There is no guarantee that the first thread is still running at this point

alfred2g: Reaching this point means the program is ending, the OS will clean all the resources

RedaMaher: > There is no guarantee that the first thread is still running at this point

Oops! Yes you are right

> Reaching this point means the program is ending, the OS will clean all the resources

Where will the OS clean the resources?
Valgrind is still giving memory leak errors when I removed the incorrect first task freeing above. I called `vTaskDelete(firstTaskHandle);` from the main after ending the scheduler. But the resources were not freed also because the OS thinks that the current task is the task I want to delete.
Should this condition in `vTaskDelete`
`if( pxTCB == pxCurrentTCB )` changed to `if( ( pxTCB == pxCurrentTCB ) && ( xSchedulerRunning != pdFALSE ) )` to direct `vTaskDelete` to free the resources immediately as the idle task will not come to do it?

alfred2g: After Valgrind returns control to the OS, resources will be cleaned
which makes valgrind report leaks for that memory as it stand between the program and the OS (Java virtual machine style)

Also if the program reaches this point without calling vTaskDelete on some other tasks, lets say 10 , it will only clean one task out of the 10 and exit , and Valgrind will still show leaks 

I think this part of the code is not important to clean, as we are exiting the program, and keeping track of tasks here into a list or array, is just wasteful

RedaMaher: Yes, it is clear for me that Linux will clean all resources after the process finishes (I thought you are talking about FreeRTOS). 
It is the application responsibility to clean any resources allocated by its tasks. While it is the RTOS responsibility to clean the resources allocated by the tasks created by it (like Timer and Idle threads).
I agree with you that it is not super important to clean all resources after ending the scheduler. Ending the scheduler is not a common case on the real embedded systems. But in my case, I am running automated unit tests on the Linux Host. So it is important to exit without Valgrind errors to pass the tests.

I think vTaskDelete should clean the task resources immediately if the scheduler is not running (I may create another Pull request for it).

cobusve: Hmmm I am trying to follow the logic here but it does not make sense to me. 

When we exit from the scheduler FreeRTOS (the operating system itself) has died, this is not like ending a process in Linux, this is literally like ending Linux itself. 

In the operating system, we want to make sure that we do not waste memory or code space as these are scarce resources. Adding code to clean up memory while the machine is rebooting seems to be very wasteful?

RedaMaher: @cobusve This is just a FreeRTOS simulator runs on top of a Linux system (not a complete OS controls the machine). So it is literally a process (with multiple threads) running on top of the Linux OS. The Linux is securing itself and will free any resources allocated by the process (FreeRTOS Simulator) after it exits. But from the point of unit tests and memory checks, we are doing something wrong. This is my case and I think it will apply to many of the FreeRTOS simulator users.
On the normal Linux machines, the resources are not scarce, so it will be better if we can do things correctly.

BTW, this part of the PR is outdated as no guarantee that the first task is still running at this time as mentioned in the previous comments by @alfred2g

cobusve: I agree that it would be more ideal to clean everything up nicely when the simulator is stopping but I think this poses more challenges than you may think. For example, some threads may be holding onto a mutex at this point in time and will not be at a cancellation point (this means attempts to cancel the thread will fail). Further, any memory allocated on the heap will not be identifiable so cleaning it up will not be practical.

Consider also the case where a thread is in the process of allocating new memory on the heap, the heap is at that ping in an inconsistent state, and the scheduler may be stopped at this point from an interrupt. It is important that the changes we make consider each and every one of these cases.

For many of these reasons the decision was made to just let Linux do its job in cleaning up when the process is being killed.

We would gladly accept any PR that improves the cleanup process for the simulator as long we are careful not to change FreeRTOS for the worse to try to make the simulator better.

cobusve: BTW. If I think one way to do this is to walk the list of TCB's and check the pthread structure at the top of the stack for each FreeRTOS task. You would have to be careful because there is no guarantee that each of these threads are still running, but you could attempt to cancel each thread, and if it is not possible to cancel it because it is not at a cancellation point, you would have to just give up on that one and move on to the next one. It may be necessary to add a state variable to the thread structure to track which ones are running and which ones are not.

I think the biggest problem to overcome is that anything that makes the scheduler exit like that leaves the FreeRTOS TCB structures in a state that will probably not be guaranteed to be consistent, and if that is the case it could fail in very exotic ways when e.g. the TCB linked list was in the process of being updated and you somehow kill the scheduler from an ISR it may not be possible to clean up reliably at all.

I also think the heap is a lost cause, I think we will just have to make peace with the fact that memory allocated on the heap will be cleaned up by Linux.

RedaMaher: @cobusve I agree with you that we should not add any burden on the FreeRTOS to make the simulator better. Also, we should not complicate the simulator itself. We need to differentiate between two points (The user responsibilities and the FreeRTOS responsibilities). The first point, the user should select the appropriate time to end the scheduler. The second point, if the user ended the scheduler at the correct time, the FreeRTOS should exit cleanly. In case the user did not select the appropriate time, then it is not the FreeRTOS job to care about the inconsistencies.

Anyway, now this PR is just cleaning the resources of the Idle task (if the scheduler was ended) and the task condition variable (if a task is deleted from another task). So no changes in FreeRTOS itself.

RedaMaher: I finally noticed that the PR title is misleading so I updated it

cobusve (dismissed):

portable/ThirdParty/GCC/Posix/port.c
cobusve: That looks good and safe to me. 

What do you think about also cleaning up the timer task? You can get the timer thread handle by calling xTimerGetTimerDaemonTaskHandle just like you were getting the idle task handle above.

RedaMaher: I cleaned the timer task as well and refined the fix by cleaning the condition variable in the correct place.

alfred2g (approved):

cobusve (approved):

PR #183, opened 2020-09-23
Add support for RV32E core for GCC compiler and add support for FPU (simple and double precision) for GCC compiler

e-puerto:
add support for RV32E core for GCC compiler
add support for FPU (simple and double precision) for GCC compiler
Description
Change vPortSetupTimerInterrupt in order to support 64bits RISC-V and modify mtime writing on 32bits RISC-V
Change portASM.s in order to support RV32E (16 registers instead of 32 registers on RV32I)
Change portASM.s in order to save/restore FPU registers on simple/double precision

Test Steps
tested on ARTY board with SIFIVE standard core E31, U54, S54 plus custom core with FPU (simple/double precision)
use "example-freertos-blinky" from freedom-e-sdk from SIFIVE (This change have not been yet committed into our public release, do not hesitate to communicate with me if you want to test it).

Related Issue
NONE

### DIFF
--- a/portable/GCC/RISC-V/port.c
+++ b/portable/GCC/RISC-V/port.c
@@ -166,9 +166,9 @@ task stack, not the ISR stack). */
     #if( __riscv_xlen == 64 )
         void vPortSetupTimerInterrupt( void )
         {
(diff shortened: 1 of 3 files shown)


e-puerto:
Issue with rev32 and FPU double precision
need to fix.

PR #184, opened 2020-09-24
Posix: Fix no task switching issue if a task ended its main function

RedaMaher:
Description
-----------

When the main function of a task exits, no task switching happened.
This is because all the remaining tasks are waiting on the condition
variable. The fix is to trigger a task switch and mark the exiting
task as "Dying" to be suspened and exited properly from the scheduler.

Test Steps
-----------

Create an application with multi tasks. Make one task to exit normally from its main function. Then no task switching will happen.

Related Issue
-----------

### DIFF
--- a/portable/ThirdParty/GCC/Posix/port.c
+++ b/portable/ThirdParty/GCC/Posix/port.c
@@ -66,6 +66,7 @@
 /*-----------------------------------------------------------*/
 
 #define SIG_RESUME SIGUSR1
+#define portTASK_ENDED_DELAY ( 500 )
 
 typedef struct THREAD
 {
@@ -420,6 +421,14 @@ Thread_t *pxThread = pvParams;
 
 	/* Call the task's entry point. */
 	pxThread->pxCode( pxThread->pvParams );
+
+	/* The task has finished, we need to trigger a task switch here
+	 * to avoid exiting while all tasks are waiting to be signaled.
+	 * So we will mark the task as Dying here then delay the task
+	 * with any value to trigger a task switch where the task will
+	 * be suspended and exited. */
+	pxThread->xDying = pdTRUE;
+	vTaskDelay( portTASK_ENDED_DELAY );
 
 	return NULL;
 }
--- a/portable/ThirdParty/GCC/Posix/port.c
+++ b/portable/ThirdParty/GCC/Posix/port.c
@@ -67,7 +67,6 @@
 /*-----------------------------------------------------------*/
 
 #define SIG_RESUME SIGUSR1
-#define portTASK_ENDED_DELAY ( 500 )
 
 typedef struct THREAD
 {
@@ -418,6 +417,24 @@ Thread_t *pxThreadToCancel = prvGetThreadFromTask( pxTaskToDelete );
 }
 /*-----------------------------------------------------------*/
 
+static void prvTaskExitError( void )
+{
+	/* A function that implements a task must not exit or attempt to return to
+	* its caller as there is nothing to return to. If a task wants to exit it
+	* should instead call vTaskDelete( NULL ). Artificially force an assert()
+	* to be triggered if configASSERT() is defined, then stop here so
+	* application writers can catch the error. */
+	configASSERT( pdFALSE );
+
+	portDISABLE_INTERRUPTS();
+
+	while( 1 )
+	{
+		/* Do Nothing */
+	}
+}
+/*-----------------------------------------------------------*/
+
 static void *prvWaitForStart( void * pvParams )
 {
 Thread_t *pxThread = pvParams;
@@ -431,13 +448,8 @@ Thread_t *pxThread = pvParams;
 	/* Call the task's entry point. */
 	pxThread->pxCode( pxThread->pvParams );
 
-	/* The task has finished, we need to trigger a task switch here
-	 * to avoid exiting while all tasks are waiting to be signaled.
-	 * So we will mark the task as Dying here then delay the task
-	 * with any value to trigger a task switch where the task will
-	 * be suspended and exited. */
-	pxThread->xDying = pdTRUE;
-	vTaskDelay( portTASK_ENDED_DELAY );
+	/* Task function should not return */
+	prvTaskExitError();
 
 	return NULL;
 }
--- a/portable/ThirdParty/GCC/Posix/port.c
+++ b/portable/ThirdParty/GCC/Posix/port.c
@@ -417,24 +417,6 @@ Thread_t *pxThreadToCancel = prvGetThreadFromTask( pxTaskToDelete );
 }
 /*-----------------------------------------------------------*/
 
-static void prvTaskExitError( void )
-{
-	/* A function that implements a task must not exit or attempt to return to
-	* its caller as there is nothing to return to. If a task wants to exit it
-	* should instead call vTaskDelete( NULL ). Artificially force an assert()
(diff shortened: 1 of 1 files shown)


portable/ThirdParty/GCC/Posix/utils/wait_for_event.c
```diff
@@ -21,11 +21,17 @@ struct event * event_create()
...
-    free( ev );
+    struct event * ev = *ev_ptr;
+
+    if ( ev )
+    {
+        pthread_mutex_destroy( &ev->mutex );
+        pthread_cond_destroy( &ev->cond );
+        free( ev );
+        *ev_ptr = NULL;
+    }
```
alfred2g, line 34: This would hide the doube deletion of an even if called twice.
We would like it to fail as soon as possible, as it could make a design issue not apparent by trying to delete a task twice.
This is similar to how free works, if you call free twice on a pointer the second time it would crash instead of checking the validity of that pointer

RedaMaher: Yes, you are correct. This will hide the double deletion. I added this check to be able to work with the fix in PR #181 when `event_delete` where called from two places. But now this is not needed after refining the fix of PR #181 by deleting the condition variable in the correct place (When a thread is canceled).

alfred2g (approved):

portable/ThirdParty/GCC/Posix/port.c
```diff
@@ -421,6 +422,14 @@ Thread_t *pxThread = pvParams;
 	/* Call the task's entry point. */
 	pxThread->pxCode( pxThread->pvParams );
 
+	/* The task has finished, we need to trigger a task switch here
+	 * to avoid exiting while all tasks are waiting to be signaled.
+	 * So we will mark the task as Dying here then delay the task
+	 * with any value to trigger a task switch where the task will
+	 * be suspended and exited. */
+	pxThread->xDying = pdTRUE;
+	vTaskDelay( portTASK_ENDED_DELAY );
```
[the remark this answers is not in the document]
RedaMaher: It will return to the same task when there are no other tasks at a higher or equal priority. So I get stuck again.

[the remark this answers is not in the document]
alfred2g: The main issue is the delay
Why was it set to 500 ? would it work with 200? what about 100? etc
Can we make it so it doesn't wait at all?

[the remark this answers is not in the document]
RedaMaher: The delay can be with any value as I described it in the comment. The task actually will not wait at all. Once the task switching is in place, the task will exit as it is marked as "Dying". The purpose of the delay is to trigger an actual task switching and call `prvSwitchThread` with different task handles.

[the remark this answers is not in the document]
cobusve: Hmmm, I am still confused here. It seems like we are talking about the case where a FreeRTOS task function returns. By design when this happens the stack will be corrupted and the OS will crash, so what are we trying to do here?

I am trying to understand under which circumstances this code is helping us.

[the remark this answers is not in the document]
RichardBarry: A task should call vTaskDelete( NULL ) at its end.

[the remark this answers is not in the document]
RedaMaher: @RichardBarry The task resources will be freed in the idle task when the task deletes itself. This does not handle the following case:
1. Task (x) allocated memory to be used for a new higher priority static task (y).
2. Task (x) created a static task (y) and used the allocated memory from the previous step.
3. Task (y) scheduled then finished and deleted itself.
4. Scheduler returns to Task (x) which was waiting for Task (y) to finish.
5. Task (x) frees the allocated resources.
6. When The Idle task runs, it will try to free the resources of Task (y) and access a freed memory (and causes a Memory Corruption)

So I moved to delete Task (y) from the Task(x) and handle the task exit properly.
There is a workaround for this issue by adding `vTaskDelay` in Task (x) (before step 5) to give a chance for the idle task to run and free the resources of Task (y).

@cobusve Yes, I was trying to handle the case where the task function finishes. And It looks like it is not the proper way by design. I got confused because the task exit silently in Posix port without assertion or crashes. I will update the PR to just assert in this case to be easy to catch.

alfred2g (dismissed):

portable/ThirdParty/GCC/Posix/port.c
```diff
@@ -430,6 +448,9 @@ Thread_t *pxThread = pvParams;
 	/* Call the task's entry point. */
 	pxThread->pxCode( pxThread->pvParams );
 
+	/* Task function should not return */
+	prvTaskExitError();
```
cobusve, line 452: I think the normal way to handle asserts is like this https://github.com/FreeRTOS/FreeRTOS/blob/6cc5310f380524c7885181a3be0c60ead4b59a22/FreeRTOS/Demo/Posix_GCC/main.c#L283

That means you probably just have to call configASSERT( pdFALSE );  over here and be done.

RedaMaher: @cobusve This is exactly what `prvTaskExitError();` is doing in addition to stop if `configASSERT()` is not actually configured. This is useful to cover the issue of task ending in FreeRTOS and be easy to catch.
The same function is used in many ports (For example https://github.com/FreeRTOS/FreeRTOS-Kernel/blob/master/portable/GCC/ARM_CM3/port.c#L194)

alfred2g: In some port it is defined by the demo/application, and in some ports it is defined by the port
in this case this function is defined by the demo application in: https://github.com/FreeRTOS/FreeRTOS/blob/master/FreeRTOS/Demo/Posix_GCC/main.c#L283

RedaMaher: Sorry I did not get what is the issue here. If Posix port **can** be used without defining `configASSERT` then the stop loop is needed. If Posix port **can not** be used without defining `configASSERT` then no need for the stop loop but we need to write this down in some documentation.

alfred2g: https://www.freertos.org/a00110.html#configASSERT
As you can see in the doc, config assert is an application side setting not a port side setting, as the user has the option of turning it on or off depending on the maturity of the application in order to save space.

RedaMaher: This is exactly what I am talking about. So that the stop loop in `prvTaskExitError` after config assert is important to avoid dead silent stop when config assert is not set in the application side

alfred2g: If the user is confident about his application, and is sure that they called vTaskDelete at the end of every task, extra code and code size is not desired, so vConfigAssert will rightfully be disabled to remove the extra code. By adding the prvTaskExitError function this would not be possible

Worth mentioning, this is a linux port and extra code is not as important, but to stay consistent, with other new ports,  be as true as possible to an embedded environment, and provide a good example.

alfred2g (approved):

cobusve (approved):

portable/ThirdParty/GCC/Posix/port.c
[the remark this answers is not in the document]
RedaMaher: @RichardBarry @cobusve  any suggestion on how to handle the above scenario?

PR #185, opened 2020-09-25
Consequent TickType_t  stack types

feilipu:
Consequent usage of `TickType_t` and `configSTACK_DEPTH_TYPE`.
-----------
Currently `uint32_t` is used in several places, where more accurately`TickType_t` and `configSTACK_DEPTH_TYPE` should be used. The correction is relevant in ports that have either 16 bit or 64 bit tick timers or 16 bit stack depths.

Test Steps
-----------
Improved definition.
Optimisation where 16 bit variables and parameters can be handled rather than 32 bit variables and parameters.

-----------


feilipu:
Closing this, as `main` has diverged substantially, so merge is difficult.

Will reissue a PR to address the `configSTACK_DEPTH_TYPE` issue.

Issue #33, opened 2020-03-17
[BUG] Re-introduce uxTopUsedPriority for OpenOCD Thread Support


dereks:
OpenOCD is a popular Open Source On-Chip Debugger:

http://openocd.org/

One of its best features is thread support (Task support). With OpenOCD, the GNU **gdb** debugger can view the list of Tasks, and can look at an individual Task to see its stack trace, step through code in just that one Task, etc Basically, OpenOCD enables these gdb features:

https://ftp.gnu.org/old-gnu/Manuals/gdb/html_node/gdb_24.html

(Eclipse also has desktop GUI support for walking through OpenOCD Tasks, much like a normal POSIX threaded program.)

Without thread support, gdb is very restricted and hard to use with a FreeRTOS application. For example, constant SysTick interrupts keep jumping you into the scheduler, making it very hard to follow your application code. It's basically only useful for debugging `main()` and ISRs.

In an older release of FreeRTOS (V7.5.3) there was a symbol removed called `uxTopUsedPriority`. That symbol is necessary for OpenOCD to find the list of Tasks and enable thread support.

Since that release, we developers have had to manually re-introduce that symbol so that we can debug our FreeRTOS apps. Usually this means adding a copy of this file to the build:

https://raw.githubusercontent.com/arduino/OpenOCD/master/contrib/rtos-helpers/FreeRTOS-openocd.c

Here is that entire file:

```
/*
 * Since at least FreeRTOS V7.5.3 uxTopUsedPriority is no longer
 * present in the kernel, so it has to be supplied by other means for
 * OpenOCD's threads awareness.
 *
 * Add this file to your project, and, if you're using --gc-sections,
 * ``--undefined=uxTopUsedPriority'' (or
 * ``-Wl,--undefined=uxTopUsedPriority'' when using gcc for final
 * linking) to your LDFLAGS; same with all the other symbols you need.
 */

#include "FreeRTOS.h"

#ifdef __GNUC__
#define USED __attribute__((used))
#else
#define USED
#endif

const int USED uxTopUsedPriority = configMAX_PRIORITIES - 1;
```

Some distributions of FreeRTOS, like WICED SDK, already put this back for you.

The Internet is littered with forum posts of developers asking why they can't debug FreeRTOS Tasks, and responses telling them about this annoying requirement. Just google "uxTopUsedPriority" to see what I'm talking about.

[...]

PR #188, opened 2020-09-30
OpenOCD Support: Re-introduce uxTopUsedPriority

dachalco:
Description
-----------
Re-introduce `uxTopUsedPriority` for use by OpenOCD. 

The kernel tracks _Ready_ tasks via priority lists. For tracking the state of remote threads, OpenOCD uses `uxTopUsedPriority` to determine the number of priority lists to read back from the remote target. 

See OpenOCD's [FreeRTOS.c](https://sourceforge.net/p/openocd/code/ci/master/tree/src/rtos/FreeRTOS.c)

Test Steps
-----------
Build any project and verify the presence of `uxTopUsedPriority` in the executable's symbol table. 

To exercise OpenOCD awareness of FreeRTOS threads, add `-rtos FreeRTOS` in your call to `configure` within your `openocd.cfg` config file. 

**example-openocd.cfg** (snippet):
```cfg
$_TARGETNAME.0 configure -rtos FreeRTOS -work-area-phys 0x80000000 -work-area-size 0x4000 -work-area-backup 0
```

Note, OpenOCD does not support `-rtos FreeRTOS` for all its architectures. See [FreeRTOS.c::FreeRTOS_params_list](https://sourceforge.net/p/openocd/code/ci/master/tree/src/rtos/FreeRTOS.c#l55)

Related Issue
-----------

- Issue #33 

-----------

### DIFF
--- a/tasks.c
+++ b/tasks.c
@@ -378,6 +378,7 @@ PRIVILEGED_DATA static volatile BaseType_t xNumOfOverflows = ( BaseType_t ) 0;
 PRIVILEGED_DATA static UBaseType_t uxTaskNumber = ( UBaseType_t ) 0U;
 PRIVILEGED_DATA static volatile TickType_t xNextTaskUnblockTime = ( TickType_t ) 0U; /* Initialised to portMAX_DELAY before the scheduler starts. */
 PRIVILEGED_DATA static TaskHandle_t xIdleTaskHandle = NULL;                          /*< Holds the handle of the idle task.  The idle task is created automatically when the scheduler is started. */
+const UBaseType_t uxTopUsedPriority = configMAX_PRIORITIES - 1U;
 
 /* Context switches are held pending while the scheduler is suspended.  Also,
  * interrupts must not manipulate the xStateListItem of a TCB, or any of the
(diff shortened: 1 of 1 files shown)


tasks.c
```diff
@@ -378,6 +378,7 @@ PRIVILEGED_DATA static volatile BaseType_t xNumOfOverflows = ( BaseType_t ) 0;
 PRIVILEGED_DATA static UBaseType_t uxTaskNumber = ( UBaseType_t ) 0U;
 PRIVILEGED_DATA static volatile TickType_t xNextTaskUnblockTime = ( TickType_t ) 0U; /* Initialised to portMAX_DELAY before the scheduler starts. */
 PRIVILEGED_DATA static TaskHandle_t xIdleTaskHandle = NULL;                          /*< Holds the handle of the idle task.  The idle task is created automatically when the scheduler is started. */
+const UBaseType_t uxTopUsedPriority = configMAX_PRIORITIES - 1U;
```
RichardBarry, line 381: Can you please add a comment to say why it is here - otherwise it might be confusing to see a const variable that is apparently not used.  It can be like the comment added the variable is cast to void.

RichardBarry (changes requested):

tasks.c
dachalco: Of course. Added in latest commit

cobusve (approved):

RichardBarry (approved):

PR #192, opened 2020-10-01
Fix: interrupt enabling at the end of vTaskExitCritical

kuopinghsu:
Fix: interrupt enabling at the end of vTaskExitCritical

Description
-----------
When xSchedulerRunning is false interrupts are always disabled in vTaskExitCritical. Adds portENABLE_INTERRUPTS() in vTaskExitCritical when scheduling is false.

Test Steps
-----------
Call malloc() before scheduling, the interrupt will be disabled.

Related Issue
-----------
N/A

### DIFF
--- a/tasks.c
+++ b/tasks.c
@@ -4367,7 +4367,7 @@ static void prvResetNextTaskUnblockTime( void )
         }
         else
         {
-            mtCOVERAGE_TEST_MARKER();
+            portENABLE_INTERRUPTS();
         }
     }
 


RichardBarry:
It is a deliberate policy to keep interrupts disabled between the first FreeRTOS API call made before the scheduler is started and the first task starting to run after the scheduler is started.  This stems from the early days of FreeRTOS where a lot of support questions arose from interrupts trying to use the FreeRTOS scheduler before the scheduler was started - which normally resulted in hard to debug crashes.  This is also why the critical nesting variables are never initialised to zero in the C start up code, but only just before the scheduler is started.  If it were initialised to zero interrupts would get enabled too soon.

PR #195, opened 2020-10-04
Fix data lengths caps used within xStreamBufferSend().

RichardBarry:
Description
-----------
A recent contribution capped the data length passed into xStreamBufferSend() to the capacity of the buffer.  The intent is to stop a task waiting out any specified block time (which could be indefinite) in cases where the task is attempting to send more data than the buffer can actually hold.  However the value used for capping the length was wrong, making the contribution ineffectual.  This PR corrects the length used as the cap value.

Test Steps
-----------
FreeRTOS/Semo/Common/Minimal/StreamBufferDemo.c and FreeRTOS/Semo/Common/Minimal/MessageBufferDemo.c have both been updated (submitted via a separate PR) to test sending data lengths that get capped.  Note the tests already tested for this use case, but did not test with infinite block times, and the code didn't prevent an infinite block time being used.


cobusve (approved):

simonbeaudoin0935:
Nice catch

PR #201, opened 2020-10-17
RoundRobin in task with same priority when configUSE_TIME_SLICING = 0

dougnobrega:
Fix: RoundRobin in task with same priority when configUSE_TIME_SLICING = 0
---------------------------------------------------------------------------------------------------------

Description:
---------------------------------------------------------------------------------------------------------
As we know, there are several possible scheduling policies. Some of them FIFO, Round Robin, Fixed-priority preemptive and others. By default, freeRTOS uses the Prioritized Preemptive Scheduling with Time Slicing policy. However, it is possible to modify the scheduler policy with few changes in code. Theoretically, it is only necessary to change two defines, namely configUSE_PREEMPTION and configUSE_TIME_SLICING.

However, if by chance we wanted to use the Fixed-priority preemptive policy, we would have to allow a change of context only when a higher priority task appears, that is, tasks of equal priorities will not share processing time. This was to be configured with configUSE_TIME_SLICING. Therefore, we need our scheduler to be preemptive and not always go to the next task if they have the same priority.

Therefore, we can check the information provided by freeRTOS. Following is an excerpt taken from the website[1].

"configUSE_TIME_SLICING:
By default (if configUSE_TIME_SLICING is not defined, or if configUSE_TIME_SLICING is defined as 1) FreeRTOS uses prioritized preemptive scheduling with time slicing. That means the RTOS scheduler will always run the highest priority task that is in the Ready state, and will switch between tasks of equal priority on every RTOS tick interrupt. If configUSE_TIME_SLICING is set to 0 then the RTOS scheduler will still run the highest priority task that is in the Ready state, but will not switch between tasks of equal priority just because a tick interrupt has occurred. "

However, if you set configUSE_TIME_SLICING = 0, it continues to switch context on tasks of equal priority. To solve this, it was necessary to add a check to see if the time_slicing is set or not, to decide which task will be performed, between it or next.

Test Steps:
---------------------------------------------------------------------------------------------------------
1° Create two tasks with same priority and print something many times with a delay between them.
2° Check if occur any context switch, seeing the debugs prints.
3° Check if the scheduler work in FIFO policy.

Example of task utilized:
---------------------------------------------------------------------------------------------------------
```c
inline void delay(){
        unsigned int i;
        for(i = 0; i < 800000000; i++) {
                asm("nop");
        }
}

void taskLed(void *pvParameters){
        int led = 0, i = 0;
        char *name = (char *)pvParameters;
        for (i = 0;i < 10; i++) {
                if(led){
                        printf("Led %s ON\n" , name);
                }else{
                        printf("Led %s OFF\n", name);
                }
                led ^= 1, delay();
        }
        vTaskDelete(NULL);
}
```

Related Issue:
---------------------------------------------------------------------------------------------------------

https://www.freertos.org/FreeRTOS_Support_Forum_Archive/December_2013/freertos_configUSE_TIME_SLICING_278999daj.html

http://www.openrtos.net/FreeRTOS_Support_Forum_Archive/December_2015/freertos_Tasks_w_same_Priority_Interrupt_even_with_time_slice_off_9336ef42j.html

https://www.freertos.org/FreeRTOS_Support_Forum_Archive/February_2018/freertos_Preemption_by_an_equal_priority_task_c8cb93d0j.html

https://www.freertos.org/FreeRTOS_Support_Forum_Archive/March_2019/freertos_In_the_case_of_configUSE_TIME_SLICING_0_aaaec33df6j.html

References:
---------------------------------------------------------------------------------------------------------
[1]https://www.freertos.org/a00110.html


dougnobrega:
If possible, put hackoutuberfest in label hehehe (:

RichardBarry:
Thanks for posting such a detailed explanation.  I'm looking into this now and think I may need a little more information in order to replicate your scenario.

The below code creates to tasks at the same priority.  The tasks do nothing other than increment a variable - task1 increments x and task2 increments y.

When I execute the code with these settings:

```
#define configUSE_PREEMPTION			1
#define configUSE_TIME_SLICING			1
```
then both x and y increment at roughly the same rate.  If I change the configuration to:

```
#define configUSE_PREEMPTION			1
#define configUSE_TIME_SLICING			0
```
Then only y increments.  x remains at 0.

Do you have any other tasks in your system?  If so, do they ever block?  If they do then that will cause a context switch to the task as configUSE_TIME_SLICING only prevents context switches just because a tick interrupt occurred - all other context switches occur as normal.

Test code follows.

```
volatile uint32_t x = 0, y = 0;

inline void delay(){
        unsigned int i;
        for(i = 0; i < 80000; i++) {
                asm("nop");
        }
}

void vTask1( void *pv )
{
	for( ;; )
	{
		x++;
		delay();
	}
}

void vTask2( void *pv )
{
	for( ;; )
	{
		y++;
		delay();
	}
}

int main( void )
{
	prvSetupHardware();
	xTaskCreate( vTask1, "task1", mainOLED_TASK_STACK_SIZE, NULL, configMAX_PRIORITIES - 1, NULL );
	xTaskCreate( vTask2, "task2", mainOLED_TASK_STACK_SIZE, NULL, configMAX_PRIORITIES - 1, NULL );
	vTaskStartScheduler();
}
```

dachalco:
Closing for now, but please re-open if these are reconciled:
 - Pending questions from Richard
 - For good practice, such a proposal should contain:
     - A lot of experiment data
     - More formal design analysis, and in this case, comparison. ([Arbitrary "googled" example](https://theses.ubn.ru.nl/bitstream/handle/123456789/168/Bex,%20P.MScThesis.pdf?sequence=1)).

PR #202, opened 2020-10-19
Print task priority value in vTaskGetRunTimeStats()

gomonovych:
Show task priority for better understanding of task run time consumption
with context of task priority

Description
-----------
I found it useful to observe task priority in the context of run time consumption.
This modification is platform independent.

Test Steps
-----------
I did simple test just run this code in my environment after task name it is priority number
IDLE           	0		265862		59%
Tmr Svc        	6		427		<1%
UART       	1		942		<1%

Related Issue
-----------


n9wxu:
The purpose of the vTaskGetRunTimeStats is to simply provide the runtime information of every task.  Keeping the output to a minimum is important to save memory and time.  If different information is required it is recommended that the function [uxTaskGetSystemState()](https://freertos.org/uxTaskGetSystemState.html) be called and a custom output function used to produce the required report.

PR #206, opened 2020-10-22
Add description for vTaskList

gomonovych:
Description
-----------

Describe each column which `vTaskList` print:
```
task name, task status, task priority, task stack unused watermark level, task number
```
Add description in function comment about columns which vTaskList print. Maybe like a next step it will be useful to add one top row with column descriptions?


n9wxu (approved):

include/task.h
```diff
@@ -1795,8 +1795,10 @@ UBaseType_t uxTaskGetSystemState( TaskStatus_t * const pxTaskStatusArray,
  * demo applications.  Do not consider it to be part of the scheduler.
  *
  * vTaskList() calls uxTaskGetSystemState(), then formats part of the
- * uxTaskGetSystemState() output into a human readable table that displays task
- * names, states and stack usage.
+ * uxTaskGetSystemState() output into a human readable table that displays task:
+ * names, states, priority, stack usage and task number.
+ * Stack usage specified as the number of unused variables stack can hold
```
RichardBarry, line 1800: This is very imprecise as variables can be almost any number of bytes.  Does it mean words?  A word being 16-bits on 16-bit devices, 32-bits on 32-bit devices?

tasks.c
```diff
@@ -4434,7 +4434,9 @@ static void prvResetNextTaskUnblockTime( void )
          *
          * vTaskList() calls uxTaskGetSystemState(), then formats part of the
          * uxTaskGetSystemState() output into a human readable table that
-         * displays task names, states and stack usage.
+         * displays task: names, states, priority, stack usage and task number.
```
RichardBarry, line 4437: Same comment as per prevision.

RichardBarry (changes requested):

tasks.c
gomonovych: Replaced with: Stack usage specified as the number of unused StackType_t words

include/task.h
gomonovych: Replaced with: Stack usage specified as the number of unused StackType_t words

n9wxu (approved):

PR #224, opened 2020-12-05
Improve heap2 bounds checking

cobusve:
This PR improves the heap2 bounds checking. There was a mostly theoretical case where an overflow could occur if the size of the requested memory block is between 4,294,967,288 and 4,294,967,296 bytes.

### DIFF
--- a/portable/MemMang/heap_4.c
+++ b/portable/MemMang/heap_4.c
@@ -136,44 +136,36 @@ void * pvPortMalloc( size_t xWantedSize )
          * kernel, so it must be free. */
         if( ( xWantedSize & xBlockAllocatedBit ) == 0 )
         {
-            /* The wanted size is increased so it can contain a BlockLink_t
+            /* The wanted size must be increased so it can contain a BlockLink_t
              * structure in addition to the requested amount of bytes. */
-            if( xWantedSize > 0 )
+            if( ( xWantedSize > 0 ) && 
+                ( ( xWantedSize + xHeapStructSize ) >  xWantedSize ) ) /* Overflow check */
             {
-                /* Check for overflow */
-                if ( ( xWantedSize + xHeapStructSize ) >  xWantedSize ) 
-                {
-                    xWantedSize += xHeapStructSize;
+                xWantedSize += xHeapStructSize;
 
-                    /* Ensure that blocks are always aligned to the required number
-                     * of bytes. */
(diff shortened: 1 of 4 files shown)


RichardBarry (approved):

portable/MemMang/heap_4.c
```diff
@@ -136,34 +136,42 @@ void * pvPortMalloc( size_t xWantedSize )
          * kernel, so it must be free. */
         if( ( xWantedSize & xBlockAllocatedBit ) == 0 )
```
sepetk1, line 137: Hello @cobusve and @RichardBarry!

For heap_4.c and heap_5.c it seems an integer overflow (i.e. from 0xFFFFFFFF to 0x00000000) is already prevented due to checking the most significant bit of `xWantedSize` here. Right? It seems like the only way for an integer overflow to happen would be if `xHeapStructSize ` plus a few bytes for byte alignment are large enough to make `xWantedSize` go from 0x7FFFFFFF to 0x00000000. (That's 2GB for 32-bit systems and 32.768KB for 16-bit systems, and `xHeapStructSize ` isn't that big.)

So should the new bounds checks in heap_4.c and heap_5.c check for unintentionally setting the most significant bit of `xWantedSize` instead? (i.e. `xWantedSize` goes from 0x7FFFFFFF to 0x80000000) If I understood correctly, the most significant bit is reserved, so the largest buffer one can allocate is technically 2^31 instead of 2^32. So I thought that new bounds checking might look something like this.
```C
if( ( xWantedSize & xBlockAllocatedBit ) == 0 )
{
    if( ( xWantedSize > 0 ) && 
        ( ( xWantedSize + xHeapStructSize ) < xBlockAllocatedBit ) ) /* <-- CHANGED THIS LINE */
    {
        xWantedSize += xHeapStructSize;
        // ...
    }
    // ...
}
```
I'm not sure if I'm right about this, but hope this makes sense. Thanks in advance!

dan4thewin: Thank you for diving deep on this.  It seems quite likely that your changes work as desired.  We have another effort underway to use automated reasoning to improve our code.  This includes using proofs with CBMC, a bounded model checker.  As we write a proof for this code, we expect to change these overflow checks to prove that overflows are impossible.  Here's a recent example of such changes: https://github.com/FreeRTOS/FreeRTOS-Kernel/pull/328 .  We'd welcome your feedback on these efforts, and are grateful for your continued interest.

PR #237, opened 2020-12-21
ARM_CA53: add support for FIQ

n-royer:
FIQ is used in the context of ARM SCP firmware instead of IRQ.
This patch adds FIQ support for this context.

Signed-off-by: Nicolas Royer <nroyer@baylibre.com>

Description
-----------

Test Steps
-----------

Related Issue
-----------


RichardBarry:
Apologies for the delay here.  We have a new mechanism for more timely handling of PRs but we were unsure of this one so I was asked to look.  Historically FreeRTOS has supported IRQs, including with nesting, but left the FIQ completely untouched for the application writer to do as they wish - often there will be a single use for the FIQ.  Can you give more detail of the intent of this PR - is it enabling support for either the IRQ or the FIQ, rather than both at once?  Would it be better to handle the FIQ separately?

PR #240, opened 2021-01-01
Wrap macros in do { ... } while( 0 )

j4cbo:
Wrap macros in do { ... } while( 0 )

Description
-----------
If statements in macros can be hazardous. Here's a contrived example:

    if( length == 0 )
        configASSERT( data == NULL )
    else
        configASSERT( data != NULL )

This looks sensible at first glance, but it expands to:

    if( length == 0 )
        if (!( data == NULL )) { ... }
    else
        if (!(data != NULL )) { ... }

... which is not intended; the `else` block associates with the `if` from the assert macro instead of the `if( length == 0)`. The usual pattern to avoid this is to wrap the macro in `do { ... } while (0)` which causes the right thing to happen.

Test Steps
-----------
n/a


RichardBarry:
Our coding standard does not permit decision points without blocks.  So your example would look like this:

```
if( length == 0 )
{
    configASSERT( data == NULL )
}
else
{
    configASSERT( data != NULL )
}
```

j4cbo:
That's fine within FreeRTOS, but users may have different standards.

Additionally, the existing structure causes redundant semicolons, which trigger compiler warnings; some of the existing macros are defined as `if( xSwitchRequired ) statement` which leave the user to add the semicolon themselves, while others are `if( xSwitchRequired ) { statement; }` where a semicolon afterwards would be redundant.

j4cbo:
Your Github pre-commit hook wants me to remove third party copyright headers from the files I've changed, in violation of the stated license terms of those files (e.g. portable/ThirdParty/CDK/T-HEAD_CK802/portmacro.h, `Copyright (C) 2017 C-SKY Microsystems Co., Ltd.`). I am not comfortable making those changes.

RichardBarry:
> Your Github pre-commit hook wants me to remove third party copyright headers from the files I've changed, in violation of the stated license terms of those files (e.g. portable/ThirdParty/CDK/T-HEAD_CK802/portmacro.h, `Copyright (C) 2017 C-SKY Microsystems Co., Ltd.`). I am not comfortable making those changes.

Think you are right not to.  I will bring this to the teams attention.  Thanks.

dachalco:
We pessimistically assume any unknown new files _should_ contain, to prevent new kernel files from sneaking in with incorrect headers. 

However, I realize that unlike `FreeRTOS/FreeRTOS` this check does not follow up with description of what is going on. I'll be updating this shortly. 

Fixed some outdated headers and added the others to check's ignore list.

aggarw13 (approved):

n9wxu (approved):

PR #250, opened 2021-01-20
Posix: fix copyright

unspecd:
This fixes compilation error when `configENABLE_BACKWARD_COMPATIBILITY` is 0.

### DIFF
--- a/portable/ThirdParty/GCC/Posix/portmacro.h
+++ b/portable/ThirdParty/GCC/Posix/portmacro.h
@@ -22,7 +22,6 @@
  * https://www.FreeRTOS.org
  * https://github.com/FreeRTOS
  *
- * 1 tab == 4 spaces!
  */
 
 
--- a/.github/scripts/kernel_checker.py
+++ b/.github/scripts/kernel_checker.py
@@ -36,6 +36,7 @@ KERNEL_IGNORED_EXTENSIONS = [
 
 KERNEL_IGNORED_PATTERNS = [
     r'.*\.git.*',
+    r'.*portable/ThirdParty/GCC/Posix/port*',
     r'.*portable.*Xtensa_ESP32\/include\/portmacro\.h',
     r'.*portable.*Xtensa_ESP32.*port\.c',
     r'.*portable.*Xtensa_ESP32.*portasm\.S',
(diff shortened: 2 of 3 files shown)


unspecd:
Should I squash everything into 1 commit?

dachalco:
@unspecd 

Thanks for bringing this up. Let me check with some colleagues. Iff I recall correctly, this deficiency remained present by design.

dachalco:
@unspecd 

Is there a particular reason for removing these macros?

The preprocessor can do the same replacement with:
```
#define pdTASK_CODE TaskFunction_t
```

Also the copyright line has been modified and I'm not sure whether the copyright line change is appropriate for `port.c`

unspecd:
> This removes the convenient ability to redefine the task function signature in Thread_t via pdTASK_CODE

Oh, I didn't take this into account.

unspecd:
> port.c copyright should remain unchanged 

OK, changed to "Cambridge Consultants Ltd".

portable/ThirdParty/GCC/Posix/portmacro.h
```diff
@@ -70,7 +69,7 @@ typedef unsigned long TickType_t;
 #define portSTACK_GROWTH			( -1 )
 #define portHAS_STACK_OVERFLOW_CHECKING	( 1 )
 #define portTICK_PERIOD_MS			( ( TickType_t ) 1000 / configTICK_RATE_HZ )
-#define portTICK_RATE_MICROSECONDS	( ( portTickType ) 1000000 / configTICK_RATE_HZ )
+#define portTICK_RATE_MICROSECONDS	( ( TickType_t ) 1000000 / configTICK_RATE_HZ )
```
dachalco, line 72: What are the implications of removing `portTickType` here?

`TickType_t` is limited to 16 or 32 bits and one utility of having `portTickType` here is that 64-bit posix targets can define `portTickType` as 64-bit instead, allowing finer granularity.


I'm not very familiar with how our Posix handles/quantifies tick cycles.
However I can say that

dachalco: Is there specific issue that removing `portTickType` here solves? Why the removal?

unspecd: There are no specific issues, I just wanted to fix a compilation error.

paulbartell (approved):

dan4thewin (approved):

PR #258, opened 2021-02-09
Code: Remove dead code

alfred2g:
Remove dead code in  list.c

Description
-----------
Cde in uxListRemove
```c
    if( pxList->pxIndex == pxItemToRemove )
    {
            pxList->pxIndex = pxItemToRemove->pxPrevious;
    }
```
Can never evaluate to true, as pxList->pxIndex points to ListEnd which is MiniListItem_t, and should always be present, and doesn't count towards the count of the list elements

### DIFF
--- a/list.c
+++ b/list.c
@@ -192,16 +192,6 @@ UBaseType_t uxListRemove( ListItem_t * const pxItemToRemove )
     /* Only used during decision coverage testing. */
     mtCOVERAGE_TEST_DELAY();
 
-    /* Make sure the index is left pointing to a valid item. */
-    if( pxList->pxIndex == pxItemToRemove )
-    {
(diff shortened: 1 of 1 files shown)


RichardBarry:
pxList->pxIndex is used to walk through a linked list so the next item in the list can be returned on subsequent calls - it is effectively a marker in the list.  It can therefore point to any item in a linked list.  If you delete an item from a list then the code your PR removes checks to see if the item being deleted happens to be the item pointed to by pxIndex - and if it is moves pxIndex on to ensure pxIndex points to a valid (i.e. not deleted) item in the list.  Code coverage results and experimentation by setting a break point on the line being removed by this PR both indicate the line is executed very frequently - hence closing this PR without merging.

PR #263, opened 2021-02-15
Posix: use thread local storage for critical section counter

twpedersen:
This commit allows non-FreeRTOS pthreads to safely call
vPortEnterCritical()/vPortExitCritical(). ie. avoid lockups due to seemingly nonblocked signals creating deadlocks.

Non-FreeRTOS Task pthreads may need to enter a critical
section (for eg. printf()). In this case a single global
uxCriticalNesting is bogus since it is not copied and
restored on pthread scheduling.

GCC Thread-Local Storage to the rescue.

See https://gcc.gnu.org/onlinedocs/gcc/Thread-Local.html

### DIFF
--- a/portable/ThirdParty/GCC/Posix/port.c
+++ b/portable/ThirdParty/GCC/Posix/port.c
@@ -366,8 +366,11 @@ int iRet;
 
 static void vPortSystemTickHandler( int sig )
 {
+#if ( configUSE_PREEMPTION == 1 )
 Thread_t *pxThreadToSuspend;
 Thread_t *pxThreadToResume;
+#endif
+
 /* uint64_t xExpectedTicks; */
 
     uxCriticalNesting++; /* Signals are blocked in this signal handler. */
--- a/portable/ThirdParty/GCC/Posix/port.c
+++ b/portable/ThirdParty/GCC/Posix/port.c
@@ -94,7 +94,7 @@ static sigset_t xResumeSignals;
 static sigset_t xAllSignals;
 static sigset_t xSchedulerOriginalSignalMask;
 static pthread_t hMainThread = ( pthread_t )NULL;
-static volatile portBASE_TYPE uxCriticalNesting;
+static volatile __thread portBASE_TYPE uxCriticalNesting = 0;
 /*-----------------------------------------------------------*/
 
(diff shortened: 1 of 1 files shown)


twpedersen:
can someone take a look at this? I think it is kind of a critical fix for LTS.

alfred2g:
> can someone take a look at this? I think it is kind of a critical fix for LTS.

I did take a look at this, non FreeRTOS pthreads should not call FreeRTOS specific constructs and should not be mixed with FreeRTOS tasks, as it messes the management of threads with FreeRTOS, and makes debugging trickier.

We solve the issue by disabling signals on non-FreeRTOS threads such as here: https://github.com/FreeRTOS/FreeRTOS-Plus-TCP/blob/9a25860e761036a9eb780799c9db632e3eff60c9/portable/NetworkInterface/linux/NetworkInterface.c#L661
This solves the problem because it blocks the non-FreeRTOS threads from receiving signals and losing their execution path and sleep forever. if you need some other signals, obviously you don't have to block the whole set.

Doing that, you will be able to call  printf and any other blocking calls without any problems.

twpedersen:
Thanks for your feedback @alfred2g. In our case we had some code which was called by both FreeRTOS and non-FreeRTOS threads which needed to enter a critical section.

Copying the critical section counter onto the stack and restoring after a context switch also seems harder to reason about (and debug) than saying "this value is definitely thread-local".

alfred2g:
How does that translate into the real world(real hardware) ?
An extra pthread is meant to simulate, an interrupt (hardware side)  which is not allowed, or doesn't have have access to FreeRTOS functions.
For your case you might be needing a FreeRTOS task instead of a pthread ? 
Cant you separate that code so no overlap occurs ? 

> Copying the critical section counter onto the stack and restoring after a context switch also seems harder to reason about (and debug) than saying "this value is definitely thread-local".

Yeah,  you've got a point here, I will have to ask around about introducing C11 syntax, as not all our users have the latest compilers, maybe using pthread_once instead?

alfred2g:
After discussing it internally, and we think this change, even though it is good, and probably better than storing the value on the stack, we think it would make it easier for someone to do the wrong thing.
I will close the PR, if you have a real life simulation problem that this solves  please reopen the case.

PR #273, opened 2021-03-05
Add some assertions and coverage exceptions to queue.c

paulbartell:
Add some assertions and coverage exceptions to queue.c

Description
-----------
* Correct some typos in queue.c
* Assert that the semaphore handle passed into xQueueGetMutexHolder is not NULL.
* Add assertions to check when invalid parameters are passed into Queue Registry related functions.
* Add a configASSERT when vQueueAddToRegistry is called but the queue registry is full.
* Add LCOV_BRANCH coverage exception for a configASSERT on pxQueueSetContainer with a condition that is unreachable.
* Add an LCOV_BRANCH exception for the check that sizeof( StaticQueue_t ) == sizeof( Queue_t )
* Run uncrustify

Test Steps
-----------
Only assertions and comments have been changed. This should be verified by inspection.

### DIFF
--- a/queue.c
+++ b/queue.c
@@ -345,7 +345,7 @@ BaseType_t xQueueGenericReset( QueueHandle_t xQueue,
 
                 /* This assertion cannot be branch covered in unit tests */
                 configASSERT( xSize == sizeof( Queue_t ) ); /* LCOV_EXCL_BR_LINE */
-                ( void ) xSize; /* Keeps lint quiet when configASSERT() is not defined. */
+                ( void ) xSize;                             /* Keeps lint quiet when configASSERT() is not defined. */
             }
         #endif /* configASSERT_DEFINED */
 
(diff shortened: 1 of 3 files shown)


queue.c
```diff
@@ -342,8 +342,10 @@ BaseType_t xQueueGenericReset( QueueHandle_t xQueue,
                  * variable of type StaticQueue_t or StaticSemaphore_t equals the size of
                  * the real queue and semaphore structures. */
                 volatile size_t xSize = sizeof( StaticQueue_t );
-                configASSERT( xSize == sizeof( Queue_t ) );
-                ( void ) xSize; /* Keeps lint quiet when configASSERT() is not defined. */
+
+                /* This assertion cannot be branch covered in unit tests */
+                configASSERT( xSize == sizeof( Queue_t ) ); /* LCOV_EXCL_BR_LINE */
```
paulbartell, line 347: This is really a job for _Static_assert (C11).

queue.c
```diff
@@ -2741,6 +2748,9 @@ BaseType_t xQueueIsQueueFullFromISR( const QueueHandle_t xQueue )
                 mtCOVERAGE_TEST_MARKER();
             }
         }
+
+        /* Assert that the queue was added successfully */
+        configASSERT( ( ux != configQUEUE_REGISTRY_SIZE ) );
```
paulbartell, line 2753: I'm not totally sure if this should be added or not. It seems like it would be helpful to warn about failed vQueueAddToRegistry calls, but this might cause undesirable behavior. 

Currently, a user can add a given queue handle multiple times to the Queue Registry and vQueueAddToRegistry will not overwrite the existing entries for that queue handle.

Calls to pcQueueGetName always returns the first entry found for a given queue handle or NULL if the entry is not found.

queue.c
```diff
@@ -2723,6 +2727,9 @@ BaseType_t xQueueIsQueueFullFromISR( const QueueHandle_t xQueue )
     {
         UBaseType_t ux;
 
+        configASSERT( xQueue );
```
paulbartell, line 2730: This change will result in users with configASSERT enabled not being able to add a NULL queue handle to the queue registry.

n9wxu (approved):

ravibhagavandas (approved):

PR #274, opened 2021-03-06
Request For Comment: Add additional PR checks for formatting

paulbartell:
Add some additional linters from the pre-commit project.

Specifically enable these checks on files that have changed as part of a pull-request:

- trailing-whitespace (remove trailing whitespace from lines)
- end-of-file-fixer (check that files end with an LF or CRLF)
- check-yaml (link yaml files)
- check-added-large-files (really of less usefulness after a push)
- mixed-line-ending (reject files with a mix of LF and CRLF line endings)
- check-case-conflict (check  files with the same name on case-insensitive filesystems)
- check-executables-have-shebangs
- check-json (link json files)
- check-merge-conflict (check for unresolved merge conflicts)
- check-symlinks (check of dangling symlinks)
- check-ast (validate python)


.pre-commit-config.yaml
```diff
@@ -0,0 +1,18 @@
+# See https://pre-commit.com for more information
+# See https://pre-commit.com/hooks.html for more hooks
```
milesfrain, line 2: May want to add the `uncrustify` check from `github.com/pocc/pre-commit-hooks` and move a copy of the formatting config to this kernel repo. https://github.com/FreeRTOS/FreeRTOS/blob/master/tools/uncrustify.cfg

milesfrain:
https://github.com/paulbartell/FreeRTOS-Kernel/pull/10 is an initial attempt to use uncrustify for formatting.

Not sure if a pre-commit hook is the best path forward, since that can be pretty annoying with unexpected formatting rewrites and having to constantly use `--no-verify` to suppress the hook. A better option might be to just catch this stuff in PR CI and provide an easy way for folks to run the autoformatter locally.

paulbartell:
> [paulbartell#10](https://github.com/paulbartell/FreeRTOS-Kernel/pull/10) is an initial attempt to use uncrustify for formatting.
> 
> Not sure if a pre-commit hook is the best path forward, since that can be pretty annoying with unexpected formatting rewrites and having to constantly use `--no-verify` to suppress the hook. A better option might be to just catch this stuff in PR CI and provide an easy way for folks to run the autoformatter locally.

The uncrustify pre-commit hook needs some configuration so that it acts as one would expect. Namely, it needs to only run on files that have changed and operate in-place without leaving backup files everywhere. I'll do some debugging on this later this week.

milesfrain:
> The uncrustify pre-commit hook needs some configuration so that it acts as one would expect. Namely, it needs to only run on files that have changed and operate in-place without leaving backup files everywhere.

I believe that's how it currently operates by default (at least that's what I've observed in paulbartell#10 locally). 

My point is that we may not want pre-commit to auto-edit files "behind the scenes" when a user simply wants to make a local commit.

paulbartell:
> > The uncrustify pre-commit hook needs some configuration so that it acts as one would expect. Namely, it needs to only run on files that have changed and operate in-place without leaving backup files everywhere.
> 
> I believe that's how it currently operates by default (at least that's what I've observed in [paulbartell#10](https://github.com/paulbartell/FreeRTOS-Kernel/pull/10) locally).
> 
> My point is that we may not want pre-commit to auto-edit files "behind the scenes" when a user simply wants to make a local commit.

For sure. It should be optional. You can always run "pre-commit" from within a git repo without installing it as a pre-commit hook too.

PR #277, opened 2021-03-15
Improve comments and assertions in stream buffer

ravibhagavandas:
Description
-----------
PR includes following changes:

* Update comments in stream buffer code.
* Removes redundant check within stream buffer create static function.
* Fixes an edge case when trigger level is set equal to stream buffer size + 1
* Fixes a read before null check in stream buffer send API.

### DIFF
--- a/stream_buffer.c
+++ b/stream_buffer.c
@@ -301,8 +301,6 @@ static void prvInitialiseNewStreamBuffer( StreamBuffer_t * const pxStreamBuffer,
         StreamBufferHandle_t xReturn;
         uint8_t ucFlags;
 
-        configASSERT( pucStreamBufferStorageArea );
-        configASSERT( pxStaticStreamBuffer );
         configASSERT( xTriggerLevelBytes <= xBufferSizeBytes );
 
         /* A trigger level of 0 would cause a waiting task to unblock even when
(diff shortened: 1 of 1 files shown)


stream_buffer.c
```diff
@@ -340,27 +339,19 @@ static void prvInitialiseNewStreamBuffer( StreamBuffer_t * const pxStreamBuffer,
             } /*lint !e529 xSize is referenced is configASSERT() is defined. */
         #endif /* configASSERT_DEFINED */
 
-        if( ( pucStreamBufferStorageArea != NULL ) && ( pxStaticStreamBuffer != NULL ) )
```
paulbartell, line 343: Can you clarify why you want to remove the null pointer checks here?

ravibhagavandas: Check is redundant since there is an assertion for this at line 304 and 305.

paulbartell: configASSERT is not always defined for production builds.

ravibhagavandas: Agree. Adding the NULL check here makes sense and it confirms to the API documentation as well. However I feel both assert and null check is redundant and adds to difference in API behavior  between production and non-production builds.

ravibhagavandas: Added `configASSERT` back to keep the current API behavior same.

paulbartell (approved):

gkwicker (approved):

PR #293, opened 2021-03-24
Check NULL pointer input for vEventGroupDelete.

mingyue86010:

### DIFF
--- a/event_groups.c
+++ b/event_groups.c
@@ -625,49 +625,45 @@ EventBits_t xEventGroupSetBits( EventGroupHandle_t xEventGroup,
 
 void vEventGroupDelete( EventGroupHandle_t xEventGroup )
 {
-    EventGroup_t * pxEventBits;
-    const List_t * pxTasksWaitingForBits;
+    configASSERT( xEventGroup );
 
-    if ( NULL != xEventGroup )
+    EventGroup_t * pxEventBits = xEventGroup;
+    const List_t * pxTasksWaitingForBits = &( pxEventBits->xTasksWaitingForBits );
+
+    vTaskSuspendAll();
     {
-        pxEventBits = xEventGroup;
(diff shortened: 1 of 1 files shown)


event_groups.c
```diff
@@ -625,43 +625,49 @@ EventBits_t xEventGroupSetBits( EventGroupHandle_t xEventGroup,
 
 void vEventGroupDelete( EventGroupHandle_t xEventGroup )
 {
-    EventGroup_t * pxEventBits = xEventGroup;
-    const List_t * pxTasksWaitingForBits = &( pxEventBits->xTasksWaitingForBits );
+    EventGroup_t * pxEventBits;
+    const List_t * pxTasksWaitingForBits;
 
-    vTaskSuspendAll();
+    if ( NULL != xEventGroup )
```
paulbartell, line 631: It might be best to add a configASSERT ( xEventGroup ) so that the error condition can be reported during development instead of silently failing.

mingyue86010: I considered of that but I think not using assert is more common cases of a funciton that free a memory with no side effects. I don't think it's a good idea to break the code running by this.

event_groups.c
```diff
@@ -625,43 +625,49 @@ EventBits_t xEventGroupSetBits( EventGroupHandle_t xEventGroup,
...
 {
-    EventGroup_t * pxEventBits = xEventGroup;
-    const List_t * pxTasksWaitingForBits = &( pxEventBits->xTasksWaitingForBits );
+    EventGroup_t * pxEventBits;
+    const List_t * pxTasksWaitingForBits;
 
-    vTaskSuspendAll();
+    if ( NULL != xEventGroup )
     {
-        traceEVENT_GROUP_DELETE( xEventGroup );
```
cobusve, line 633: I thiink it is actually important that the trace hook is the first line in the function here?

paulbartell (approved):

aggarg (approved):

PR #294, opened 2021-03-25
Added CMake support

robamu:
# CMake support for FreeRTOS

Description
-----------
This adds very basic CMake support for FreeRTOS. It has has become a very common build system so I think this would be a nice feature.  Could also be easily extended. I have tested it in our projects as well.

Details, Instructions
-----------
Compiling FreeRTOS requires external includes (e.g. path to `FreeRTOSConfig.h`) and additional sources like the port source and the memory management source file, so the user can set a list of includes and sources from an upper `CMakeLists.txt` file before adding the FreeRTOS folder and linking against `freertos`:

1. Source list with the name `FREERTOS_ADDITIONAL_SOURCES`
2. Include list with the name `FREERTOS_ADDITIONAL_INCLUDE_PATHS`

I am using a GCC compiler. For GCC compilers, the CMake support will add the required flags and link options to remove unused sections and also enable some more warnings. I compiled FreeRTOS with these warning flags and there were no warnings. These features can also be disabled from an upper `CMakeLists.txt` file.

If FreeRTOS needs to be compiled again another libraries for any reason (e.g. port requires this), the variable `FREERTOS_ADDITIONAL_LIBS` can be set to link FreeRTOS against these libraries.


rmspacefish:
Marked WIP because I need to replace `CMAKE_CXX_COMPILER_ID` with the more generic `CMAKE_C_COMPILER_ID`

robamu:
I might add a different version which adds an interface library instead of building a static library

kstribrnAmzn:
CI checks are failing due to trailing whitespaces in the CMakeLists.txt file. I'm going to remove these whitespaces and push a small change.

yhsb2k:
Hi All,
I looked through this PR and can say that it isn't enough flexible for common use. Here are some arguments:
* Passing custom compile options only for GCC isn't good practice for integration to different projects
* Making user responsible for setting `FREERTOS_ADDITIONAL_SRC` and `FREERTOS_ADDITIONAL_INCLUDE_PATHS` will make the FreeRTOS more difficult to use (integrate). In this case, user should find and pass all specific sources and include paths to FreeRTOS cmake, but he shouldn't do it
* This approach doesn't allow to set custom heap implementation. This is a fairly common usecase
* Overall, it looks overcomplicated due to many custom variables, foreach loops and hardcoded compile options

Perhaps it would be better to use the approach described in this PR https://github.com/FreeRTOS/FreeRTOS-Kernel/pull/421 ?

robamu:
I agree that building a static library is not really the best approach here. I worked with different approaches to integrate external libraries and for FreeRTOS, it makes more sense to let the user supply the arguments themselves and add an interface library like mentioned above.  This is how it was done in the https://github.com/ObKo/stm32-cmake library. #421 looks interesting as well

kstribrnAmzn:
Thank you for your inputs @yhsb2k and @robamu. I think we can close this PR while we continue the discussion on the following PR: https://github.com/FreeRTOS/FreeRTOS-Kernel/pull/421. 
We are looking into adding the CMake support in a manner which is most flexible as well as easy to use. Thank you for your time and effort submitting this PR and we hope to have CMake support for you soon.

PR #296, opened 2021-03-26
add cmake for posix port

marcinbor85:
Description
-----------
Basic CMake scripts to easy configure source files and include directories.
Script configures two variables: FREERTOS_SRC and FREERTOS_INC to use in other top-level CMake scripts.
This makes it convenient to import freertos files to the top level CMake scripts.
It is very easy concept but extremely flexible I think.
In the opposite to others CMake integrations, this one configure variables with files and include directories only.
The build method, compiler flags, etc..., are left for top-level CMake.
It allows to make many different builds for multi-target projects.

Usage
-----------
```bash
# configure memory management (by filename without extension)
set( FREERTOS_MEMMANG heap_3 )
# configure port (by portable path)
set( FREERTOS_PORT ThirdParty/GCC/Posix )
# add top-level cmake script
add_subdirectory( FreeRTOS-Kernel )
# here we get a FREERTOS_SRC and FREERTOS_INC variables to free use in any case.
```

Limitations
-----------
It doesn't build any static libraries due to missing FreeRTOSConfig.h file.


CMakeLists.txt
```diff
@@ -0,0 +1,29 @@
...
+set(
+    FREERTOS_SRC
+    ${CMAKE_CURRENT_SOURCE_DIR}/tasks.c
+    ${CMAKE_CURRENT_SOURCE_DIR}/timers.c
+    ${CMAKE_CURRENT_SOURCE_DIR}/stream_buffer.c
+    ${CMAKE_CURRENT_SOURCE_DIR}/queue.c
+    ${CMAKE_CURRENT_SOURCE_DIR}/list.c
+    ${CMAKE_CURRENT_SOURCE_DIR}/event_groups.c
+    ${CMAKE_CURRENT_SOURCE_DIR}/croutine.c
+    ${CMAKE_CURRENT_SOURCE_DIR}/portable/MemMang/${FREERTOS_MEMMANG}.c
```
jwbmwv, line 23: In the interest of accommodating all common use cases (in addition to laying the foundation for using Cmake on other targets), you might want to consider allowing FREERTOS_MEMMANAGE to accommodate a user defined solution. The most common ones used on STM and NXP ARM systems is Dave Nadler's freely available implementation.

marcinbor85: Maybe it will be useful if I will make a FREERTOS_MEMMANG variable optional. if it is defined, it will be used as above, otherwise the user will have to take care of including the equivalent of the files with pvPortMalloc and vPortFree implementations. The main idea was to keep the script as little as possible and give it as much flexibility as possible.

AniruddhaKanhere:
Hello @marcinbor85, thank you for this PR. Firstly, apologies for the very late response.

I am closing this now since a all-encompassing PR (https://github.com/FreeRTOS/FreeRTOS-Kernel/pull/421) has already been merged in the repository.

For anyone wanting to build the Posix port of FreeRTOS-Kernel, one can follow the below instructions when including the Kernel as a sub-directory:
```
set(FREERTOS_PORT GCC_POSIX CACHE STRING "")
set(FREERTOS_CONFIG_FILE_DIRECTORY ${CMAKE_CURRENT_LIST_DIR} CACHE STRING "")
add_subdirectory(third_party/FreeRTOS-Kernel)
```
Thanks again

PR #302, opened 2021-04-07
Remove Co-routines from FreeRTOS-Kernel.

paulbartell:
Remove Co-routines from FreeRTOS-Kernel.

Description
-----------
Remove the co-routines implementation from FreeRTOS-Kernel since this code is no longer maintained or widely used.

Related Forum post:
-----------
https://forums.freertos.org/t/about-coroutine/11716/2


include/FreeRTOS.h
```diff
@@ -106,10 +106,6 @@
     #error Missing definition:  configUSE_16_BIT_TICKS must be defined in FreeRTOSConfig.h as either 1 or 0.  See the Configuration section of the FreeRTOS API documentation for details.
 #endif
 
-#ifndef configUSE_CO_ROUTINES
-    #define configUSE_CO_ROUTINES    0
-#endif
```
RichardBarry, line 111: Please add a #warning message if configUSE_CO_ROUTINES is defined and defined to 1.  Message should tell people which version of code has the source files, preferably with a link.  We should tell people how to add the files back into their projects if they upgrade FreeRTOS, on the understanding the code is not maintained so support is limited.

RichardBarry (changes requested):
If the source files is to be removed then we would also have to update or remove any demo project that builds the files.

include/FreeRTOS.h
paulbartell: Will do.

I've started porting over some of the demos where configUSE_CO_ROUTINES == 1. Are there any in this list that I should drop entirely rather than port over to use tasks rather than co-routines?

TI / Luminary Micro:
- CORTEX_LM3S102_Rowley
- CORTEX_LM3S316_IAR
- CORTEX_LM3S811_IAR

Microchip / Atmel
- dsPIC_MPLAB
- PIC24_MPLAB
- AVR_Dx_MPLAB.X
- AVR_Dx_IAR
- AVR_Dx_Atmel_Studio
- AVR_ATMega4809_MPLAB.X
- AVR_ATMega4809_Atmel_Studio
- AVR_ATMega4809_IAR
- AVR_ATMega323_WinAVR
- AVR_ATMega323_IAR

Fujitsu / Cypress:
- MB91460_Softune
- MB96340_Softune
- MB96350_Softune_Dice_Kit

NXP / Freescale:
- ColdFire_MCF52221_CodeWarrior

Simulator:
- WIN32-MSVC

n9wxu:
At least the AVR examples have coroutines so closing this PR.

PR #303, opened 2021-04-08
Fix stop/delete of backlogged auto-reload timer

jefftenney:
#### Fix stop/delete backlogged auto-reload timer

Description
-----------
Prior to this PR, if an application attempts to stop or delete an auto-reload timer while it is backlogged, the timer unexpectedly restarts.  A timer is "backlogged" when at the time of the auto-reload, the new timer period has already elapsed.  In the case of a deleted timer, the timer continues running using freed memory which may eventually lead to a crash or failure of other unrelated timers when the memory is reallocated.

Test Steps
-----------
See https://github.com/FreeRTOS/FreeRTOS/pull/553.

Related Issue
-----------
https://forums.freertos.org/t/freertos-time-correction/11856
https://forums.freertos.org/t/timer-start-dont-trace-race-condition/12168

### DIFF
--- a/include/timers.h
+++ b/include/timers.h
@@ -51,10 +51,16 @@
  * be used solely through the macros that make up the public software timer API,
  * as defined below.  The commands that are sent from interrupts must use the
  * highest numbers as tmrFIRST_FROM_ISR_COMMAND is used to determine if the task
- * or interrupt version of the queue send function should be used. */
+ * or interrupt version of the queue send function should be used.  Priority
+ * commands, which are sent to the front of the command queue, must begin at
+ * zero as tmrLAST_PRIORITY_COMMAND is used to determine if a command is a
+ * priority command. */
 #define tmrCOMMAND_EXECUTE_CALLBACK_FROM_ISR    ( ( BaseType_t ) -2 )
 #define tmrCOMMAND_EXECUTE_CALLBACK             ( ( BaseType_t ) -1 )
+
 #define tmrCOMMAND_START_DONT_TRACE             ( ( BaseType_t ) 0 )
+#define tmrLAST_PRIORITY_COMMAND                ( ( BaseType_t ) 0 )
+
 #define tmrCOMMAND_START                        ( ( BaseType_t ) 1 )
 #define tmrCOMMAND_RESET                        ( ( BaseType_t ) 2 )
 #define tmrCOMMAND_STOP                         ( ( BaseType_t ) 3 )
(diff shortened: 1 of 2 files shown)


RichardBarry:
> Preferred solution would be that timer.c will check that timer was already deleted

This is not an easy thing to do because the memory used by the timer's structures is freed so could contain anything after the timer has been deleted.  Even in the case of a timer being allocated statically, the statically allocated memory may be re-used to create another timer.

There is an analogous problem with tasks.  Kernel aware debuggers that display tasks had no idea if a task had been deleted and re-created since it last viewed the memory.  In that situation we added a unique number to each task structure.  If the kernel aware debugger saw a task structure (TCB) occupying the same memory location but with a different unique number it knows its a different task even if the task name and handle are the same (the handle just being a pointer to the TCB).  I was wondering if we could do something similar with timers.  If a timer is recognised from a combination of its handle and unique number contained within its structure - then a command operating on a timer that had a different unique number to that expected could be discarded.

jefftenney:
> We could consider that in those cases the error is on the part of the application writer as they are sending a command to a timer they know they have already deleted, whereas the autoreload event is a special case because it happens automatically. That would seem reasonable. 

Agreed.

> Can we think of any other cases that need to same treatment (sending to the front)?

No, but let me fix this up a little bit to prompt developers in the future to consider any new command for sending to front.  It would be too easy with the code as is for a developer not to consider the question.

My first thought is to set a bit in each command value as a tag, either `tmrCMD_GROUP_FROM_ISR`, `tmrCMD_GROUP_PRIORITY`, `tmrCMD_GROUP_STANDARD`.  Then `xTimerGenericCommand()` would use those bits to determine whether to call `xQueueSendToBackFromISR()`, `xQueueSendToFront()`, or `xQueueSendToBack()`.  It currently uses `xCommandID < tmrFIRST_FROM_ISR_COMMAND` to make the determination.

jefftenney:
> Seems a good solution, and simpler that those I was contemplating.

You might have been right contemplating other solutions.  Please see #305 for what I think is a better solution.  Let me know which way you'd like to go.  This solution does have the advantage of having fewer changes to the code.  But it's awkward in its operation and not as clean in its design as the rest of FreeRTOS code.  So sorry if you've wasted time on my waffling.

Note also that in #305 I fixed a couple of other issues.  Will gladly fix (or not fix) them here in this PR as you prefer.

jefftenney:
Closing this PR since #305 was better.

PR #306, opened 2021-04-09
Overwrite existing QueueRegistry entries when a handle is added multiple times.

paulbartell:
Description
-----------
Overwrite an existing entry for a given xQueue handle when vQueueAddToRegistry is called with an xQueue handle of a queue that is already in the Queue Registry.

Test Steps
-----------
Test will be updated in https://github.com/FreeRTOS/FreeRTOS/pull/552.

### DIFF
--- a/queue.c
+++ b/queue.c
@@ -2730,23 +2730,36 @@ BaseType_t xQueueIsQueueFullFromISR( const QueueHandle_t xQueue )
         configASSERT( xQueue );
         configASSERT( pcQueueName );
 
+        QueueRegistryItem_t * pxEntryToWrite = NULL;
+
         /* See if there is an empty space in the registry.  A NULL name denotes
          * a free slot. */
         for( ux = ( UBaseType_t ) 0U; ux < ( UBaseType_t ) configQUEUE_REGISTRY_SIZE; ux++ )
         {
-            if( xQueueRegistry[ ux ].pcQueueName == NULL )
+            /* Replace an existing entry if the queue is already in the registry. */
+            if( xQueueRegistry[ ux ].xHandle == xQueue )
             {
-                /* Store the information on this queue. */
-                xQueueRegistry[ ux ].pcQueueName = pcQueueName;
-                xQueueRegistry[ ux ].xHandle = xQueue;
-
-                traceQUEUE_REGISTRY_ADD( xQueue, pcQueueName );
+                pxEntryToWrite = &( xQueueRegistry[ ux ] );
                 break;
+            }
+            /* Otherwise, store in the next empty location */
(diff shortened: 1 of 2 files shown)


queue.c
```diff
@@ -2730,24 +2730,37 @@ BaseType_t xQueueIsQueueFullFromISR( const QueueHandle_t xQueue )
...
-                xQueueRegistry[ ux ].xHandle = xQueue;
-
-                traceQUEUE_REGISTRY_ADD( xQueue, pcQueueName );
+                pxEntryToWrite = &( xQueueRegistry[ ux ] );
                 break;
             }
+            /* Otherwise, store in the next empty location */
+            else if( ( NULL == pxEntryToWrite ) && ( xQueueRegistry[ ux ].pcQueueName == NULL ) )
+            {
+                pxEntryToWrite = &( xQueueRegistry[ ux ] );
```
ravibhagavandas, line 2748: Need to break here as well, otherwise it can cause a hole to be not filled.

queue.c
```diff
@@ -2730,24 +2730,37 @@ BaseType_t xQueueIsQueueFullFromISR( const QueueHandle_t xQueue )
...
+            }
             else
             {
                 mtCOVERAGE_TEST_MARKER();
             }
         }
+
+        if( NULL != pxEntryToWrite )
```
paulbartell, line 2756: pxEntryToWrite will be NULL if the entire registry is full.

queue.c
paulbartell: No break statement is necessary here. The code must visit each item in the registry to see if the queue handle has already been added.

queue.c
```diff
@@ -2730,24 +2730,37 @@ BaseType_t xQueueIsQueueFullFromISR( const QueueHandle_t xQueue )
...
-            if( xQueueRegistry[ ux ].pcQueueName == NULL )
+            /* Replace an existing entry if the queue is already in the registry. */
+            if( xQueueRegistry[ ux ].xHandle == xQueue )
             {
-                /* Store the information on this queue. */
-                xQueueRegistry[ ux ].pcQueueName = pcQueueName;
-                xQueueRegistry[ ux ].xHandle = xQueue;
-
-                traceQUEUE_REGISTRY_ADD( xQueue, pcQueueName );
+                pxEntryToWrite = &( xQueueRegistry[ ux ] );
```
paulbartell, line 2742: If the queue handle is already in the registry, the code breaks out of the loop and overwrites the name asssociated with that queue handle.

ravibhagavandas (approved):

n9wxu (approved):

PR #310, opened 2021-04-14
use space to replace tab and remove meaningless space in the end of each line

mysterywolf:
use space to replace tab and remove meaningless space in the end of each line

Description
-----------
1. use Space(4 spaces) to replace Tab to prevent that there are Spaces and Tab mixed in files.
2. remove meaningless space in the end of each line

Test Steps
-----------

Related Issue
-----------
negative 


AniruddhaKanhere:
Hello @mysterywolf,

First of all, thank you for this PR!

All these changes seem to be very harmless. But, there is a catch: in the past, we have tried doing this but it broke some of the toolchains. Some of the toolchains don't handle the changing of tabs to spaces very well.

So, here is what I would recommend as a path forward:
- We can easily accept changes that are not in the `portable` directory like the ones in `include/croutine.h`.
- But for the changes in the portable directory, we would require that it should be done one port at a time. The rationale being that if something breaks, we can roll back the change easily without affecting other changes. And it'll be easier to test and review.
- In essence, there should be multiple PRs touching one port at a time.

Thanks again! Do ask if you have any questions or concerns.

mysterywolf:
OK thanks for you recommend, I will ignore portable directory (except `portable/MemMang`), and deal with other files.

PR #313, opened 2021-04-16
Replace configASSERT( pcQueueName ) in vQueueAddToRegistry with a NULL pointer check.

### LINKED ISSUE #311: new configASSERT breaking existing code
in freertos V10.4.3 in queue.c, an extra ConfigASSERT as been added for pcQueuName.

    void vQueueAddToRegistry( QueueHandle_t xQueue,
                              const char * pcQueueName ) /*lint !e971 Unqualified char types are allowed for strings and single characters only. */
    {
        UBaseType_t ux;

        configASSERT( xQueue );
        configASSERT( pcQueueName );

In previous version the name could be NULL, Is it a necessary to have pcQueueName  defined to non NULL.
My code was running fine before with pcQueueName  NULL, but now I get this assert.

paulbartell:
Replace configASSERT( pcQueueName ) in vQueueAddToRegistry with a NULL pointer check.

Description
-----------
Replace the configASSERT added to vQueueAddToRegistry with a NULL pointer check so that it does not break usage in https://github.com/ARM-software/CMSIS-FreeRTOS.

Test Steps
-----------
Unit test currently in a Pull Request will be updated.

Related Issue
-----------
Fixes #311

### DIFF
--- a/queue.c
+++ b/queue.c
@@ -2728,28 +2728,30 @@ BaseType_t xQueueIsQueueFullFromISR( const QueueHandle_t xQueue )
         UBaseType_t ux;
 
         configASSERT( xQueue );
-        configASSERT( pcQueueName );
 
         QueueRegistryItem_t * pxEntryToWrite = NULL;
 
-        /* See if there is an empty space in the registry.  A NULL name denotes
-         * a free slot. */
-        for( ux = ( UBaseType_t ) 0U; ux < ( UBaseType_t ) configQUEUE_REGISTRY_SIZE; ux++ )
+        if( NULL != pcQueueName )
         {
-            /* Replace an existing entry if the queue is already in the registry. */
-            if( xQueueRegistry[ ux ].xHandle == xQueue )
+            /* See if there is an empty space in the registry.  A NULL name denotes
+             * a free slot. */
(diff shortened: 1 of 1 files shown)


aggarg (approved):

queue.c
```diff
@@ -2728,28 +2728,30 @@ BaseType_t xQueueIsQueueFullFromISR( const QueueHandle_t xQueue )
...
-            {
-                mtCOVERAGE_TEST_MARKER();
+                /* Replace an existing entry if the queue is already in the registry. */
+                if( xQueueRegistry[ ux ].xHandle == xQueue )
+                {
+                    pxEntryToWrite = &( xQueueRegistry[ ux ] );
+                    break;
+                }
+                /* Otherwise, store in the next empty location */
+                else if( ( NULL == pxEntryToWrite ) && ( xQueueRegistry[ ux ].pcQueueName == NULL ) )
```
RichardBarry, line 2747: Please make this line consistent with itself - one side of the && has the NULL on the left side, while the other has the NULL on the right side.  For better or worse the kernel has always used the latter (also relevant to line 2743) although I understand the wisdom of doing it the other way around.

RichardBarry (changes requested):

queue.c
paulbartell: Done.

aggarg (approved):

RichardBarry (approved):

PR #323, opened 2021-05-12
Static allocation and lightweight idle tasks

n9wxu:
Description
-----------
This change adds static allocation for all tasks.  It also creates an idle task for each core.  The first idle task is the same as the legacy idle task and performs housekeeping and calls the vApplicationIdleHook() function.  The remaining idle tasks are very light and only call taskYield if needed.

If static allocation is enabled, the first idle task will call the task allocator just like it has always done.  But each additional idle task will have the minimum stack memory (configMINIMAL_STACK_SIZE) statically allocated from within the start scheduler function.

Test Steps
-----------
This was tested with the XMOS port with 1, 3 and 7 cores active.

### DIFF
--- a/tasks.c
+++ b/tasks.c
@@ -2715,6 +2715,8 @@ void vTaskStartScheduler( void )
             }
         #else  /* if ( configSUPPORT_STATIC_ALLOCATION == 1 ) */
             {
+                if(xCoreID == 0)
+                {
                 /* The Idle task is being created using dynamically allocated RAM. */
                 xReturn = xTaskCreate( prvIdleTask,
                                        cIdleName,
@@ -2722,6 +2724,18 @@ void vTaskStartScheduler( void )
                                        ( void * ) NULL,
                                        portPRIVILEGE_BIT,             /* In effect ( tskIDLE_PRIORITY | portPRIVILEGE_BIT ), but tskIDLE_PRIORITY is zero. */
                                        &xIdleTaskHandle[ xCoreID ] ); /*lint !e961 MISRA exception, justified as it is not a redundant explicit cast to all supported compilers. */
+                }
+                #if( configNUM_CORES > 1 )
+                else
+                {
+                xReturn = xTaskCreate( prvMinimalIdleTask,
+                                       cIdleName,
+                                       configMINIMAL_STACK_SIZE,
+                                       ( void * ) NULL,
+                                       portPRIVILEGE_BIT,             /* In effect ( tskIDLE_PRIORITY | portPRIVILEGE_BIT ), but tskIDLE_PRIORITY is zero. */
+                                       &xIdleTaskHandle[ xCoreID ] ); /*lint !e961 MISRA exception, justified as it is not a redundant explicit cast to all supported compilers. */
+                }
+                #endif
             }
         #endif /* configSUPPORT_STATIC_ALLOCATION */
     }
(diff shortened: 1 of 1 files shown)


tasks.c
```diff
@@ -467,6 +467,9 @@ static void prvInitialiseTaskLists( void ) PRIVILEGED_FUNCTION;
  *
  */
 static portTASK_FUNCTION_PROTO( prvIdleTask, pvParameters ) PRIVILEGED_FUNCTION;
+#if ( configNUM_CORES > 1 )
+static portTASK_FUNCTION_PROTO( prvMinimalIdleTask, pvParameters ) PRIVILEGED_FUNCTION;
```
RichardBarry, line 471: There is no need to use portTASK_FUNCTION_PROTO - that is only useful on tiny parts or when a compiler has special requirements for functions that implement tasks - neither is going to be the case for SMP ports.  This can just be:
static void prvMinimalIdleTask( void *pvParameters ) PRIVILEGED_FUNCTION;

tasks.c
```diff
@@ -1545,7 +1548,11 @@ static void prvInitialiseNewTask( TaskFunction_t pxTaskCode,
     pxNewTCB->xTaskRunState = taskTASK_NOT_RUNNING;
 
     /* Is this an idle task? */
+    #if(configNUM_CORES > 1)
+    pxNewTCB->xIsIdle = ( pxTaskCode == prvIdleTask ) || (pxTaskCode == prvMinimalIdleTask);
```
RichardBarry, line 1552: To comply with the coding standard this will have to be written out long hand, rather than all on one line.

tasks.c
```diff
@@ -2663,22 +2670,40 @@ void vTaskStartScheduler( void )
...
+                                                                    portPRIVILEGE_BIT,     /* In effect ( tskIDLE_PRIORITY | portPRIVILEGE_BIT ), but tskIDLE_PRIORITY is zero. */
+                                                                    pxIdleTaskStackBuffer,
+                                                                    pxIdleTaskTCBBuffer ); /*lint !e961 MISRA exception, justified as it is not a redundant explicit cast to all supported compilers. */
+                }
+                #if( configNUM_CORES > 1)
+                else
+                {
+                    struct taskMemory{
+                        StaticTask_t TCB;
+                        StackType_t stack[configMINIMAL_STACK_SIZE];
```
RichardBarry, line 2695: If these tasks cannot contain user code (don't call an idle hook), then I think it is ok for them all to have the same stack size (so no requested change).  Just watch the variable/struct naming convention.  We may want to look at breaking this whole function up a bit as it was already quite long, and is [obviously] longer now.  Perhaps a prvCreateIdleTasks() function to contain this part.

tasks.c
```diff
@@ -4146,6 +4185,61 @@ void vTaskMissedYield( void )
...
+ *
+ * void prvMinimalIdleTask( void *pvParameters );
+ *
+ * The minimal idle task is used for all the additional Cores in a SMP system.
+ * There must be only 1 idle task and the rest are minimal idle tasks.
+ * 
+ * @todo additional conditional compiles to remove this function.
+ */
+#if (configNUM_CORES > 1)
+static portTASK_FUNCTION( prvMinimalIdleTask, pvParameters )
```
RichardBarry, line 4204: Same comment ref no need to have the portTASK_FUNCTION macro.

RichardBarry (changes requested):

tasks.c
n9wxu: prvCreateIdleTasks makes sense.  I will go ahead and do it.

tasks.c
n9wxu: Per our discussion, I left this use of the macro.  Some ports still need this macro to decorate the function.

dan4thewin (approved):

PR #328, opened 2021-05-19
queue.c: Change some asserts into conditionals and improve overflow checks

dan4thewin:
This PR is paired with changes to the CBMC queue proofs: https://github.com/FreeRTOS/FreeRTOS/pull/603

These changes allow certain CBMC proofs to be written with few or no assumptions.  The overflow checks now preempt an overflow, rather than checking for the consequence of an overflow.


queue.c
```diff
@@ -369,8 +375,7 @@ BaseType_t xQueueGenericReset( QueueHandle_t xQueue,
         }
         else
         {
-            traceQUEUE_CREATE_FAILED( ucQueueType );
-            mtCOVERAGE_TEST_MARKER();
```
AniruddhaKanhere, line 373: Why was this removed? It seems that it is there for logging/tracing. Or am I missing something?

queue.c
```diff
@@ -369,8 +375,7 @@ BaseType_t xQueueGenericReset( QueueHandle_t xQueue,
         }
         else
         {
-            traceQUEUE_CREATE_FAILED( ucQueueType );
-            mtCOVERAGE_TEST_MARKER();
+            configASSERT( pdFAIL );
```
AniruddhaKanhere, line 378: This `configASSERT` would not help while debugging. In the previous source, one could simply see which `configASSERT` (with a proper implementation using `__FILE__` and `__LINE__`) was failing and find the issue.
Now one would need to print out the values or step through the code with debugger. Shall we try to find a better way?

queue.c
```diff
@@ -385,56 +390,59 @@ BaseType_t xQueueGenericReset( QueueHandle_t xQueue,
...
+            {
+                traceQUEUE_CREATE_FAILED( ucQueueType );
+                mtCOVERAGE_TEST_MARKER();
+            }
         }
         else
         {
-            traceQUEUE_CREATE_FAILED( ucQueueType );
-            mtCOVERAGE_TEST_MARKER();
+            configASSERT( pdFAIL );
```
AniruddhaKanhere, line 445: Same comment as above. This might not be very useful while debugging.

queue.c
```diff
@@ -719,24 +727,26 @@ static void prvInitialiseNewQueue( const UBaseType_t uxQueueLength,
...
+
+                traceCREATE_COUNTING_SEMAPHORE();
+            }
+            else
+            {
+                traceCREATE_COUNTING_SEMAPHORE_FAILED();
+            }
         }
 
+        configASSERT( xHandle );
```
AniruddhaKanhere, line 749: Here as well. What failed, one cannot say just with this assert.

queue.c
```diff
@@ -750,22 +760,24 @@ static void prvInitialiseNewQueue( const UBaseType_t uxQueueLength,
...
```
AniruddhaKanhere, line 780: And here too.

AniruddhaKanhere (changes requested):
Please see the comments.
I am not sure whether replacing multiple asserts with just one assert would benefit users.

queue.c
dan4thewin: This was the approach Richard suggested.  Changing the asserts to conditionals means there is still protection when assertions are compiled out, but loses debugging information as you point out.  Adding in the trailing assert is a compromise to provide some information without repeating every check.

queue.c
dan4thewin: This case of `pxQueue` being `NULL` cannot be reached, and the trace event seems to belong to the dynamic case.

queue.c
dan4thewin: see my previous reply

queue.c
dan4thewin: see my previous reply

queue.c
dan4thewin: see my previous reply

queue.c
dan4thewin: I'm assuming you mean that we could have an else block where the tests are repeated as asserts.  Manually keeping the two sets of tests in sync strikes me as a big drawback.  Let's get Richard's opinion.

paulbartell: This is a great place for an ASSERT_CONTINUE macro. The idea would be to log the assertion to NV storage or or via a debugging port, but not go into the typical infinite loop. This also lets customers test the edge cases in their applications without entering an infinite loop.

paulbartell:
Note: the unit tests for queue.c need to be updated over in the FreeRTOS/FreeRTOS repository ( https://github.com/FreeRTOS/FreeRTOS/tree/main/FreeRTOS/Test/CMock/queue ).

To validate that they run correctly on github, I would suggest making a branch in the FreeRTOS/FreeRTOS repository with the Source submodule pointer set to the tip of this branch. Then update the tests as necessary and disregard any failures displayed for unit tests in this PR.

Alternatively, you could modify the github workflow for unit tests in this repository to check for a branch with the same name from the same user's fork and use that branch is available.

markrtuttle (dismissed):

RichardBarry:
Brief answer for now due to it being late - I need to do a deeper dive into the PR before a full answer.

My understanding is that the prior asserts at the top of the file made the CBMC memory safety checks cumbersome and potentially long running - hence they were replaced with inline checks that are much more CBMC friendly.  The new assert at the end then made it backward compatible for code that relied on the assert behavior.

I understand the desire to make it easier to debug - debuggability is a very important attribute.  Adding the asserts() back at the front would probably re-introduce the CBMC issue this PR is trying to improve, plus it would look cumbersome to effectively have the same checks twice.  It may be possible to replace the single assert at the end with something that provides more information to developers, but as per, I would need to look a bit deeper before having an opinion on that.

queue.c
```diff
@@ -268,8 +268,13 @@ BaseType_t xQueueGenericReset( QueueHandle_t xQueue,
 
     configASSERT( pxQueue );
 
-    taskENTER_CRITICAL();
+    if( ( pxQueue != NULL ) &&
+        ( pxQueue->uxLength >= 1 ) &&
+        /* Check for multiplication overflow. */
+        ( ( SIZE_MAX / pxQueue->uxLength ) >= pxQueue->uxItemSize ) )
```
RichardBarry, line 274: uxLength and uxItemSize are both UBaseType_t, rather than size_t, so this assumes sizeof( UBaseType_t ) == sizeof( size_t ) -> can we create a local constant that is equivalent to UBaseType_t_MAX?

queue.c
```diff
@@ -268,8 +268,13 @@ BaseType_t xQueueGenericReset( QueueHandle_t xQueue,
 
     configASSERT( pxQueue );
 
-    taskENTER_CRITICAL();
+    if( ( pxQueue != NULL ) &&
+        ( pxQueue->uxLength >= 1 ) &&
```
RichardBarry, line 272: Should put a "U" on the 1, as per 1U.

queue.c
```diff
@@ -306,8 +311,12 @@ BaseType_t xQueueGenericReset( QueueHandle_t xQueue,
             vListInitialise( &( pxQueue->xTasksWaitingToSend ) );
             vListInitialise( &( pxQueue->xTasksWaitingToReceive ) );
         }
+        taskEXIT_CRITICAL();
+    }
+    else
+    {
+        configASSERT( pdFAIL );
```
RichardBarry, line 318: The problem with asserting on "pdFAIL" is that the condition is always true so this will generate compiler warnings.  That is - the assert test will always trigger (which is the intention of course), and there is no ASSERT_ALWAYS() macro.

The previous version of this function only asserted when pxQueue was NULL, and that assert still exists at the top of the file, so that behaviour has not changed.  Now however the code will also assert if either of the two additional checks fail.  Perhaps introduce a status variable that is set to pdPASS in the happy path, and pdFAIL on what is currently line 318 - then assert on and return the status variable at the end of the function:

```
else
{
    xReturn = pdFAIL;
}

configASSERT( xReturn != pdFAIL );
return xReturn;

```

queue.c
RichardBarry: Same comment as in the queue reset function about using a constant as the parameter in the configASSERT() macro - will generate compiler warnings ("condition is always true").  Extending Paul's thoughts above, what we really need is an assert always function.  I think, also as per my comments in the queue reset function, using a status variable that gets set to pass or fail, then asserting on the status variable just before exiting the function would be fine.  I understand that loses information when debugging, but if you see that line is failing I think it would be possible to see why from within the debugger.

queue.c
RichardBarry, line 445: Same comment here on asserting on an always true test.

RichardBarry (changes requested):

queue.c
dan4thewin: fixed

queue.c
dan4thewin: I think SIZE_MAX or PTRDIFF_MAX are reasonable choices.  Given that the `pxQueue->uxLength` and `pxQueue->uxItemSize` are used to do pointer arithmetic and also with `pvPortMalloc()`, we want the product of these to always be  <= to the largest theoretical object.

queue.c
dan4thewin: changed

queue.c
dan4thewin: changed to assert on the return value

queue.c
dan4thewin: changed

RichardBarry (approved):

AniruddhaKanhere (approved):

PR #338, opened 2021-05-29
use configSTACK_DEPTH_TYPE consequently

feilipu:
Use configSTACK_DEPTH_TYPE consequently.
-----------

Previously the constant `configSTACK_DEPTH_TYPE` was introduced to support stacks larger than `UBaseType_t` in 8-bit machines.
This PR uses the `configSTACK_DEPTH_TYPE` type constant to consequently set the stack size report variable to the appropriate (smallest correct) type.

Extend the use of `configSTACK_DEPTH_TYPE` which enables developers to define the type used to hold stack usage counters. Defaults to uint16_t for backward compatibility. #define configSTACK_DEPTH_TYPE to a type (for example, `uint64_t`) in FreeRTOSConfig.h to override the default.

__EDIT__ Revised the various names for stack variables under the standardised naming convention.

This now aligns with #350, so that the stack depth type is also configurable.

Related Issue
-----------
Previously discussed and `configSTACK_DEPTH_TYPE` constant [introduced here](https://github.com/FreeRTOS/FreeRTOS-Kernel/commit/c6de0001fa89664f21b1f2a03f87c2539c193766).

These changes have been in Arduino FreeRTOS since Jan 2018, as noted [here](https://github.com/feilipu/Arduino_FreeRTOS_Library/pull/29) and in later commits.


feilipu:
Hi @RichardBarry could you get your pr-bar-raiser to take a look please?
Thanks, :pray:

include/FreeRTOS.h
```diff
@@ -900,9 +900,9 @@
 
 #ifndef configSTACK_DEPTH_TYPE
 
-/* Defaults to uint16_t for backward compatibility, but can be overridden
+/* Defaults to uint32_t for compatibility, but can be overridden
  * in FreeRTOSConfig.h if uint16_t is too restrictive. */
-    #define configSTACK_DEPTH_TYPE    uint16_t
+    #define configSTACK_DEPTH_TYPE    uint32_t
```
RichardBarry, line 905: This constant is to ensure dropping the code into a a project that doesn't define configSTACK_DEPTH will work as per prior versions - so uint16_t was the correct value.  https://github.com/FreeRTOS/FreeRTOS-Kernel/blob/V8.0.0/FreeRTOS/Source/tasks.c#L515

include/task.h
```diff
@@ -358,8 +358,8 @@ typedef enum
  * task. h
  * <pre>
 * TaskHandle_t xTaskCreateStatic( TaskFunction_t pxTaskCode,
- *                               const char *pcName,
- *                               uint32_t ulStackDepth,
+ *                               const char * const pcName,
+ *                               const configSTACK_DEPTH_TYPE ulStackDepth,
```
RichardBarry, line 362: This function always used uint32_t so this change is not consistent with configSTACK_DEPTH_TYPE defaulting to uin16_t, as per my previous comment.

include/task.h
```diff
@@ -465,7 +465,7 @@ typedef enum
 #if ( configSUPPORT_STATIC_ALLOCATION == 1 )
     TaskHandle_t xTaskCreateStatic( TaskFunction_t pxTaskCode,
                                     const char * const pcName,     /*lint !e971 Unqualified char types are allowed for strings and single characters only. */
-                                    const uint32_t ulStackDepth,
+                                    const configSTACK_DEPTH_TYPE ulStackDepth,
```
RichardBarry, line 468: As per my comment for the prototype of this function in the code comments.

include/task.h
```diff
@@ -1652,7 +1652,7 @@ configSTACK_DEPTH_TYPE uxTaskGetStackHighWaterMark2( TaskHandle_t xTask ) PRIVIL
      */
     void vApplicationGetIdleTaskMemory( StaticTask_t ** ppxIdleTaskTCBBuffer,
                                                StackType_t ** ppxIdleTaskStackBuffer,
-                                               uint32_t * pulIdleTaskStackSize ); /*lint !e526 Symbol not defined as it is an application callback. */
+                                               configSTACK_DEPTH_TYPE * pulIdleTaskStackSize ); /*lint !e526 Symbol not defined as it is an application callback. */
```
RichardBarry, line 1655: As per my comment for the prototype in the code comments.

include/task.h
```diff
@@ -1641,7 +1641,7 @@ configSTACK_DEPTH_TYPE uxTaskGetStackHighWaterMark2( TaskHandle_t xTask ) PRIVIL
 #if ( configSUPPORT_STATIC_ALLOCATION == 1 )
     /**
      * task.h
-     * <pre>void vApplicationGetIdleTaskMemory( StaticTask_t ** ppxIdleTaskTCBBuffer, StackType_t ** ppxIdleTaskStackBuffer, uint32_t *pulIdleTaskStackSize ) </pre>
+     * <pre>void vApplicationGetIdleTaskMemory( StaticTask_t ** ppxIdleTaskTCBBuffer, StackType_t ** ppxIdleTaskStackBuffer, configSTACK_DEPTH_TYPE * pulIdleTaskStackSize ) </pre>
```
RichardBarry, line 1644: Here too - I think this was always uint32_t so should not use th econfigSTACK_DEPTH_TYPE constant.

include/timers.h
```diff
@@ -1330,7 +1330,7 @@ BaseType_t xTimerGenericCommand( TimerHandle_t xTimer,
 
     /**
      * task.h
-     * <pre>void vApplicationGetTimerTaskMemory( StaticTask_t ** ppxTimerTaskTCBBuffer, StackType_t ** ppxTimerTaskStackBuffer, uint32_t *pulTimerTaskStackSize ) </pre>
+     * <pre>void vApplicationGetTimerTaskMemory( StaticTask_t ** ppxTimerTaskTCBBuffer, StackType_t ** ppxTimerTaskStackBuffer, configSTACK_DEPTH_TYPE * pulTimerTaskStackSize ) </pre>
```
RichardBarry, line 1333: This too was always uint32_t.

include/timers.h
```diff
@@ -1341,7 +1341,7 @@ BaseType_t xTimerGenericCommand( TimerHandle_t xTimer,
      */
     void vApplicationGetTimerTaskMemory( StaticTask_t ** ppxTimerTaskTCBBuffer,
                                           StackType_t ** ppxTimerTaskStackBuffer,
-                                              uint32_t * pulTimerTaskStackSize );
+                                          configSTACK_DEPTH_TYPE * pulTimerTaskStackSize );
```
RichardBarry, line 1344: As per comment on the prototype for this function in the code comments.

RichardBarry (changes requested):
I didn't make the same comment on each change, but basically the configSTACK_DEPTH_TYPE default should match the prior code, which means uint16_t, which then invalidates the use of the constant where the data type has always been uint32_t.  There are some updates to the code comments that should be made though.

feilipu:
@RichardBarry thank you for the detailed comments.

The purpose of this PR is indeed to change the data type to a known **configurable** value for the size of stack return.

Whether this `configSTACK_DEPTH_TYPE` defaults to `uint16_t` or `uint32_t` is of little importance. I set the default to `uint32_t` to pass the tests. It can be set to `uint16_t` if the tests are updated to suit.

> the use of the constant where the data type has always been uint32_t

In my opinion, this is bug. Simply setting the `configSTACK_DEPTH_TYPE`  to `uint64_t` would demonstrate it as such.
The return stack type needs to be consistently configurable across all of its uses.

RichardBarry:
Perhaps I should give some background.

Original ports of FreeRTOS were to 8 and 16-bit chips, hence using uint16_t to hold stack sizes made sense.  uint16_t soon proved too small as FreeRTOS became popular on 32 and now even 64-bit chips though.  Hence newer functions, like xTaskCreateStatic(), used uint32_t instead of uint16_t and there were feature requests to likewise change to uint32_t in the older functions too.  However FreeRTOS has a policy of not breaking backward compatibility unless absolutely necessary (which I think has only been a couple of times in its history).  That, in part, is achieved using the [configENABLE_BACKWARD_COMPATABILITY](https://github.com/FreeRTOS/FreeRTOS-Kernel/blob/main/include/FreeRTOS.h#L963) constant.  That defaults to 1, so the default is backward compatibility.  All constants that are enabled by configENABLE_BACKWARD_COMPATABILITY must default to a value that makes the code behave as it did before a change was made, so configSTACK_DEPTH_TYPE must be uint16_t, otherwise there is no point to the constant.  When it is uint16_t all function prototypes have the exact same data types as before configSTACK_DEPTH_TYPE was defined.

All constants that start "config" are intended to be user definable in FreeRTOSConfig.h, so if you want all the stack size variables to be of size configSTACK_DEPTH_TYPE then it is intended you define it as such in FreeRTOSConfig.h.

feilipu:
Thanks, and I understand the reasons for, and history of, compatibility.

Would it be more acceptable to wrap the `uint32_t` data type definitions in `configENABLE_BACKWARD_COMPATABILITY`, and have the correct `configSTACK_DEPTH_TYPE` data type definitions as the non-backwards compatible alternative?

If so, I can update the PR to that end.

include/FreeRTOS.h
feilipu: Default reverted to `uint16_t`.

n9wxu:
There are many good ideas here but many of the changes will break backwards compatibility more than we can address with configENABLE_BACKWARD_COMPATIBILITY.  Your PR did point out a big issue with mpu_xTaskCreate() that will need to be addressed.  I am going to close this PR but we will continue to keep the stack types in mind when we can break backward compatibility.

PR #344, opened 2021-06-06
Add T-HEAD's RISC-V CPU E906, Support E906(rv32imac), E906F(rv32imafc),E906FD(rv32imafdc)

wormsX:
[New CPU]Add Alibaba T-HEAD's RISC-V CPU E906, Support E906(rv32imac), E906F(rv32imafc),E906FD(rv32imafdc)

Description
-----------
T-head Semiconductor Co., Ltd. (Alibaba), the latest commercial RISCV CPU series, The same code, supports 3 types isa： rv32imac, rv32imafc, rv32imafdc. Between different isa, separated by the built-in macro(__riscv_flen) of the toolchain.

Test Steps
-----------

Related Issue
-----------

### DIFF
--- a/portable/GCC/RISC-V/chip_specific_extensions/THEAD_RV32/freertos_risc_v_chip_specific_extensions.h
+++ b/portable/GCC/RISC-V/chip_specific_extensions/THEAD_RV32/freertos_risc_v_chip_specific_extensions.h
@@ -1,0 +1,210 @@
+/*
+ * FreeRTOS Kernel V10.4.0
+ * Copyright (C) 2020 Amazon.com, Inc. or its affiliates.  All Rights Reserved.
+ *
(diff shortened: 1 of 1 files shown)


RichardBarry:
Thanks for this.  If I understand correctly this is adding floating point support.  Is there anything non-standard about the floating point code in your chip?  If not, would we be better off moving this code to be a build option within the main port rather than as a header file that implements an extension?  That way it would work for all compliant RISC-V parts.

wormsX:
s

> Thanks for this. If I understand correctly this is adding floating point support. Is there anything non-standard about the floating point code in your chip? If not, would we be better off moving this code to be a build option within the main port rather than as a header file that implements an extension? That way it would work for all compliant RISC-V parts.

Thank you very much for your reply, I very much agree with your point of view.
This part of the code is about single-precision/double-precision floating point, then I will re-move the code to the main port

PR #346, opened 2021-06-07
Posix: fix event_wait_timed()

twpedersen:
event_wait_timed() was ignoring a timeout of 1000 ms.
Presumably this is because pthread_cond_timedwait() only
considers tv_nsec less than one second.

Convert the timeout in miliseconds to second and nanosecond
components to fix this.

### DIFF
--- a/portable/ThirdParty/GCC/Posix/utils/wait_for_event.c
+++ b/portable/ThirdParty/GCC/Posix/utils/wait_for_event.c
@@ -76,8 +76,8 @@ bool event_wait_timed( struct event * ev,
     int ret = 0;
 
     clock_gettime( CLOCK_REALTIME, &ts );
-    //ts.tv_sec += ms;
-    ts.tv_nsec += (ms * 1000000);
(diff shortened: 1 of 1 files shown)


mingyue86010 (approved):

alfred2g (approved):

n9wxu (approved):

portable/ThirdParty/GCC/Posix/utils/wait_for_event.c
```diff
@@ -76,8 +76,8 @@ bool event_wait_timed( struct event * ev,
     int ret = 0;
 
     clock_gettime( CLOCK_REALTIME, &ts );
-    //ts.tv_sec += ms;
-    ts.tv_nsec += (ms * 1000000);
+    ts.tv_sec += ms / 1000;
+    ts.tv_nsec += ((ms % 1000) * 1000000);
```
alfred2g, line 80: Nice catch.. thanks!

Issue #312, opened 2021-04-15
[Feature Request] Enable each stream buffer to use a different sbSEND_COMPLETED implementation


RichardBarry:
**Is your feature request related to a problem? Please describe.**
sbSEND_COMPLETED is the macro that performs any processing necessary after data has been sent to a stream buffer.  By default this will unblock a task if a task is waiting for data from the stream buffer.  The application writer can provide their own definition that, for example, can raise an interrupt in another core when the stream buffer is used for core to core communication.  However currently a single global definition of sbSEND_COMPLETED applies to every stream buffer, enforcing each stream buffer to behave in the same way.  

**Describe the solution you'd like**
While the above can be worked around within the application it would be an improvement if each stream buffer can define its own sbSEND_COMPLETED equivalent, and only use the default implementation when  a stream buffer specific implementation is not supplied.

**Additional context**
See https://forums.freertos.org/t/messagebuffers-for-core-to-core-communication/12318

eddie9712:
Hi @RichardBarry , I've sent a PR related to this issue would you please check if the solution is that you would like.

RichardBarry:
Yes - thanks - I've checked out the branch already but git is telling me I have local changes in my repo (which I don't) so so far I've only done a visual diff.

eddie9712:
Ok, If there is anything I can help please let me know :)

kar-rahul-aws:
Closing this request, since PR #437 which addresses this request is merged.

PR #360, opened 2021-06-27
Let each stream/message can use its own sbSEND_COMPLETED

eddie9712:
Description
-----------
According to issue #312, this PR provides `configUSE_SB_COMPLETED_CALLBACK` that makes each stream/message buffer use its own `sbSEND_COMPLETED`

Test Steps
-----------
In `FreeRTOSconfig.h`,
If `configUSE_SB_COMPLETED_CALLBACK` is not set or set to 0, the single global definition of `sbSEND_COMPLETED` applies to every stream/message buffer.
If `configUSE_SB_COMPLETED_CALLBACK` is set to 1, each stream/mesasage buffer can define a callback function as its own `sbSEND_COMPLETED`, Users can pass the callback function as the argument(if pass NULL, it will use the default implementaion) when the buffer is (static)created:
```
void func1()
{
  /* User-defined sbSEND_SOMPLETED */
}
xStreamBuffer = xStreamBufferCreate( xBufferSizeBytes, xTriggerLevelBytes, func1)
```
The default type of the callback fucntion is "void function()", users can change the type of the callback function in `stream_buffer.h`:
```
#if (configUSE_SB_COMPLETED_CALLBACK == 1)
    typedef void (* StreamBufferCallbackFunction_t)();   //Declare a function pointer that points to the stream buffer's callback
#endif
```
* Validated with STM32F407VG

Related Issue
-----------
Issue #312

### DIFF
--- a/include/stream_buffer.h
+++ b/include/stream_buffer.h
@@ -70,7 +70,9 @@
  */
 struct StreamBufferDef_t;
 typedef struct StreamBufferDef_t * StreamBufferHandle_t;
-
+#if ( configUSE_SB_COMPLETED_CALLBACK == 1 )
+    typedef void (* StreamBufferCallbackFunction_t)(); /*Declare a function pointer that points to the stream buffer's callback */
+#endif
 
 /**
  * message_buffer.h
@@ -102,6 +104,10 @@ typedef struct StreamBufferDef_t * StreamBufferHandle_t;
  * are actually available.  Setting a trigger level of 0 will result in a
  * trigger level of 1 being used.  It is not valid to specify a trigger level
  * that is greater than the buffer size.
+ *
+ * @param StreamBufferCallbackFunction_t If configUSE_SB_COMPLETED_CALLBACK is set to 1,
+ * the callback function name should be passed for this parameter. If NULL is passed, it will use the default
+ * implementation of sbSEND_COMPLETED
  *
  * @return If NULL is returned, then the stream buffer cannot be created
  * because there is insufficient heap memory available for FreeRTOS to allocate
@@ -137,7 +143,13 @@ typedef struct StreamBufferDef_t * StreamBufferHandle_t;
  * \defgroup xStreamBufferCreate xStreamBufferCreate
  * \ingroup StreamBufferManagement
  */
-#define xStreamBufferCreate( xBufferSizeBytes, xTriggerLevelBytes )    xStreamBufferGenericCreate( xBufferSizeBytes, xTriggerLevelBytes, pdFALSE )
+
+#if ( configUSE_SB_COMPLETED_CALLBACK == 1 )
+    #define xStreamBufferCreate( xBufferSizeBytes, xTriggerLevelBytes, pxStreamBufferCallbackFunction ) \
+    xStreamBufferGenericCreate( xBufferSizeBytes, xTriggerLevelBytes, pdFALSE, pxStreamBufferCallbackFunction )
+#else
+    #define xStreamBufferCreate( xBufferSizeBytes, xTriggerLevelBytes )    xStreamBufferGenericCreate( xBufferSizeBytes, xTriggerLevelBytes, pdFALSE )
+#endif
 
 /**
  * stream_buffer.h
@@ -179,6 +191,10 @@ typedef struct StreamBufferDef_t * StreamBufferHandle_t;
  * StaticStreamBuffer_t, which will be used to hold the stream buffer's data
  * structure.
  *
+ * @param StreamBufferCallbackFunction_t If configUSE_SB_COMPLETED_CALLBACK is set to 1,
+ * the callback function name should be passed for this parameter. If NULL is passed, it will use the default
+ * implementation of sbSEND_COMPLETED
+ *
  * @return If the stream buffer is created successfully then a handle to the
  * created stream buffer is returned. If either pucStreamBufferStorageArea or
  * pxStaticstreamBuffer are NULL then NULL is returned.
@@ -218,8 +234,14 @@ typedef struct StreamBufferDef_t * StreamBufferHandle_t;
  * \defgroup xStreamBufferCreateStatic xStreamBufferCreateStatic
  * \ingroup StreamBufferManagement
  */
-#define xStreamBufferCreateStatic( xBufferSizeBytes, xTriggerLevelBytes, pucStreamBufferStorageArea, pxStaticStreamBuffer ) \
+
+#if ( configUSE_SB_COMPLETED_CALLBACK == 1 )
+    #define xStreamBufferCreateStatic( xBufferSizeBytes, xTriggerLevelBytes, pucStreamBufferStorageArea, pxStaticStreamBuffer, pxStreamBufferCallbackFunction ) \
+    xStreamBufferGenericCreateStatic( xBufferSizeBytes, xTriggerLevelBytes, pdFALSE, pucStreamBufferStorageArea, pxStaticStreamBuffer, pxStreamBufferCallbackFunction )
+#else
+    #define xStreamBufferCreateStatic( xBufferSizeBytes, xTriggerLevelBytes, pucStreamBufferStorageArea, pxStaticStreamBuffer ) \
     xStreamBufferGenericCreateStatic( xBufferSizeBytes, xTriggerLevelBytes, pdFALSE, pucStreamBufferStorageArea, pxStaticStreamBuffer )
+#endif
 
 /**
  * stream_buffer.h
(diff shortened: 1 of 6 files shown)


RichardBarry:
Thanks for this - it is very helpful.  Here are my initial thoughts:

1) We can probably do this without introducing the configUSE_SB_COMPLETED_CALLBACK constant.  In particular at the moment the signature of the xStreamBufferCreate() function (really a macro) changes depending on whether configUSE_SB_COMPLETED_CALLBACK is defined or not, which could cause issues.

2) I like having the additional parameter to xStreamBufferCreate() to set the callback - however (for better or worse) we try and maintain backward compatibility and previous FreeRTOS versions don't have that parameter.  There would be two options to maintain backward compatibility.  

- First would be to keep the exact existing behavior whereby creating a steam buffer defaults to using the "send complete" macro and then introduce a new function that sets a callback should you want one.  In this case creating a stream buffer with a callback would require two calls - the first to create the stream buffer and obtain its handle, the second to pass the handle and a callback into a new (not yet existing) function that sets the callback for that stream buffer.  

- The other option would also keep the existing behavior if you call xStreamBufferCreate(), but then introduce a new macro xStreamBufferCreateWithCallback() (or whatever name we choose) that has the additional parameter as per your PR.  The xStreamBufferGenericCreate() then handles the callback as per your PR.

My preference is probably the second option.

eddie9712:
Yes, thanks for your suggestion, I think the second option has the better backward compatibility, so I have added some new macros just like "xStreamBufferCreateWithCallBack()", if the the configUSE_SB_COMPLETED_CALLBACK is set to 1 then we can use these macro. However, the buffers created by "xStreamBufferCreate()" will use the default sbSEND_COMPLETED(do xTaskNotify).

include/stream_buffer.h
```diff
@@ -137,7 +143,15 @@ typedef struct StreamBufferDef_t * StreamBufferHandle_t;
  * \defgroup xStreamBufferCreate xStreamBufferCreate
  * \ingroup StreamBufferManagement
  */
-#define xStreamBufferCreate( xBufferSizeBytes, xTriggerLevelBytes )    xStreamBufferGenericCreate( xBufferSizeBytes, xTriggerLevelBytes, pdFALSE )
+
+#if ( configUSE_SB_COMPLETED_CALLBACK == 1 )
+    #define xStreamBufferCreate( xBufferSizeBytes, xTriggerLevelBytes ) \
+        xStreamBufferGenericCreate( xBufferSizeBytes, xTriggerLevelBytes, pdFALSE, NULL)
+    #define xStreamBufferCreate( xBufferSizeBytes, xTriggerLevelBytes, pxStreamBufferCallbackFunction ) \
```
RichardBarry, line 150: I'm getting errs now with things like this macro being defined twice.  I will make some suggested edits then post back.

RichardBarry (changes requested):

RichardBarry:
I'm getting some build errors now, as per the comment I left in the code.  I will make some suggested edits then push back to your branch for your review.

RichardBarry:
Here is the edited code: https://github.com/RichardBarry/FreeRTOS-Kernel/tree/eddie9712-new_branch

Notes:

- This is backward compatible no matter what the #define settings.
- Code now builds in all combinations.
- For now I removed the #define used to build in or out the ability to set a stream buffer callback.  My intention was to put it back, but in a more limited way - namely just to build in or out the additional members of the stream buffer structure, however....
- Having tried to use this I think I'm going to change my mind and prefer the method of adding in functions so you can create the stream buffer and then use new functions to set the callback.  The reason is I think the callback needs a parameter so you can use the same callback for multiple stream buffers, and if there is a send complete callback there should also be a receive complete callback also with a parameter.  That makes four additional parameters rather than 1, which is a bit unwieldy.  

Would you like to have a go at leaving the xStreamBufferCreate() function the same as it was, and adding in an additional function to set is send complete and receive complete callbacks?  If you create a stream buffer without subsequently setting its callbacks then it would behave as before and just call the sbSEND_COMPLETE macro.

If you don't want to do the above I will have a go myself.

include/stream_buffer.h
eddie9712: I think I forgot to add the "WithCallback" to the second #define,  what I would like should be:
```
#if ( configUSE_SB_COMPLETED_CALLBACK == 1 )
    #define xStreamBufferCreate( xBufferSizeBytes, xTriggerLevelBytes ) \
        xStreamBufferGenericCreate( xBufferSizeBytes, xTriggerLevelBytes, pdFALSE, NULL)
    #define xStreamBufferCreateWithCallback( xBufferSizeBytes, xTriggerLevelBytes, pxStreamBufferCallbackFunction ) \
```

eddie9712:
Sure, I think it's a good idea to add an additional function "xStreambufferCreateWithCallback()" directly instead of using "configUSE_SB_COMPLETED_CALLBACK" as you had suggested. However, there is something confuses me:

- As you mentioned above:
> That makes four additional parameters rather than 1, which is a bit unwieldy.

I don't understand why we need 4 additional parameters for the buffer created(the callback for send complete and receive complete just 2 parameters).

- With regard to receive complete, how about add a callback for parameter just like what we had done to send completed? (which means we will pass two callbacks for send complete and receive complete) If this is what you would like, I'll have a go with it.
```
#define xMessageBufferCreate( xBufferSizeBytes ) \
    ( MessageBufferHandle_t ) xStreamBufferGenericCreate( xBufferSizeBytes, ( size_t ) 0, pdTRUE, NULL)
#define xMessageBufferCreateWithCallback( xBufferSizeBytes, pxStreamBufferSendCallbackFunction,  pxStreamBufferReceiveCallbackFunction) \
    ( MessageBufferHandle_t ) xStreamBufferGenericCreate( xBufferSizeBytes, ( size_t ) 0, pdTRUE,  pxStreamBufferSendCallbackFunction,  pxStreamBufferReceiveCallbackFunction)
```

RichardBarry:
> I don't understand why we need 4 additional parameters for the buffer created(the callback for send complete and receive complete just 2 parameters).

Currently we have one additional parameter, which is a pointer to a function with the following prototype:

`void StreamBufferCallbackFunction( void );`

That function does not take any parameters, so the implementation of the function will not know which stream buffer object called it.  That may lead to needing a separate callback implementation for each stream buffer.  If the callback took a parameter then the parameter can be used to determine which stream buffer called it.  The parameter could be a pointer to the calling stream buffer, as per.....

`void StreamBufferCallbackFunction( StreamBufferHandle_t pxCallingStreamBuffer );`

......In this case the parameter is already known and doesn't need passing into the function that creates the stream buffer.  This would be consistent with how software timers work.  Software timer callbacks have a parameter that is automatically set to a pointer to the calling timer object.

Alternatively the application writer could set the callback value, which might be an integer (just as an example).  The implementation of the callback function would then look at the value of the integer parameter to determine what to do.  In that case the function that creates the stream buffer would require an additional two parameter.  The first is the pointer to the callback, as now.  The second would be the value of the parameter to pass into the callback.  Then the function that creates the stream before would become:

```
StreamBufferHandle_t xStreamBufferGenericCreate( size_t xBufferSizeBytes,
                                                 size_t xTriggerLevelBytes,
                                                 BaseType_t xIsMessageBuffer,
                                                 StreamBufferCallbackFunction_t pxStreamBufferCallbackFunction,
                                                 void * pvParameter );
```

So there are two additional parameters.  If we then do the same for the receive callback, that will add another two parameters - hence four additional parameters.

If we just pass a pointer to the stream buffer as the parameter then it would be two additional parameters, which is more manageable.

eddie9712:
Thanks for your explanation, I get it, and I think my initial thought is consistent with how the software timers call back work, so I had edited my commit for the following changes:
- Change the parameter of the function(pass the pointer to calling stream buffer), and the parameter will be passed when the sbSEND_COMPLETE is called 
```
....
/* Type used as a stream buffer's optional callback. */
typedef void (* StreamBufferCallbackFunction_t)( StreamBufferHandle_t pxCallingStreamBuffer );
....
#ifndef sbSEND_COMPLETED
    #define sbSEND_COMPLETED( pxStreamBuffer )                                                                           \
    vTaskSuspendAll();                                                                                                   \
    {                                                                                                                    \
        if( ( pxStreamBuffer )->pxStreamBufferSendCallbackFunction != NULL )                                             \
        {                                                                                                                \
            ( pxStreamBuffer )->pxStreamBufferSendCallbackFunction( ( StreamBufferHandle_t )pxStreamBuffer );            \
        }                                                                                                                \
...
``` 
- Do the same thing to the sbRECEIVE_COMPLETE(receive callback)

ravibhagavandas:
Hi @eddie9712 
We have merged your changes with some additions in this PR: https://github.com/FreeRTOS/FreeRTOS-Kernel/pull/437
I will close this PR. Feel free to open if you have questions.

PR #366, opened 2021-07-15
RISC-V Nested Interrupt Support

cmdrf:
RISC-V Nested Interrupt Support

Description
-----------
This adds nested interrupt support to the RISC-V port. As a bonus it adds a `xPortIsInsideInterrupt` macro. Inspired by [Canaan's port](https://github.com/kendryte/kendryte-freertos-sdk/blob/develop/lib/bsp/crt.S).

Test Steps
-----------
Tested on a Kendryte K210 64-bit RISC-V MCU with an interrupt-heavy application that would crash reproducibly without this change.

Disclaimer
----------
I'm not a RISC-V or OS programming expert. I only tested this on a single target. I'm not mad if this PR stays open for a long time until someone can confirm its correctness (or not).


RichardBarry:
Thanks for your contribution.  We currently have three vectored interrupt implementations so I will ask for your patience while we consolidate to a single upstream version.

PR #372, opened 2021-07-28
Move uxCoreAffinityMask after xMPUSettings

N3xed:
Allows for much easier assembly access to the `uxCoreAffinityMask` field. This allows for the port implementation to pin the task to one or more cores much easier and without relying on hardcoded field offset constants.

One such occasion would be in the ESP32 port. The ESP32 port needs to pin the task to a single core when the FPU is used in a multi-core environment. This is because each core has its own FPU where its contents are lazily saved and restored only upon usage. When a task is switched to another core in the middle of a floating-point calculation, the state remains in the other core. To work around this without pinning the core, one would need to either save the FPU state for every task context switch or use some cross-core interrupt magic.


aggarg (approved):

mbruno-xmos:
This should have come with a corresponding adjustment to the xSTATIC_TCB struct

PR #393, opened 2021-09-13
The Interupt need to be disabled over the restore of the FPU and rest…

fireblade156:
…ore Stackaligment

ARM CORTEX A9 Port

Description
-----------
In the ISR should be the Interrupts locked over the Restore FPU and Restore Stackaligment. In other case, the interrupts can interrupts the Restore the FPU register / Restore Stackaligment

Please Integrate this change in the FreeRTOS LTS

### DIFF
--- a/portable/GCC/ARM_CA9/portASM.S
+++ b/portable/GCC/ARM_CA9/portASM.S
@@ -201,12 +201,13 @@ FreeRTOS_IRQ_Handler:
 	PUSH	{r0-r4, lr}
 	LDR		r1, vApplicationIRQHandlerConst
 	BLX		r1
-	POP		{r0-r4, lr}
-	ADD		sp, sp, r2
-
+	
 	CPSID	i
 	DSB
 	ISB
+	
+	POP		{r0-r4, lr}
+	ADD		sp, sp, r2
 
 	/* Write the value read from ICCIAR to ICCEOIR. */
 	LDR 	r4, ulICCEOIRConst
@@ -295,6 +296,10 @@ vApplicationIRQHandler:
 
 	LDR		r1, vApplicationFPUSafeIRQHandlerConst
 	BLX		r1
+	
+	CPSID	i
+	DSB
+	ISB
 
 	POP		{R0}
 	VPOP	{D16-D31}


portable/GCC/ARM_CA9/portASM.S
```diff
@@ -201,12 +201,13 @@ FreeRTOS_IRQ_Handler:
...
 	BLX		r1
-	POP		{r0-r4, lr}
-	ADD		sp, sp, r2
-
+	
 	CPSID	i
 	DSB
 	ISB
+	
+	POP		{r0-r4, lr}
```
RichardBarry, line 209: Grateful if you can explain why this change is needed.

Its been a while since I worked on this port, but as I recall these registers are pushed and popped to retain their values across a function call, rather than to protect them from interrupts, hence they are popped from outside of the critical section (i.e. before interrupts are disabled).  In the current release code version, if an interrupt occurred between the registers being popped and interrupts being disabled then the CPU would switch to IRQ mode, causing it to switch from the SYSTEM mode stack to the IRQ mode stack, and so not clobber these register values.

Maybe you are highlighting a problem in the interrupt nesting code, if so, please provide an explanation of the sequence that could cause a problem.  The registers are not protected from interrupts occurring before the function call, or within the function call, so I'm not sure why they need protecting after the function call.

portable/GCC/ARM_CA9/portASM.S
```diff
@@ -295,6 +296,10 @@ vApplicationIRQHandler:
 
 	LDR		r1, vApplicationFPUSafeIRQHandlerConst
 	BLX		r1
+
```
RichardBarry, line 299: Again grateful if you can provide an explanation of this change.  What if the IRQ handler disables interrupt?

fireblade156:
Hello Richard

The reason is, that at GCC the register R0-R3 are defined als caller 
saved. This mean, this register need to be store be the caller of a 
functions, if it needed.

For excample, the nested Interrupt interrupt after the pop, a function 
modify the the a regiser from R0-R3, call a subroutine, but doesn't need 
the value in the register after the Subroutine call. In this case, the 
register willl not be restore after it enter back in this code.

The same problem exising, in by restore of the FPU register after the call 
of vApplicationFPUSafeIRQHandler.

In the same case, I have a other question, that I havn't validated now. By 
the call of vTaskSwitchContext isn't a stack aligment to 8 bit. For me is 
the here the Question, if this is needed, then the ARM EBI define for 
C-Functions a 8 bit Stack aligment.

Have a nice day
Maurus




Von:    "RichardBarry" ***@***.***>
An:     "FreeRTOS/FreeRTOS-Kernel" ***@***.***>
Kopie:  "fireblade156" ***@***.***>, "Author" 
***@***.***>
Datum:  13.09.2021 23:36
Betreff:        Re: [FreeRTOS/FreeRTOS-Kernel] The Interupt need to be 
disabled over the restore of the FPU and rest… (#393)



@RichardBarry commented on this pull request.

In portable/GCC/ARM_CA9/portASM.S:
>                CPSID           i
                 DSB
                 ISB
+ 
+                POP                             {r0-r4, lr}

Grateful if you can explain why this change is needed.
Its been a while since I worked on this port, but as I recall these 
registers are pushed and popped to retain their values across a function 
call, rather than to protect them from interrupts, hence they are popped 
from outside of the critical section (i.e. before interrupts are 
disabled). In the current release code version, if an interrupt occurred 
between the registers being popped and interrupts being disabled then the 
CPU would switch to IRQ mode, causing it to switch from the SYSTEM mode 
stack to the IRQ mode stack, and so not clobber these register values.
Maybe you are highlighting a problem in the interrupt nesting code, if so, 
please provide an explanation of the sequence that could cause a problem. 
The registers are not protected from interrupts occurring before the 
function call, or within the function call, so I'm not sure why they need 
protecting after the function call.

In portable/GCC/ARM_CA9/portASM.S:
> @@ -295,6 +296,10 @@ vApplicationIRQHandler:
 
                 LDR                             r1, 
vApplicationFPUSafeIRQHandlerConst
                 BLX                             r1
+ 

Again grateful if you can provide an explanation of this change. What if 
the IRQ handler disables interrupt?
—
You are receiving this because you authored the thread.
Reply to this email directly, view it on GitHub, or unsubscribe.
Triage notifications on the go with GitHub Mobile for iOS or Android. 

______________________________________________________________________
This email has been scanned by the Symantec Email Security.cloud service.
For more information please visit http://www.symanteccloud.com
______________________________________________________________________



This message contains information that may be privileged or confidential 
and is the property of the NUM Group. It is intended only for 
the person to whom it is addressed. If you are not the intended recipient, 
you are not authorized to read, print, retain, copy, disseminate, 
distribute, or use this message or any part thereof. If you receive this 
message in error, please notify the sender immediately and delete 
all copies of this message.


______________________________________________________________________
This email has been scanned by the Symantec Email Security.cloud service.
For more information please visit http://www.symanteccloud.com
______________________________________________________________________

RichardBarry:
I'm afraid I'm going t need more detail.  Perhaps it is easiest if I try and describe the execution sequence, then you can point out the part where our understanding is different.  I also don't understand why the r0-r3 values would only need to be protected after existing the application defined IRQ handler, rather than everywhere else in the handler function they get used.


This is my reading of the code:

At [this](https://github.com/FreeRTOS/FreeRTOS-Kernel/blob/V10.4.5/portable/GCC/ARM_CA9/portASM.S#L204) point in the code the processor is in supervisor mode.  If an interrupt occurs the processor will switch back to IRQ mode, and start executing from [here](https://github.com/FreeRTOS/FreeRTOS-Kernel/blob/V10.4.5/portable/GCC/ARM_CA9/portASM.S#L166).  The IRQ entry code switches back to Supervisor mode [here](https://github.com/FreeRTOS/FreeRTOS-Kernel/blob/V10.4.5/portable/GCC/ARM_CA9/portASM.S#L174), before pushing the caller saved registers on the next line.

The application defined handler is called, and once execution returns from the application handler, the original values of the caller saved registers are popped [here](https://github.com/FreeRTOS/FreeRTOS-Kernel/blob/V10.4.5/portable/GCC/ARM_CA9/portASM.S#L234).

The nested interrupt then returns to the original interrupts, with its original r0-r3 register values.

RichardBarry:
Additionally - it would be great to be having this conversation in the support forum so others can contribute, and the conversation is available to everybody in the future.  We do recommend creating a support request before posting bugs.

fireblade156:
I seens, you are rigth and this is correct. In this case, my problems aren't comes from here and I have only hide somethings.

Issue #404, opened 2021-11-01
[BUG] Stack initialisation incorrect on dsPIC.


andrewtjs:
**Describe the bug**
When the deprecated pre-processor constant MPLAB_DSPIC_PORT is undefined, the initialisation of xInitialStack has an incorrect size, resulting in stack corruption.

**Target**
- Instruction Set Architecture: Microchip dsPIC
- IDE and version: MPLAB X v5.50
- Toolchain and version: MPLAB XC16 v1.70

cobusve:
Fixed with PR #405

PR #405, opened 2021-11-01
bugfix: Initialise stack correctly on dsPIC port

andrewtjs:
Description
-----------
Remove the use of MPLAB_DSPIC_PORT for determining correct stack initialisation, use the XC16 compiler device family macros instead.

Related Issue
-----------
[BUG] Stack initialisation incorrect on dsPIC. #404 


cobusve:
This looks like a good change, can we also remove all other instances of this define, I still see it in one more place and also in the comments?

aggarg (approved):

cobusve (approved):

PR #426, opened 2021-12-19
Update the memory alignment within the Cortex-A9 port asm code

RichardBarry:
Description
-----------
Update the assembly code used in the Cortex-A9 port to ensure stack alignment before calling vTaskSwitchContext() in case user code added to the function (maybe via a trace macro or other mechanism) requires the more strict alignment.  That would be the case if, as an example, floating point instructions were used.

Test Steps
-----------
The self monitoring integration tests have been running for several weeks with no issues.


aggarg (approved):

chinglee-iot (approved):
Re-verify the ARM_CA9 port on zynq board without problem.

chinglee-iot (approved):

aggarg (approved):

PR #438, opened 2022-01-07
CMake Support

jonenz:
Description
-----------
Add CMake support for the FreeRTOS Kernel. This allows the user to configure the project and provides some simple checks.

Usage
-----------
Required parameters are:
- `FREERTOS_KERNEL_CONFIG`
- `FREERTOS_KERNEL_PORT` 
- `FREERTOS_KERNEL_HEAP_FILE`

Optional parameters are:
- `FREERTOS_KERNEL_COMPILER_PORT`
- `FREERTOS_KERNEL_USE_CUSTOM_PORT`
- `FREERTOS_KERNEL_USE_CUSTOM_HEAP`

Simple example configuration:
```
    set(FREERTOS_KERNEL_CONFIG ${CMAKE_CURRENT_SOURCE_DIR}/include)
    set(FREERTOS_KERNEL_PORT ARM_CM0)
    set(FREERTOS_KERNEL_HEAP_FILE heap_4)

    add_subdirectory(${CMAKE_CURRENT_SOURCE_DIR}/FreeRTOS-Kernel)
```

Example configuration that specifies the compiler port instead of letting CMake identify it:
```
    set(FREERTOS_KERNEL_CONFIG ${CMAKE_CURRENT_SOURCE_DIR}/include)
    set(FREERTOS_KERNEL_COMPILER_PORT IAR)
    set(FREERTOS_KERNEL_PORT ARM_CM0)
    set(FREERTOS_KERNEL_HEAP_FILE heap_4)

    add_subdirectory(${CMAKE_CURRENT_SOURCE_DIR}/FreeRTOS-Kernel)
```

Example configuration that uses a custom port/heap:
```
    set(FREERTOS_KERNEL_USE_CUSTOM_PORT ON)
    set(FREERTOS_KERNEL_PORT ${CMAKE_CURRENT_SOURCE_DIR}/my/FreeRTOS/port)

    set(FREERTOS_KERNEL_USE_CUSTOM_HEAP OFF)
    set(FREERTOS_KERNEL_HEAP_FILE ${CMAKE_CURRENT_SOURCE_DIR}/my/FreeRTOS/heap/my_heap.c)

    set(FREERTOS_KERNEL_CONFIG ${CMAKE_CURRENT_SOURCE_DIR}/include)

    add_subdirectory(${CMAKE_CURRENT_SOURCE_DIR}/FreeRTOS-Kernel)
```

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 3.16)
+cmake_minimum_required(VERSION 3.12)
 
 set(CMAKE_TRY_COMPILE_TARGET_TYPE "STATIC_LIBRARY")
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -9,11 +9,6 @@ list(APPEND CMAKE_MESSAGE_CONTEXT "FreeRTOS-Kernel")
 add_library(FreeRTOS-Kernel OBJECT)
 
 # Optional parameters.
-option(FREERTOS_KERNEL_ENABLE_COROUTINE "Enable co-routines by compiling croutine.c as part of the kernel source files." ON)
-option(FREERTOS_KERNEL_ENABLE_EVENT_GROUPS "Enable event groups by compiling event_groups.c as part of the kernel source files." ON)
-option(FREERTOS_KERNEL_ENABLE_STREAM_BUFFER "Enable stream and message buffers by compiling stream_buffer.c as part of the kernel source files." ON)
-option(FREERTOS_KERNEL_ENABLE_TIMER "Enable timers by compiling timers.c as part of the kernel source files." ON)
-
 option(FREERTOS_KERNEL_USE_CUSTOM_PORT "Enable the use of a custom port implementation from a different project. When enabling this option set <FREERTOS_KERNEL_PORT> to the path that contains port files." OFF)
 option(FREERTOS_KERNEL_USE_CUSTOM_HEAP "Enable the use of a custom heap implementation from a different project. When enabling this option set <FREERTOS_KERNEL_HEAP_FILE> to the file that implements the memory management." OFF)
 
@@ -128,27 +123,14 @@ target_include_directories(FreeRTOS-Kernel PUBLIC
 ## FreeRTOS Kernel Sources ##
 # Add core kernel source files.
 target_sources(FreeRTOS-Kernel PRIVATE
+    "${CMAKE_CURRENT_SOURCE_DIR}/croutine.c"
+    "${CMAKE_CURRENT_SOURCE_DIR}/event_groups.c"
     "${CMAKE_CURRENT_SOURCE_DIR}/list.c"
     "${CMAKE_CURRENT_SOURCE_DIR}/queue.c"
+    "${CMAKE_CURRENT_SOURCE_DIR}/stream_buffer.c"
     "${CMAKE_CURRENT_SOURCE_DIR}/tasks.c"
-)
-
-# Add optional kernel soruce files.
-if(FREERTOS_ENABLE_COROUTINE)
-    target_sources(FreeRTOS-Kernel PRIVATE "${CMAKE_CURRENT_SOURCE_DIR}/croutine.c" )
-endif(FREERTOS_ENABLE_COROUTINE)
-
-if(FREERTOS_ENABLE_EVENT_GROUPS)
-    target_sources(FreeRTOS-Kernel PRIVATE "${CMAKE_CURRENT_SOURCE_DIR}/event_groups.c" )
-endif(FREERTOS_ENABLE_EVENT_GROUPS)
-
-if(FREERTOS_ENABLE_STREAM_BUFFER)
-    target_sources(FreeRTOS-Kernel PRIVATE "${CMAKE_CURRENT_SOURCE_DIR}/stream_buffer.c")
-endif(FREERTOS_ENABLE_STREAM_BUFFER)
-
-if(FREERTOS_ENABLE_TIMER)
-    target_sources(FreeRTOS-Kernel PRIVATE "${CMAKE_CURRENT_SOURCE_DIR}/timers.c")
-endif(FREERTOS_ENABLE_TIMER)
+    "${CMAKE_CURRENT_SOURCE_DIR}/timers.c"
+    )
 
 # Add heap implemntation source file.
 target_sources(FreeRTOS-Kernel PRIVATE ${FREERTOS_KERNEL_HEAP_FILE})
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,3 +1,44 @@
+#[[
+
+Required parameters are:
+ - FREERTOS_KERNEL_CONFIG
+ - FREERTOS_KERNEL_PORT
+ - FREERTOS_KERNEL_HEAP_FILE
+
+ Optional parameters are:
+ - FREERTOS_KERNEL_COMPILER_PORT
+ - FREERTOS_KERNEL_USE_CUSTOM_PORT
+ - FREERTOS_KERNEL_USE_CUSTOM_HEAP
+
+
+Simple example configuration:
+    set(FREERTOS_KERNEL_CONFIG "${CMAKE_CURRENT_SOURCE_DIR}/include")
+    set(FREERTOS_KERNEL_PORT ARM_CM0)
+    set(FREERTOS_KERNEL_HEAP_FILE heap_4)
+
+    add_subdirectory("${CMAKE_CURRENT_SOURCE_DIR}/FreeRTOS-Kernel")
+
+
+Example configuration that specifies the compiler port instead of letting CMake identify it:
+    set(FREERTOS_KERNEL_CONFIG "${CMAKE_CURRENT_SOURCE_DIR}/include")
+    set(FREERTOS_KERNEL_COMPILER_PORT IAR)
+    set(FREERTOS_KERNEL_PORT ARM_CM0)
+    set(FREERTOS_KERNEL_HEAP_FILE heap_4)
+
+
+Example configuration that uses a custom port/heap:
+    set(FREERTOS_KERNEL_USE_CUSTOM_PORT ON)
+    set(FREERTOS_KERNEL_PORT "${CMAKE_CURRENT_SOURCE_DIR}/my/FreeRTOS/port")
+
+    set(FREERTOS_KERNEL_USE_CUSTOM_HEAP OFF)
+    set(FREERTOS_KERNEL_HEAP_FILE "${CMAKE_CURRENT_SOURCE_DIR}/my/FreeRTOS/heap/my_heap.c")
+
+    set(FREERTOS_KERNEL_CONFIG "${CMAKE_CURRENT_SOURCE_DIR}/include")
+
+    add_subdirectory("${CMAKE_CURRENT_SOURCE_DIR}/FreeRTOS-Kernel")
+
+]]
+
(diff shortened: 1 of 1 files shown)


AniruddhaKanhere (commented):
Hello @jonenz,

Thank you for taking the time to contribute to FreeRTOS. I have one question about the PR, can you please take a look?

Thanks

CMakeLists.txt
```diff
@@ -0,0 +1,160 @@
...
+add_library(FreeRTOS-Kernel OBJECT)
+
+# Optional parameters.
+option(FREERTOS_KERNEL_ENABLE_COROUTINE "Enable co-routines by compiling croutine.c as part of the kernel source files." ON)
+option(FREERTOS_KERNEL_ENABLE_EVENT_GROUPS "Enable event groups by compiling event_groups.c as part of the kernel source files." ON)
+option(FREERTOS_KERNEL_ENABLE_STREAM_BUFFER "Enable stream and message buffers by compiling stream_buffer.c as part of the kernel source files." ON)
+option(FREERTOS_KERNEL_ENABLE_TIMER "Enable timers by compiling timers.c as part of the kernel source files." ON)
+
+option(FREERTOS_KERNEL_USE_CUSTOM_PORT "Enable the use of a custom port implementation from a different project. When enabling this option set <FREERTOS_KERNEL_PORT> to the path that contains port files." OFF)
+option(FREERTOS_KERNEL_USE_CUSTOM_HEAP "Enable the use of a custom heap implementation from a different project. When enabling this option set <FREERTOS_KERNEL_HEAP_FILE> to the file that implements the memory management." OFF)
```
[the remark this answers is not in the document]
jonenz: The intention here wasn't to set the `configUSE_TIMERS` directive, but to give the user the option to omit compiling unnecessary source files.

In practice most projects probably compile these sources whether they use them or not which is why they default to `ON`. If someone were to `set(FREERTOS_KERNEL_ENABLE_TIMER OFF)` and `#define configUSE_TIMERS    (1)` in their FreeRTOSConfig.h they would get a linker error when trying to use timer functions.

These options could be removed if they're more confusing than useful.

CMakeLists.txt
```diff
@@ -0,0 +1,160 @@
...
+    elseif(${CMAKE_C_COMPILER_ID} STREQUAL "IAR")
+        set(FREERTOS_COMPILER_PORT_DIRECTORY "${CMAKE_CURRENT_SOURCE_DIR}/portable/IAR")
+    elseif(${CMAKE_C_COMPILER_ID} STREQUAL "MSVC")
+        set(FREERTOS_COMPILER_PORT_DIRECTORY "${CMAKE_CURRENT_SOURCE_DIR}/portable/MSVC-MingW")
+    elseif(${CMAKE_C_COMPILER_ID} STREQUAL "OpenWatcom")
+        set(FREERTOS_COMPILER_PORT_DIRECTORY "${CMAKE_CURRENT_SOURCE_DIR}/portable/oWatcom/16BitDOS")
+    elseif(${CMAKE_C_COMPILER_ID} STREQUAL "SDCC")
+        set(FREERTOS_COMPILER_PORT_DIRECTORY "${CMAKE_CURRENT_SOURCE_DIR}/portable/SDCC")
+    else()
+        message(FATAL_ERROR "The C compiler identification is ${CMAKE_C_COMPILER_ID}. No port is available for this compiler. Check your toolchain configuration or set <FREERTOS_KERNEL_USE_CUSTOM_PORT> to use a custom FreeRTOS port.")
```
kstribrnAmzn, line 57: Any reason for leaving out the other ports? Or are these to be added when needed?

CMakeLists.txt
```diff
@@ -0,0 +1,160 @@
...
+    endif()
+    message(VERBOSE "The C compiler identification is: ${CMAKE_C_COMPILER_ID}. Compiler port folder is: ${FREERTOS_COMPILER_PORT_DIRECTORY}")
+
+    # Search the port folder for available port configurations.
+    file(GLOB PORTS RELATIVE "${FREERTOS_COMPILER_PORT_DIRECTORY}" CONFIGURE_DEPENDS "${FREERTOS_COMPILER_PORT_DIRECTORY}/*")
```
kstribrnAmzn, line 62: Generally file globbing is discouraged as it makes the [CMake regeneration process a bit trickier](https://cmake.org/cmake/help/v3.16/command/file.html#filesystem).

I realize you did this for brevity, but it may be best to move away from this.

CMakeLists.txt
```diff
@@ -0,0 +1,160 @@
...
+    endif()
+else()
+    # Search the memory management directory for available heap implementations.
+    set(FREERTOS_MEMORY_MANAGEMENT_DIRECTORY "${CMAKE_CURRENT_SOURCE_DIR}/portable/MemMang")
+    file(GLOB HEAP_FILES RELATIVE "${FREERTOS_MEMORY_MANAGEMENT_DIRECTORY}" CONFIGURE_DEPENDS "${FREERTOS_MEMORY_MANAGEMENT_DIRECTORY}/*.c")
+    message(VERBOSE "Available heap files in ${FREERTOS_MEMORY_MANAGEMENT_DIRECTORY} are: ${HEAP_FILES}")
+
+    # Check that <FREERTOS_KERNEL_HEAP_FILE> contains a valid heap file in the memory management directory.
+    if(NOT DEFINED FREERTOS_KERNEL_HEAP_FILE)
+        message(FATAL_ERROR "<FREERTOS_KERNEL_HEAP_FILE> is undefined. Set <FREERTOS_KERNEL_HEAP_FILE> to the desired heap implementation. Available implementations are: ${HEAPS}")
```
kstribrnAmzn, line 106: The _verbose_ level message doesn't really add value, does it? If the heap file can't be found then the user will get both the verbose and error level messages which state the same thing - the available heap files.

I'd consider removing this (and similar messages in this file). Happy case verbose messages are fine as they won't duplicate.

CMakeLists.txt
```diff
@@ -0,0 +1,160 @@
+cmake_minimum_required(VERSION 3.16)
```
kstribrnAmzn, line 1: What's the oldest minimum version we can use here? I am not completely familiar with all the [3.16 changes](https://cmake.org/cmake/help/latest/release/3.16.html) but I'm not seeing any specific features of it that are being used here.

The reason I ask is the older the version of CMake which is supported, the more likely a user will be able to just build the project out of the gate. Less likely to need to setup tools.

CMakeLists.txt
jonenz: My understanding of the general discouragement around file globbing sources is that fails to detect new/deleted files without regeneration. From the link: `If no CMakeLists.txt file changes when a source is added or removed then the generated build system cannot know when to ask CMake to regenerate.`

I don't think that's something to be concerned about here because the files/folders in the portable directory are fairly stable. If a user changes their kernel version to one that added/removed a port, then yes, they would have to regenerate. In my opinion that's reasonable.

CMakeLists.txt
jonenz: I had it in there as a way to list the available heap implementations without needing to run a misconfigured project for them to print. It would only be duplicated when printing verbose messages with an invalid configuration. I'm fine removing it if that's not useful.

CMakeLists.txt
jonenz: I'm not totally sure here, but it's probably 3.12 which added object library support for `target_link_libraries`.

There's probably a separate debate to be had about whether this should be a `STATIC` or `OBJECT` library.

CMakeLists.txt
yhsb2k: `portable/MSVC-MingW` x86 port is not only for MSVC, but for GCC (MinGW64) too.
With your current code, it is not possible to build x86 port with GCC (Windows or Linux) so far.

yhsb2k: And it doesn't handle ports in `portable/ThirdParty/GCC`.

CMakeLists.txt
yhsb2k: `file(GLOB ..)` approach doesn't work for include paths.
For example, we have:
* `portable/GCC/ARM_CM23/non_secure` and `/portable/GCC/ARM_CM23/secure`. Same for ARM_CM33.
* `portable/ThirdParty/GCC/Xtensa_ESP32/include`
* `portable/ThirdParty/Partner-Supported-Ports/GCC/AVR_Mega0`
* `portable/ThirdParty/XCC/Xtensa`

All of paths above are not present in cmake freertos module include path, which is not good for user.

CMakeLists.txt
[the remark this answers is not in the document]
AniruddhaKanhere: I see, I personally did not like the idea that I'd need to make a change at 2 locations to add/remove any option to my build. @kstribrnAmzn - your thoughts?
I would like it better if it is at one location only - creating less confusion.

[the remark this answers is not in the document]
kstribrnAmzn: > I see, I personally did not like the idea that I'd need to make a change at 2 locations to add/remove any option to my build.

I agree, does seem strange.

CMakeLists.txt
kstribrnAmzn: >I tired to match up all the ports for compilers identified by CMake listed here

Thanks for sharing your justification. Totally makes sense. I'd probably do the same thing if I were in your shoes.
I do think the ports your PR handles could be expanded without significant trouble so I won't hold up the PR for it.

jonenz: > `portable/MSVC-MingW` x86 port is not only for MSVC, but for GCC (MinGW-w64) too.
This is a good point and something I didn't think about. It wouldn't be impossible because the user could manually provide the port path, but it would be misleading and something they shouldn't expect to have to do for a natively supported port.

I'll think about how this could be improved.

CMakeLists.txt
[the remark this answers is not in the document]
jonenz: > I would like it better if it is at one location only - creating less confusion.

Since `configUSE_TIMERS`, `configUSE_CO_ROUTINES`, etc define out the entirety of the source file there's no need for these options. They can be removed.

jonenz:
I made some updates that should allow the rest of the ports to be used.

Any folder in the `portable` directory that contains at least one of the port macro header files is considered a compiler port. The user can define which to use by setting `FREERTOS_KERNEL_COMPILER_PORT` or leave it undefined and the script will try and identify one by using `CMAKE_C_COMPILER_ID`. This variable will mostly be useful if CMake detects the incorrect compiler port, or if CMake is not able to identify the compiler being used.

Any folder in the compiler port directory that contains at least one of the port macro header files is considered a port. This improvement allows the user to specify a nested port. For example, `set(FREERTOS_KERNEL_PORT ARM_CM0)` and `set(FREERTOS_KERNEL_PORT ARM_CM33/non_secure)` are both valid.

AniruddhaKanhere:
Hello @jonenz,

Thank you for this PR. It was a welcome contribution. However, we decided to go with https://github.com/FreeRTOS/FreeRTOS-Kernel/pull/421 (the is why the merge conflict) because it supported all ports and did not use file globbing.

I shall close this PR now. If you'd like to contribute further in the domain of CMake, take a look at [this comment](https://github.com/FreeRTOS/FreeRTOS-Kernel/pull/421#issuecomment-1017896015) by @paulbartell. It outlines what we intend to do as 'next-steps' for cmake.

Thanks again for contributing to FreeRTOS.

Thanks and Regards,
Aniruddha

AniruddhaKanhere:
Also, I would thank @jonenz since some suggestions that I made to this PR, were taking from his PR https://github.com/FreeRTOS/FreeRTOS-Kernel/pull/438.

PR #445, opened 2022-01-25
Update uncrustify configuration and improve CI setup

swaldhoer:
Description
-----------

**Uncrustify Setup**

There are two issues with the current uncrustify setup:
- configuration file is based on version 0.67, but another version is used
  - Therefore all options than have been added to uncrustify since them are implict, and this might/will lead to undesried behavior.
  - Therefore, configuration file has been updated to the currently used version. The diff looks like some settings have gone missing, but they are not; these settings are now aligned with the uncrustify configuration and therefore future updates will be simpler.
- the CI step just installs the latest version and therefore does not specifing the exact version:
  - This can lead to CI fails in the future, once a newer version is available, that introduces breaking changes.
  - Therefore, the version is now pinned to overcome this problem.

**Linted Files**

- The portable files ``portable/Common/mcu_wrappers.c`` and ``portable/MemMang/*`` have not been autoformatted.
  - These files have been added to the CI check, to enforce a common code style to the common files (i.e., not third party/vendor specfic), since consistency is in most times a very good thing.

**Why is this a good change?**

- it impoves the usage of a tool, that is already used in the project, which (hopefully) will lead to smaller diffs in the future.
- it leads to a **common style in the common FreeRTOS code base**.

*Note*: For future work, the configuration is most easily updated by running

    uncrustify -c .github/uncrustify.cfg -o .github/uncrustify.cfg --update-config-with-doc

and then stripping the superfluous comments.

**Important:** The [github.com/FreeRTOS/FreeRTOS](https://github.com/FreeRTOS/FreeRTOS) repo does also not pin the uncrustify version (see current [main](https://github.com/FreeRTOS/FreeRTOS/blob/0744c28749efbc0a0f08424eff5f6433078a25a2/.github/workflows/ci.yml#L37)). Depending whether you like this change, I will open a PR in that repo too, so that this repo also pins the uncrustify version (which would be good I think). It might be good to exclude Kernel files from linting in that repo; because otherwise you could create failing CIs, if different uncrustify versions create different formatting, e.g., here it uses version 0.67 and in the consuming repo vesion 0.69. But once you pinned in both repos the uncrustify version, you can step by step update, have always green CIs, even if you do not exclude Kernel file sin the othe repo (I would still preferr to skip them in the other repo).
I hope, I my explanation is understandable, but if not, I am happy to elaborate further if it helps to clarify and getting this PR to be accepted.

Test Steps
-----------

Tested is done in CI.

Related Issue
-----------


alfred2g:
Thanks for your PR!
Tested the config file manually on tasks.c, and I see no changes in the file (a good thing)

alfred2g (approved):

cobusve (approved):

PR #456, opened 2022-02-16
SMP: Set Idle task affinity on creation

Dazza0:
Description
---------------
Idle tasks created in `prvCreateIdleTasks()` should be pinned to their target core on creation. Not doing can result in `prvMinimalIdleTask()` running on core 0 and `prvIdleTask()` running on cores 1 to N-1.

Test Steps
--------------
In `vApplicationIdleHook()`, assert that the current core ID is always 0

Related Issue
------------------

### DIFF
--- a/tasks.c
+++ b/tasks.c
@@ -2744,6 +2744,9 @@ static BaseType_t prvCreateIdleTasks( void )
                 #endif
             }
         #endif /* configSUPPORT_STATIC_ALLOCATION */
+
+        /* Pin the created idle task to its target core. */
+        vTaskCoreAffinitySet( xIdleTaskHandle[ xCoreID ], ( 1 << xCoreID ) );
     }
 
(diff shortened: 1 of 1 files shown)


aggarg (approved):

Dazza0:
@n9wxu Hi sorry for the late reply. The pinning of the IDLE tasks on the ESP32 were done primarily for the following reasons.

- Due to FPU lazy context saving, any tasks that uses the FPU will:
    - Automatically get pinned to a core on creation
    - Must be deleted on the same core it was pinned on (i.e., the IDLE task needs to go and clear the registers)
- Feeding of the task watchdog timer (the Idle task must be able to run on each core occasionally to guarantee that no particular CPU is 100% busy)

I'll check again if some other ways to work around the IDLE tasks not being pinned on the ESP32.

PR #470, opened 2022-03-09
SMP: Add task creation with affinity functions

Dazza0:
Description
-----------

We have some instances when a task function must be guaranteed to only ever run on a particular core such as:

- When the task function accesses some shared array indexed by  `xCoreID`
- When the task function accesses some per-cpu specific resources (such as coprocessors) 

The current task creation API expects users to set a task's affinity (using `vTaskCoreAffinitySet()`) after the task is created. For example, if Task A creates Task B using `xTaskCreate()`, Task A should then set the core affinity of Task B using `vTaskCoreAffinitySet()`.

However, as soon as Task B is created, it is possible for Task B to be scheduled on any core (as created tasks have no core affinity by default). Thus, extra logic is required in both Task A and Task B in order to guarantee that bulk of Task B only ever runs on its intended core. For example:

```c
void TaskB(void *arg)
{
    ulTaskNotifyTake(pdTRUE, portMAX_DELAY);  //Wait to be started by Task A after affinity is set
    ...
}

void TaskA(void *arg)
{
    TaskHandle_t hdl_B;
    xTaskCreate(..., &hdl_B); //Create Task B
    vTaskCoreAffinitySet(hdl_B, 1 << 0);  //Pin Task B to core 0
    xTaskNotifyGive(hdl_B); //Start TaskB
}

```

This PR adds the convenience function listed below, which allow a task's core affinity to be specified on creation. This will guarantee that the task will only ever run on it's intended cores, and saves us from needing the extra logic demonstrated in the example above.

- `xTaskCreateAffinitySet()`
- `xTaskCreateStaticAffinitySet()`
- `xTaskCreateRestrictedAffinitySet()`
- `xTaskCreateRestrictedStaticAffinitySet()`

Test Steps
-----------

Related Issue
-----------

### DIFF
--- a/include/task.h
+++ b/include/task.h
@@ -380,6 +380,16 @@ typedef enum
                             TaskHandle_t * const pxCreatedTask ) PRIVILEGED_FUNCTION;
 #endif
 
+#if ( ( configSUPPORT_DYNAMIC_ALLOCATION == 1 ) && ( configNUM_CORES > 1 ) && ( configUSE_CORE_AFFINITY == 1 ) )
+    BaseType_t xTaskCreateAffinitySet( TaskFunction_t pxTaskCode,
+                                       const char * const pcName,     /*lint !e971 Unqualified char types are allowed for strings and single characters only. */
+                                       const configSTACK_DEPTH_TYPE usStackDepth,
+                                       void * const pvParameters,
+                                       UBaseType_t uxPriority,
+                                       UBaseType_t uxCoreAffinityMask,
+                                       TaskHandle_t * const pxCreatedTask ) PRIVILEGED_FUNCTION;
+#endif
+
 /**
  * task. h
  * <pre>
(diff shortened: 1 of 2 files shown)


aggarg (approved):

n9wxu (approved):

snikeguo:
prvCreateIdleTasks->xTaskCreate->xTaskCreateAffinitySet->prvInitialiseNewTask->pxPortInitialiseStack

in tricore :
pxPortInitialiseStack()
{
  var csa=GetCurrentCpuFreeCsaArea();
  init csa; 
   init sp;
  int .....
}
For the TRICORE architecture, each core can only access its own CSA and cannot access the CSAs of other cores, so xTaskCreate must be executed on the corresponding core!

dongsheng-autra:
> prvCreateIdleTasks->xTaskCreate->xTaskCreateAffinitySet->prvInitialiseNewTask->pxPortInitialiseStack
> 
> in tricore : pxPortInitialiseStack() { var csa=GetCurrentCpuFreeCsaArea(); init csa; init sp; int ..... } For the TRICORE architecture, each core can only access its own CSA and cannot access the CSAs of other cores, so xTaskCreate must be executed on the corresponding core!

Hi @snikeguo ,

I am working on TC397, and trying multi-core support, do you have some code to share?

PR #489, opened 2022-04-20
Format code, and remove implicit tests

alfred2g:
Format code, and remove implicit tests

Description
-----------
Run uncrustify on the code
Remove implicit tests

Test Steps
-----------
 Demo build and run

### DIFF
--- a/portable/ThirdParty/GCC/Posix/port.c
+++ b/portable/ThirdParty/GCC/Posix/port.c
@@ -27,28 +27,29 @@
  */
 
 /*-----------------------------------------------------------
- * Implementation of functions defined in portable.h for the Posix port.
- *
- * Each task has a pthread which eases use of standard debuggers
- * (allowing backtraces of tasks etc). Threads for tasks that are not
- * running are blocked in sigwait().
- *
- * Task switch is done by resuming the thread for the next task by
- * signaling the condition variable and then waiting on a condition variable
- * with the current thread.
- *
- * The timer interrupt uses SIGALRM and care is taken to ensure that
- * the signal handler runs only on the thread for the current task.
- *
- * Use of part of the standard C library requires care as some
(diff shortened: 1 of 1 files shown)


portable/ThirdParty/GCC/Posix/port.c
```diff
@@ -338,12 +344,13 @@ static uint64_t prvStartTimeNs;
...
-struct itimerval itimer;
-int iRet;
+    struct itimerval itimer;
+    int iRet;
 
     /* Initialise the structure with the current timer information. */
     iRet = getitimer( ITIMER_REAL, &itimer );
-    if ( iRet )
+
+    if( iRet == -1 )
```
kstribrnAmzn, line 353: This is a difference condition from the one you're replacing. Before it was checking that the value is non-zero. Now it's checking that it's -1.

Questions:
1. Are any values other than -1 used to represent an error?
2. Could a positive value be returned?

alfred2g: if (iRet) would be true on -1, only a zero would be considered false
1.  Only -1 on error, the error code is returned in errno
2.  No positive values in this one, unlike pthread_create which returns a positive error code

kstribrnAmzn: >if (iRet) would be true on -1, only a zero would be considered false
Agreed...What I'm getting at is before `if (iRet)` would be true for _any_ non-zero number, and false for 0. What we have now is for it to be true iff `iRet` is -1. This is a very different conditional. We'd now be excluding all positive numbers from this case and all negative numbers other than one.

With that said...since the errno can only be -1 it's all good. I'll approve.

aggarg (approved):

kstribrnAmzn (approved):

PR #497, opened 2022-05-30
RP2040: Allow FreeRTOS to be added to the parent CMake project post initialization of the Pico SDK

kilograham:
Including FreeRTOS in the top level CMakeLists.txt is not always convenient, so this change allows
FreeRTOS_Kernel_import.cmake or just the CMakeLists.txt to be included wherever in the user's project.


kilograham:
Actually this does not correctly inject the headers; so I will work on this some more

kilograham:
Note, this is now ready to go. Would be good to get it merged as it is needed for an upcoming release of our software.

aggarg (dismissed):

alfred2g (approved):

paulbartell (approved):

PR #501, opened 2022-06-08
Fix RP2040 assertion due to yield spin lock info being wrongly shared between multiple cores

kilograham:
Some state relating to atomically unlocking of a Raspberry Pi Pico SDK spin lock when suspending a FreeRTOS task was wrongly shared between cores causing undefined behavior if both cores happen to release spin locks at exactly the same time. This fix creates per core copies of the state.


n9wxu (approved):

AniruddhaKanhere (approved):

AniruddhaKanhere:
Thanks @kilograham for finding this bug and raising a PR for it.
We have reviewed and approved the PR. 

I will now merge it.

PR #516, opened 2022-06-27
SMP: Change call order of vApplicationMinimalIdleHook()

Dazza0:
Change order of `vApplicationMinimalIdleHook()` call
-----------

When `configUSE_IDLE_HOOK`, `configUSE_MINIMAL_IDLE_HOOK`, and `configUSE_TICKLESS_IDLE` are all enabled, and enough ticks have passed to enter a tickless idle, then tickless idle will be entered before calling `vApplicationMinimalIdleHook()`.

Thus, when entering tickless idle, `vApplicationMinimalIdleHook()` will be invoked one time less than compared to `vApplicationIdleHook()`.

This PR moves the call of `vApplicationMinimalIdleHook()` within `prvIdleTask()` to before the handling of tickless idle. Thus, the number of times that `vApplicationIdleHook()` and `vApplicationMinimalIdleHook()` are called will be the same before entering tickless idle.


Dazza0:
Note: On second thought though, if each core is expected to call a WFI instruction from `vApplicationMinimalIdleHook()`, then it might make sense for `vApplicationMinimalIdleHook()` to be called after tickless idle (since `portSUPPRESS_TICKS_AND_SLEEP()` can set a bit within the port to skip the following WFI after waking up from tickless idle).

n9wxu:
Do you want to close this PR based upon your "second thought"?

PR #520, opened 2022-07-01
Add support for MISRA rule 8.13

alfred2g:
Add support for MISRA rule 8.13

Description
-----------
Add support for MISRA rule 8.13
MISRA rule  8.13, likes non changeable memory to be pointed by constant pointers

### DIFF
--- a/include/task.h
+++ b/include/task.h
@@ -84,8 +84,8 @@
  * \ingroup Tasks
  */
 struct tskTaskControlBlock; /* The old naming convention is used to prevent breaking kernel aware debuggers. */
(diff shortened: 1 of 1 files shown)


RichardBarry:
What is the purpose of this change?  It appears to add a new type that isn't used by the kernel anywhere so I'm assuming it's intended for use by the application.  Can the application writer provide the const version as part of the application code, rather than kernel code?

alfred2g:
yeah true, it is for the applications, in this case for the TCP stack to use that typedef, yeah the application can define it locally.
i'll probably do that, and close this PR.

PR #532, opened 2022-07-30
Block `SIG_SUSPEND` in the main thread of the POSIX port so that `sigwait` works as expected

chrisnc:
Description
-----------
The POSIX standard[^1] requires that `sigwait` should only be used with blocked signals, otherwise the behavior is undefined. When the POSIX port is used on macOS for example, the `sigwait` in `xPortStartScheduler` never returns because there is a handler set up to ignore the `SIG_SUSPEND` signal. This change removes the ignore handler and blocks the signal prior to `sigwait`ing for it in the main thread.

Test Steps
-----------
Build a basic example on a non-Linux POSIX system using the POSIX FreeRTOS port, and try to use `vTaskEndScheduler`.
Without these changes, the program does not exit on macOS. It continues to behave as expected on Linux.

```c
#include <FreeRTOS.h>
#include <task.h>

#include <limits.h>

static void fn(void *arg)
{
    vTaskEndScheduler();
}

int main(void)
{
    static StackType_t stack[PTHREAD_STACK_MIN];
    static StaticTask_t task;
    xTaskCreateStatic(fn, "task", PTHREAD_STACK_MIN, NULL, 1, stack, &task);
    vTaskStartScheduler();
}
```

[^1]: https://pubs.opengroup.org/onlinepubs/009604499/functions/sigwait.html


chrisnc:
Fixed commit message which mistakenly referred to `SIG_SUSPEND`...

chrisnc:
Branch is rebased to latest main now.

dachalco:
Hey @chrisnc 

Thank you for the fix. @alfred2g can you review?

alfred2g:
@chrisnc, I see that you are still pushing some commits, could you please let me know when you are done, so I test it and review ? 
Thanks

chrisnc:
I'm done after fixing the commit message a few days ago, the others have just been to rebase as `main` advances, since I noticed you did so with a merge before. I'll do this again now, but the change is ready for review.

alfred2g:
@chrisnc Thanks!, 
I will run some demos and review

alfred2g (approved):

archigup (approved):

PR #543, opened 2022-08-11
RISC-V: Add support for RV32E extension in GCC port

Limoto:
Add support for RV32E extension in RISC-V GCC port

Description
-----------
RV32E (E for embedded) processors only have 16 registers instead of the 32 a regular RV32 core has. This change allows running FreeRTOS on such cores, by not touching registers `x16`-`x31` when FreeRTOS is built for RV32E.

Test Steps
-----------
Compile the code for RV32E core with EABI. (e.g. `gcc -march=rv32e -mabi=ilp32e` ). The code can be then run also on a regular RV32 core.

### DIFF
--- a/portable/GCC/RISC-V/portASM.S
+++ b/portable/GCC/RISC-V/portASM.S
@@ -98,35 +98,35 @@ definitions. */
 /*-----------------------------------------------------------*/
 
 .macro portUPDATE_MTIMER_COMPARE_REGISTER
-    load_x t0, pullMachineTimerCompareRegister  /* Load address of compare register into t0. */
-    load_x t1, pullNextTime                     /* Load the address of ullNextTime into t1. */
+    load_x a0, pullMachineTimerCompareRegister  /* Load address of compare register into a0. */
+    load_x a1, pullNextTime                     /* Load the address of ullNextTime into a1. */
 
     #if( __riscv_xlen == 32 )
 
         /* Update the 64-bit mtimer compare match value in two 32-bit writes. */
-        li t4, -1
-        lw t2, 0(t1)                /* Load the low word of ullNextTime into t2. */
-        lw t3, 4(t1)                /* Load the high word of ullNextTime into t3. */
-        sw t4, 0(t0)                /* Low word no smaller than old value to start with - will be overwritten below. */
-        sw t3, 4(t0)                /* Store high word of ullNextTime into compare register.  No smaller than new value. */
-        sw t2, 0(t0)                /* Store low word of ullNextTime into compare register. */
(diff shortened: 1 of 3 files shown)


portable/GCC/RISC-V/portASM.S
```diff
@@ -206,7 +206,12 @@ pxPortInitialiseStack:
     store_x t0, 0(a0)                   /* mstatus onto the stack. */
     addi a0, a0, -portWORD_SIZE         /* Space for critical nesting count. */
     store_x x0, 0(a0)                   /* Critical nesting count starts at 0 for every task. */
+
+#ifdef __riscv_32e
```
RichardBarry, line 210: Looking at this page: https://github.com/riscv-non-isa/riscv-c-api-doc/blob/master/riscv-c-api.md - would using __riscv_e provide better future and wider compatibility than __riscv_32e?

Limoto: Regarding GCC, `__riscv_32e` was added by commit `09baee1ab152` on 2018-05-18, `__riscv_e` by commit `e3354b6de7a3` on 2020-12-11. Both are present at the moment.

Regarding clang, there is no support for RV32E in upstream yet, but the proposed patch adds both of them: https://reviews.llvm.org/D70401

There is nothing about deprecation of `__riscv_32e` in the document. So I would say `__riscv_32e` is currently wider supported.

portable/GCC/RISC-V/portmacro.h
```diff
@@ -80,7 +80,11 @@ typedef portUBASE_TYPE TickType_t;
 /* Architecture specifics. */
 #define portSTACK_GROWTH            ( -1 )
 #define portTICK_PERIOD_MS          ( ( TickType_t ) 1000 / configTICK_RATE_HZ )
-#define portBYTE_ALIGNMENT          16
+#ifdef __riscv_32e
+    #define portBYTE_ALIGNMENT      8   /* RV32E uses RISC-V EABI with reduced stack alignment requirements */
```
RichardBarry, line 84: I have no doubt the reduced alignment requirement is correct, but for my own records would like to know where it is documented.  Grateful for a link.

Limoto: It's documented here, section 4: https://github.com/riscv-non-isa/riscv-eabi-spec/blob/master/EABI.adoc#4-eabi-stack-alignment

RichardBarry:
I'm seeing x14 get corrupted - still looking into it.

n9wxu (approved):

dachalco (approved):

PR #549, opened 2022-08-22
Fix type for pointers operations

Octaviarius:
This PR fixes type of pointer operations
-----------

While I working with RL78 MCU I caught a problem: heap allocation was putting into fault my device. After investigation the problem I found sizeof(size_t) != sizeof(void*). It can be fixed with replacement size_t in pointer arithmetic to uintptr_t via FreeRTOS.h file.
-----------

Just use heap allocation on RL78 MCU
-----------

### DIFF
--- a/include/FreeRTOS.h
+++ b/include/FreeRTOS.h
@@ -390,7 +390,7 @@
 #endif
 
 #ifndef portPOINTER_SIZE_TYPE
-    #define portPOINTER_SIZE_TYPE    uint32_t
+    #define portPOINTER_SIZE_TYPE    uintptr_t
 #endif
 
 /* Remove any unused trace macros. */
--- a/include/FreeRTOS.h
+++ b/include/FreeRTOS.h
@@ -390,7 +390,7 @@
 #endif
 
 #ifndef portPOINTER_SIZE_TYPE
-    #define portPOINTER_SIZE_TYPE    uintptr_t
+    #define portPOINTER_SIZE_TYPE    uint32_t
 #endif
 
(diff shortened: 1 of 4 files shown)


Octaviarius:
Sorry, I rollback changes in FreeRTOS.h because uintptr_t isn't compatible with C89

aggarg:
There are a lot of white space changes and splitting of variable declarations on multiple lines. Are those related too? If not, can we remove those please to make the PR easy to review?

Octaviarius:
> There are a lot of white space changes and splitting of variable declarations on multiple lines. Are those related too? If not, can we remove those please to make the PR easy to review?

Octaviarius:
I made aligning in some lines and split variable declarations on separated lines for making code more readable. If it's not much complicated, could it stay as is?

aggarg:
We have auto formatter which is anyway going to override the spacing changes. Regarding the splitting changes, would you please move them to a separate PR as those are unrelated?

aggarg:
Let me know if you'd like me to make those changes for you.

RichardBarry:
For consistency across all files in the library, and all the other libraries, my preference is to keep the variable declarations as they were, so not split across lines,

Octaviarius:
Ok) Let's do something with formatting, but there in general must stay type
changing (size_t -> portPOINTER_SIZE_TYPE)

Octaviarius:
I've opened new PR without indentations. https://github.com/FreeRTOS/FreeRTOS-Kernel/pull/550

PR #550, opened 2022-08-23
[Fix] Type for pointers operations

Octaviarius:
This PR fixes type of pointer operations

Description
-----------
While I was working with RL78 MCU I caught a problem: heap allocation was putting into fault my device. After investigation the problem I found sizeof(size_t) != sizeof(void*). It can be fixed with replacement size_t in pointer arithmetic to portPOINTER_SIZE_TYPE.

Test Steps
-----------
Just use heap allocation on RL78 MCU

### DIFF
--- a/portable/MemMang/heap_5.c
+++ b/portable/MemMang/heap_5.c
@@ -445,7 +445,7 @@ void vPortDefineHeapRegions( const HeapRegion_t * const pxHeapRegions )
     portPOINTER_SIZE_TYPE xAlignedHeap;
     size_t xTotalRegionSize, xTotalHeapSize = 0;
     BaseType_t xDefinedRegions = 0;
-    size_t xAddress;
+    portPOINTER_SIZE_TYPE xAddress;
     const HeapRegion_t * pxHeapRegion;
 
     /* Can only call once! */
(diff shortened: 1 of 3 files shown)


AniruddhaKanhere (approved):

portable/MemMang/heap_5.c
```diff
@@ -442,7 +442,7 @@ static void prvInsertBlockIntoFreeList( BlockLink_t * pxBlockToInsert )
 void vPortDefineHeapRegions( const HeapRegion_t * const pxHeapRegions )
 {
     BlockLink_t * pxFirstFreeBlockInRegion = NULL, * pxPreviousFreeBlock;
-    size_t xAlignedHeap;
+    portPOINTER_SIZE_TYPE xAlignedHeap;
     size_t xTotalRegionSize, xTotalHeapSize = 0;
     BaseType_t xDefinedRegions = 0;
     size_t xAddress;
```
archigup, line 448: Should this be portPOINTER_SIZE_TYPE as well?

Octaviarius: If it must be  compatible with C89  then it's true.  Another way - use uintptr_t, but it since  C99 only

Octaviarius: I fixed it

aggarg (approved):

AniruddhaKanhere (approved):

PR #561, opened 2022-09-15
Arraylist task queue

akshayutture:
Arraylist task queue

Description
-----------
Gives an array-back implementation of lists. The new implementation does not modify the list interface. 

Test Steps
-----------
Runs with CBMC proofs. Have not tried to compile the kernel.


RichardBarry:
Is this intended for use with CBMC only, or for inclusion in release versions of the kernel?

qinheping:
> Is this intended for use with CBMC only, or for inclusion in release versions of the kernel?

It is not CBMC-use only. This PR include an array-based implementation of lists with [unbounded memory safety proof](https://github.com/FreeRTOS/FreeRTOS/pull/848). With this PR along with the proofs, we want to make the array-based implementation more visible, and to help people to see the cost and the benefit of moving from linked-list to array-list. We also want to see if there is any compatibility issue of the array-based implementation.

qinheping:
DO NOT MERGE. It is a proposal for array-based implementation.

alfred2g:
Closing this PR after discussion with the PR owners.

PR #578, opened 2022-10-18
Update the NIOSII port to enable longer jumps

RichardBarry:
Description
-----------
The NIOS II port contained a "call" instruction with limited jump range.  That can cause issues depending on the amount of RAM available and linking.  Full description here: https://forums.freertos.org/t/nios-ii-r-nios2-call26-noat-linker-error/16028

Test Steps
-----------
Requires a configuration where the offending call needs to jump more than 256Mbytes.

Related Issue
-----------
https://forums.freertos.org/t/nios-ii-r-nios2-call26-noat-linker-error/16028

### DIFF
--- a/portable/GCC/NiosII/port_asm.S
+++ b/portable/GCC/NiosII/port_asm.S
@@ -27,12 +27,12 @@
  */
 
 .extern		vTaskSwitchContext
-	
+
 .set noat
 
 # Exported to start the first task.
-.globl restore_sp_from_pxCurrentTCB		
-	
+.globl restore_sp_from_pxCurrentTCB
+
 # Entry point for exceptions.
 .section .exceptions.entry.user, "xa"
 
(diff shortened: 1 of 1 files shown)


portable/GCC/NiosII/port_asm.S
```diff
@@ -131,8 +131,9 @@ soft_exceptions:
 
 call_scheduler:
 	stw		ea, 72(sp)						# EA is PC+4 so will skip over instruction causing exception
-	call	vTaskSwitchContext				# Pick the next context.
-	br		restore_sp_from_pxCurrentTCB	# Switch in the task context and restore. 
+	movia	r15, vTaskSwitchContext			# Pick the next context - use long call version in place of "call"
```
AniruddhaKanhere, line 134: A novice question: why do we not need to save r15 here?
Would that not be overwritten?

Or is it that this will only be called when there is an exception and at that point, we would not care about overwriting the value in r15?

RichardBarry: Good question.  This is calling vTaskSwitchContext(), which is a badly named function, for historic reasons. 
 It is called after the context has been saved, and selects the next task, before restoring the context of the newly selected task (so vTaskSelectNextTask() would be a more accurate function name).  That means r15 has most likely been saved already, and will get restored to be correct for the next task.  However, we should verify that by inspecting the surrounding code and checking the compiler and chip manuals to ensure r15 is not used for a special purpose in the C code within vTaskSwitchContext().

MigraineMan33: On entry to save_context: at the top of port_asm.S, r15 is saved and then immediately munged to grab the estatus value. I figured it would be a safe register to re-munge later.  r15 is used in call_scheduler: to create the 32-bit target address for the callr instruction.  Its value on return from vTaskSwitchContext() doesn't matter, because the following "br  restore_sp_from_pxCurrentTCB" jumps up to the section that pops all the register values off the currently selected context stack.

aggarg (approved):

portable/GCC/NiosII/port_asm.S
AniruddhaKanhere: Thanks for the reply @RichardBarry and @MigraineMan33.
This solves my doubts!

AniruddhaKanhere (approved):

PR #580, opened 2022-10-21
Fix array-bounds compiler warning on gcc11+ in list.h

archigup:
Fix array-bounds compiler warning on gcc11+ in list.h

Description
-----------

listGET_OWNER_OF_NEXT_ENTRY computes `( pxConstList )->pxIndex->pxNext` after verifying that `( pxConstList )->pxIndex` points to `xListEnd`, which due to being a MiniListItem_t, can be shorter than a ListItem_t. Thus, `( pxConstList )->pxIndex` is a `ListItem_t *` that extends past the end of the `List_t` whose `xListEnd` it points to. 

On GCC 11+, this triggers an array-bounds warning. This is fixed by accessing `pxNext` through a `MiniListItem_t` instead.

<del>

listGET_OWNER_OF_NEXT_ENTRY uses uninitialized memory if the passed list is empty (only has the list end marker). While all uses of this macro are nested within a list empty check or are after an assert, this trips up GCC post GCC 11 on O3. Worse yet, with minilist enabled, this macro will dereference a struct offset past the end of the struct when passed an empty list.

Adding this check makes the behavior of the macro on empty list explicit when the only element is xListEnd, and prevents accessing xListEnd->pvOwner (Which is currently NULL where the List_t is zero-initialized; I have not checked if there are cases where it is not).

GCC gives a warning as it catches that the macro does not work with an empty list, but it does not figure that that code path is unreachable. Adding this check or adding more list non-empty checks/asserts directly outside of the macro fixes the issue. With asserts, the issue is only solved when asserts are enabled, so just adding the check is the better solution.

</del>

Test Steps
-----------
Compile tasks.h with GCC 12 (or 11) and with 03 as follows:
```sh
gcc -c tasks.c -I include -I ${KERNEL_CONFIG_PATH}/ -I portable/ThirdParty/GCC/Posix/ -Wall -Wextra -Werror -O3
```

Related Issue
-----------
https://github.com/FreeRTOS/FreeRTOS/issues/858

### DIFF
--- a/include/list.h
+++ b/include/list.h
@@ -290,7 +290,7 @@ typedef struct xLIST
         ( pxConstList )->pxIndex = ( pxConstList )->pxIndex->pxNext;                           \
         if( ( void * ) ( pxConstList )->pxIndex == ( void * ) &( ( pxConstList )->xListEnd ) ) \
         {                                                                                      \
-            ( pxConstList )->pxIndex = ( pxConstList )->pxIndex->pxNext;                       \
+            ( pxConstList )->pxIndex = ( pxConstList )->xListEnd.pxNext;                       \
         }                                                                                      \
         ( pxTCB ) = ( pxConstList )->pxIndex->pvOwner;                                         \
     }


RichardBarry:
The update adds an additional test into listGET_OWNER_OF_NEXT_ENTRY(), which is called from within taskSELECT_HIGHEST_PRIORITY_TASK(), so gets called at high frequency, including potentially in the tick interrupt.  Anything that extends the runtime of high frequency functions needs additional scrutiny, hence I would like to understand this better before merging.

I see the code will assert if the subject list is empty – and I don’t think I’ve seen that assert get hit in cases other than where portUSE_OPTIMISED_TASK_SELECTION is 0 and the lists are already corrupt.  So, if my understanding is correct, the update duplicates code.  On first look, it seems the update is only required to work around a compiler bug (the hardest bugs to find!), in which case we need to report the bug if it is not already being tracked?  Or is the compiler’s behaviour legitimate?  The update impacts all compilers, so if it is a compiler bug work around, will we take it out when the compiler is fixed?

cobusve:
Have we confirmed if this is actually a bug in the compiler? Because if it is not that changes things a lot.

RichardBarry:
As archigup says, without knowing the execution path, the compiler warning is probably legitimate at as a static check (at any optimisation level).  I would not expect there to be an actual error condition at run-time though.

archigup:
I am currently minimizing the testcase to evaluate whether it is a compiler bug.

archigup:
I've minimized the test case to this:
```c
typedef struct xLIST_ITEM
{
    struct xLIST_ITEM * pxNext;
    struct xLIST * pvContainer;
} ListItem_t;

typedef struct xMINI_LIST_ITEM
{
    struct xLIST_ITEM * pxNext;
} MiniListItem_t;

typedef struct xLIST
{
    int uxNumberOfItems;
    ListItem_t * pxIndex;
    MiniListItem_t xListEnd;
} List_t;

static List_t xSuspendedTaskList;

static void prvSearchForNameWithinSingleList( List_t * pxList )
{
    if( ( ( pxList )->uxNumberOfItems ) > 0 )
    {
        pxList->pxIndex = pxList->pxIndex->pxNext;
        if( ( void * ) pxList->pxIndex == ( void * ) &( pxList->xListEnd ) )
        {
            pxList->pxIndex = pxList->pxIndex->pxNext;
        }
    }
}

void xTaskGetHandle( void )
{
    prvSearchForNameWithinSingleList( &xSuspendedTaskList );
}
```

This makes it clear that the fix in this PR is incidental; the warning can be reproduced without that last assignment in listGET_OWNER_OF_NEXT_ENTRY.

I am also able to trigger it on 02 in addition to 03 and 0s with this. The warning also goes away when configUSE_MINI_LIST_ITEM is 0.

archigup:
https://gcc.gnu.org/onlinedocs/gcc/Warning-Options.html#index-Warray-bounds

Looking at above gcc docs explains why it only happens at those optimization levels; the warning depends on `-ftree-vrp` which is not on all optimization levels. I am able to reproduce the warning at `-00` when `-ftree-vrp` is enabled, and not able to reproduce with  `-03 -fno-tree-vrp`.

archigup:
Narrowing it down further, it seems the issue really is in:
```c
if( ( void * ) pxList->pxIndex == ( void * ) &( pxList->xListEnd ) )
{
    pxList->pxIndex = pxList->pxIndex->pxNext;
}
```

In the right side of the assignment `pxList->pxIndex` is a `ListItem_t *`, but it is known that it points to `pxList->xListEnd` due to the surrounding check; accessing xListEnd through a `ListItem_t *` does extend past the memory region that the specific calls of `prvSearchForNameWithinSingleList` pass in as the pxList parameter. This explains why its only warning where the pxList param is actually a global var. This can be fixed by not accessing it through a type that would extend outside of the memory region. The following equivalent code works:
```c
if( ( void * ) pxList->pxIndex == ( void * ) &( pxList->xListEnd ) )
{
    pxList->pxIndex = pxList->xListEnd.pxNext;
}
```

Due to the if condition, those two are the same value, but the latter code instead accesses it as a `MiniListItem_t` which is in bounds.

The following code also alternatively works:
```c
if( ( void * ) pxList->pxIndex == ( void * ) &( pxList->xListEnd ) )
{
    pxList->pxIndex = ( ( MiniListItem_t * ) pxList->pxIndex )->pxNext;
}
```

paulbartell (approved):

n9wxu (approved):

PR #581, opened 2022-10-21
Added a check wherein a mutex is given only by the owner

sudeep-mohanty:
Mutex type semaphores should be aquired and released by the same task. This commit adds an assertion check in xQueueGenericSend wherein the assertion will fail if a mutex is not given by the owner of the mutex.

Signed-off-by: Sudeep Mohanty <sudp.mohanty@gmail.com>

### DIFF
--- a/queue.c
+++ b/queue.c
@@ -850,11 +850,11 @@ BaseType_t xQueueGenericSend( QueueHandle_t xQueue,
     }
     #endif
 
-#if ( configUSE_MUTEXES == 1 && configCHECK_MUTEX_GIVEN_BY_OWNER == 1 )
-    configASSERT( pxQueue->uxQueueType != queueQUEUE_IS_MUTEX
-                 || pxQueue->u.xSemaphore.xMutexHolder == NULL
-                 || pxQueue->u.xSemaphore.xMutexHolder == xTaskGetCurrentTaskHandle() );
-#endif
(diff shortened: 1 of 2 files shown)


dachalco (approved):

queue.c
```diff
@@ -850,6 +850,12 @@ BaseType_t xQueueGenericSend( QueueHandle_t xQueue,
     }
     #endif
 
+    #if ( configUSE_MUTEXES == 1 && configCHECK_MUTEX_GIVEN_BY_OWNER == 1 )
+        configASSERT( pxQueue->uxQueueType != queueQUEUE_IS_MUTEX ||
+                      pxQueue->u.xSemaphore.xMutexHolder == NULL ||
```
AniruddhaKanhere, line 855: I think this check is added [here](https://github.com/FreeRTOS/FreeRTOS-Kernel/blob/4e2bf2c639cf769aaf26b742e4c11892189ae057/tasks.c#L4164) already.

The call graph is as below:
```
xQueueGenericSend
   |- prvCopyDataToQueue
       |- #if ( configUSE_MUTEXES == 1 )
           {
                if( pxQueue->uxQueueType == queueQUEUE_IS_MUTEX )
                {
                     |- xTaskPriorityDisinherit
                }
           }
```

Let me know if that is incorrect.

AniruddhaKanhere (commented):
Hello @sudeep-mohanty,

Thank you for this PR. I think that this check is already in place. Can you please take a look?

PR #582, opened 2022-10-21
Added a check wherein a mutex is given only by the owner

sudeep-mohanty:
Mutex type semaphores should be aquired and released by the same task. This commit adds an assertion check in xQueueGenericSend wherein the assertion will fail if a mutex is not given by the owner of the mutex.

Signed-off-by: Sudeep Mohanty <sudp.mohanty@gmail.com>

### DIFF
--- a/include/FreeRTOS.h
+++ b/include/FreeRTOS.h
@@ -240,6 +240,10 @@
     #define configUSE_TIMERS    0
 #endif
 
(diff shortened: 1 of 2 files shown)


sudeep-mohanty:
Closing this PR as the check is already covered [here](https://github.com/FreeRTOS/FreeRTOS-Kernel/blob/smp/tasks.c#L5164).

PR #586, opened 2022-11-14
Enable building the GCC Cortex-R5 port without an FPU

paulbartell:
Enable building the GCC Cortex-R5 port without an FPU

Description
-----------
Recently changes to the GCC Cortex-R5 port cause it to require an FPU for operation. Fix this by allowing configUSE_TASK_FPU_SUPPORT == 0 to support non-fpu use cases.


portable/GCC/ARM_CR5/portASM.S
```diff
@@ -45,7 +45,10 @@
 	.extern vTaskSwitchContext
```
paulbartell, line 45: I kept this file with tabs rather than spaces.. Do we have a particular preference for assembly files?

n9wxu (approved):

ChristosZosi:
Hi @paulbartell ,

funnily enough, I did also prepare a PR  to address exactly this issue, but I was bit late to the game 😃 . 

However I took an another approach as yours. At first I tried to address the issue the same way as you are doing in this PR. But there are some issues with your PR.

1) If you are using the `configUSE_TASK_FPU_SUPPORT` option in the `portASM.S`-file you need also to include the `FreeRTOSConfig.h`-header. But, by doing so we restrict the `FreeRTOSConfig.h`-header to have only code understood also by the assembler. One could, however, put C-Code that is not understandable by the assembler inside a macro, e.g.  `#ifdef __ICCARM__/#endif`, but I was not able to find such a macro for the GCC-compiler.
2) In the `portASM.S`-file you still have code accessing the FPU-Registers outside of the `configUSE_TASK_FPU_SUPPORT`

You may look my changes here, how I did solve the issue
https://github.com/FreeRTOS/FreeRTOS-Kernel/compare/main...ChristosZosi:FreeRTOS-Kernel:gcc/cortexr5

portable/GCC/ARM_CR5/portmacro.h
```diff
@@ -116,18 +116,23 @@
...
+ * created without an FPU context and must call vPortTaskUsesFPU() to give
+ * themselves an FPU context before using any FPU instructions. If
+ * configUSE_TASK_FPU_SUPPORT is set to 2 then all tasks will have an FPU context
+ * by default.
+ */
+    #if ( configUSE_TASK_FPU_SUPPORT == 1 )
+        void vPortTaskUsesFPU( void );
+    #else
+
+/*
```
aggarg, line 130: Is the weird alignment result of uncrustify?

paulbartell: Yes.

portable/GCC/ARM_CR5/portASM.S
```diff
@@ -45,7 +45,10 @@
 	.extern vTaskSwitchContext
 	.extern vApplicationIRQHandler
 	.extern ulPortInterruptNesting
+
+#if( configUSE_TASK_FPU_SUPPORT != 0 )
```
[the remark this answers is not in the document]
paulbartell: Good point. I was having trouble with non-assembler compatible definitions in my FreeRTOSConfig. I like @ChristosZosi 's approach using the `__ARM_FP` macro instead, so I'll cherry-pick those commits over here.

Given that libc might use FP registers for things like memcpy, I strongly prefer always having an FP context for each task when the compile has FP enabled rather than configUSE_TASK_FPU_SUPPORT.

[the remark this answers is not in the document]
aggarg: I like that approach too especially the part catches the incorrect configUSE_TASK_FPU_SUPPORT configurations - 
```
#if configUSE_TASK_FPU_SUPPORT == 0
    #ifdef __ARM_FP
        #error The Floating Point Unit (FPU) of the ARM Cortex-R5 is enabled. Therefore you must   \
               define configUSE_TASK_FPU_SUPPORT either to 1 or 2.
    #endif /* __ARM_FP*/
#elif (configUSE_TASK_FPU_SUPPORT == 1) || (configUSE_TASK_FPU_SUPPORT == 2)
    #ifndef __ARM_FP
        #error The Floating Point Unit (FPU) of the ARM Cortex-R5 is disabled. Therefore you must \
               define configUSE_TASK_FPU_SUPPORT to 0.
    #endif /* __ARM_FP*/
#endif /* configUSE_TASK_FPU_SUPPORT */
```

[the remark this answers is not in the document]
ChristosZosi: Hit @paulbartell ,

I am not sure what the typical workflow is, but I could push my changes on top of yours to your PR branch if this is desired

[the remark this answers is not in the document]
paulbartell: I rebased this PR using your branch.

portable/GCC/ARM_CR5/port.c
```diff
@@ -74,25 +74,43 @@
     #error configMAX_API_CALL_INTERRUPT_PRIORITY must be greater than ( configUNIQUE_INTERRUPT_PRIORITIES / 2 )
 #endif
 
-/* Some vendor specific files default configCLEAR_TICK_INTERRUPT() in
- * portmacro.h. */
+#if configUSE_TASK_FPU_SUPPORT == 0
+    #ifdef __ARM_FP
+        #error __ARM_FP is defined, so configUSE_TASK_FPU_SUPPORT must be set to either to 1 or 2.
```
ChristosZosi, line 79: Do we need to make clear that `__ARM_FP` is defined, when target is being compiled with `-mfpu=...`

portable/GCC/ARM_CR5/portmacro.h
```diff
@@ -44,161 +46,174 @@
...
-#else
-    /* Each task has an FPU context already, so define this function away to
-    nothing to prevent it being called accidentally. */
+#elif ( configUSE_TASK_FPU_SUPPORT == 2 )
+
+/*
+ * Each task has an FPU context already, so define this function away to
+ * prevent it being called accidentally.
+ */
     #define vPortTaskUsesFPU()
```
ChristosZosi, line 141: I would suggest to either define `vPortTaskUsesFPU()` for both cases of `configUSE_TASK_FPU_SUPPORT` is `0` and `2` or make `#define portTASK_USES_FLOATING_POINT()` available only for `configUSE_TASK_FPU_SUPPORT` is `1`. 

In the current state `#define portTASK_USES_FLOATING_POINT()` is still available for `configUSE_TASK_FPU_SUPPORT == 0` however  `vPortTaskUsesFPU()` not defined

ChristosZosi (commented):
LGTM except for some minor suggestions

portable/GCC/ARM_CR5/port.c
paulbartell: I added a comment above this line for clarity.

portable/GCC/ARM_CR5/portmacro.h
paulbartell: Done.

ChristosZosi (approved):

n9wxu (dismissed):

amazonKamath:
@n9wxu @paulbartell @aggarg  is this ready for merge?

portable/GCC/ARM_CR5/port.c
```diff
@@ -505,21 +595,23 @@ void FreeRTOS_Tick_Handler( void )
...
-        /* A task is registering the fact that it needs an FPU context.  Set the
-         * FPU flag (which is saved as part of the task context). */
+        /*
+         * A task is registering the fact that it needs an FPU context. Set the
+         * FPU flag (which is saved as part of the task context).
+         */
         ulPortTaskHasFPUContext = pdTRUE;
 
         /* Initialise the floating point status register. */
         __asm volatile ( "FMXR  FPSCR, %0" ::"r" ( ulInitialFPSCR ) : "memory" );
```
Skptak, line 611: Does this initialization of the FPSCR need to happen somewhere else then if configUSE_TASK_FPU_SUPPORT is set to 2? I'm not sure if the register is set to zero by default and don't see it being done anywhere else?

portable/GCC/ARM_CR5/port.c
```diff
@@ -285,34 +333,41 @@ StackType_t * pxPortInitialiseStack( StackType_t * pxTopOfStack,
...
+         */
         pxTopOfStack -= portFPU_REGISTER_WORDS;
         memset( pxTopOfStack, 0x00, portFPU_REGISTER_WORDS * sizeof( StackType_t ) );
 
         pxTopOfStack--;
         *pxTopOfStack = pdTRUE;
         ulPortTaskHasFPUContext = pdTRUE;
     }
-    #else
+    #elif ( configUSE_TASK_FPU_SUPPORT != 0 )
```
Skptak, line 368: Rather than having the #error here, would it make sense to put it in the portmacro.h file [here](https://github.com/FreeRTOS/FreeRTOS-Kernel/pull/586/files#diff-a34306d28e25a2835450b662731644df23ca0d56248c1a9d92e430ce4239741aR133)? I think it might make the actual C code look a bit cleaner that way?

portable/GCC/ARM_CR5/port.c
aggarg: When `configUSE_TASK_FPU_SUPPORT`  is set to 2, initial context created in `pxPortInitialiseStack` sets every stored value of FP context (including FPSCR) to 0 which is then restored correctly in the context restore code.

portable/GCC/ARM_CR5/port.c
aggarg: I'd prefer to have it here as putting in `portmacro.h` results in multiple errors/warnings from all the compilation units including FreeRTOS headers.

Skptak (approved):

PR #589, opened 2022-11-22
Update checks to run on latest Ubuntu version

AniruddhaKanhere:
Description
-----------
With the release of Ubuntu 22.04, the CI is also being updated to use the latest runners in GitHub actions.
To that end, there are few other updated for tools/config files etc. which are required and being added with this PR.

Test Steps
-----------

Related Issue
-----------


oblonka:
OK

Issue #594, opened 2022-11-30
[Feature Request] Enable completely disabling the heap through CMake


CrustyAuklet:
**Is your feature request related to a problem? Please describe.**
I am using FreeRTOS on a very small MCU where we want no heap implementation. This is normally done by
setting `configSUPPORT_DYNAMIC_ALLOCATION` to 0. But I don't see a way to disable the heap via CMake.

**Describe the solution you'd like**
I would like to be able to set the CMake cache variable `FREERTOS_HEAP` to a "falsey" string and have no heap file added to
the project. Strings that CMake considers false are described [here](https://cmake.org/cmake/help/latest/manual/cmake-generator-expressions.7.html#genex:BOOL)

I think this is a pretty simple fix, to change the generator expression to
```
# If FREERTOS_HEAP is digit between 1 .. 5 - it is heap number, otherwise - it is path to custom heap source file
# Disable heap if set to an empty string, 0, FALSE, OFF, N, NO, IGNORE, NOTFOUND, or ends with -NOTFOUND
$<IF:$<BOOL:${FREERTOS_HEAP}>,$<IF:$<BOOL:$<FILTER:${FREERTOS_HEAP},EXCLUDE,^[1-5]$>>,${FREERTOS_HEAP},portable/MemMang/heap_${FREERTOS_HEAP}.c>,>
```
I have this in a branch on my fork [here](https://github.com/CrustyAuklet/FreeRTOS-Kernel/blob/cmake-allow-no-heap/CMakeLists.txt)

**Describe alternatives you've considered**
Looking at the CMakeLists.txt in this repo I don't see a lot of other options. I tried setting `FREERTOS_HEAP` to an empty string and it tried to add `portable/MemMang/heap_.c` to the project. I tried just using a heap file but then I get the error

```
#error This file must not be used if configSUPPORT_DYNAMIC_ALLOCATION is 0
```

**How many devices will this feature impact?**
thousands

**What are your project timelines?**
Working on an MVP shipping date in January

**Additional context**
None.

If you have the same (or similar) feature request, please upvote this issue with thumbs up 👍
and use the comments section to provide answers to the questions above.

jasonpcarroll:
Hi @CrustyAuklet,
[...]

PR #595, opened 2022-11-30
Add option to not include heap in CMakeLists.txt

jasonpcarroll:
Description
-----------

Added an option to disable adding a heap implementation to the FreeRTOS-Kernel in response to https://github.com/FreeRTOS/FreeRTOS-Kernel/issues/594.

Test Steps
-----------

Related Issue
-----------


CMakeLists.txt
```diff
@@ -224,16 +224,25 @@ endif()
...
     queue.c
     stream_buffer.c
     tasks.c
     timers.c
+)
 
-    # If FREERTOS_HEAP is digit between 1 .. 5 - it is heap number, otherwise - it is path to custom heap source file
-    $<IF:$<BOOL:$<FILTER:${FREERTOS_HEAP},EXCLUDE,^[1-5]$>>,${FREERTOS_HEAP},portable/MemMang/heap_${FREERTOS_HEAP}.c>
+# Check if user requested to not inlude a heap implementation
+if(NOT DEFINED FREERTOS_DO_NOT_INCLUDE_HEAP)
```
paulbartell, line 237: Perhaps we should just not include a heap implementation if FREERTOS_HEAP is not defined?

CMakeLists.txt
```diff
@@ -224,16 +224,25 @@ endif()
 
 add_subdirectory(portable)
 
-add_library(freertos_kernel STATIC
+list(APPEND FREERTOS_KERNEL_SOURCES
```
phelter, line 227: Do not create yet another indirection of finding the list of files.

Use either the `add_library()` or `target_sources()` for the file lists.

CMakeLists.txt
phelter: I'm not sure it will compile without the heap definitions.  You will have linker issues with missing symbols.

If you do not want a heap then create a custom heap source that errors out whenever the heap is used/consumed by the API it defines, and do:

```cmake
set(FREERTOS_HEAP "heap_disabled.c")
```

Where `heap_disabled.c` is defined as:

```c

void * pvPortMalloc( size_t xWantedSize )
{
    (void)xWantedSize;

    #if ( configUSE_MALLOC_FAILED_HOOK == 1 )
    {
        vApplicationMallocFailedHook();
    }
    #endif

    return NULL;
}

void vPortFree( void * pv )
{
    ( void ) pv;
    /* Force an assert as it is invalid to call this function. */
    configASSERT( pv == NULL );
}

void vPortInitialiseBlocks( void )
{
}

size_t xPortGetFreeHeapSize( void )
{
    return 0;
}
```

If you want it in the FreeRTOS repo, then suggest making this an alternative heap implementation - as @paulbartell suggested and change the name to heap_0.c (NullOpt - a no-heap implementation of the heap API). - and place it with the other heap locations.  Then you'd also have to change:

```cmake
$<IF:$<BOOL:$<FILTER:${FREERTOS_HEAP},EXCLUDE,^[0-5]$>>,${FREERTOS_HEAP},portable/MemMang/heap_${FREERTOS_HEAP}.c>
```
To allow heap_0.c to be allowed.

phelter (changes requested):

CMakeLists.txt
amazonKamath: @paulbartell @jasonpcarroll  thoughts?

amazonKamath:
@phelter  Thanks for the review and feedback.

CMakeLists.txt
CrustyAuklet: Just a note since I opened the issue #594: It will compile without linker issues if `configSUPPORT_DYNAMIC_ALLOCATION` is set to 0 in the config header.

If adding the `heap_0.c` as suggested above, then I guess I would be relying on the linker to discard those sections. That is ok, but it would be neat in my opinion if that was the default. It seems weird to have to pick a heap implementation if the project is set up to not support dynamic allocation.

phelter: I still stand by my previous comment - adding an additional FREERTOS_HEAP option for the enumeration is better than adding yet another configuration for the same feature.

Since there is a linkage between FREERTOS_HEAP and SUPPORT_DYNAMIC_ALLOCATION then I suggest you make the linkage within the config via [CMakeDependentOption](https://cmake.org/cmake/help/latest/module/CMakeDependentOption.html) - so that it is explicitly known that if SUPPORT_DYNAMIC_ALLOCATION is set to 1, you must configure FREERTOS_HEAP.  Another method of fixing is to remove the SUPPORT_DYNAMIC_ALLOCATION and only use FREERTOS_HEAP=0 instead.

That way you have the error/issue identified even before compile - when building up the configuration instead.

phelter (changes requested):

CMakeLists.txt
n9wxu: configSUPPORT_DYNAMIC_ALLOCATION should be sufficient to deciding to include a heap implementation.  The portable.h header should be changed so the heap functions are not declared when configSUPPORT_DYNAMIC_ALLOCATION==0.  It does make sense to add a dependent section to the CMakeLists to make a heap selection IF configSUPPORT_DYNAMIC_ALLOCATION==1 THOUGH, I would prefer that application decisions like heap & port be made at the application level CMake and not at the kernel level CMake.  This also has the advantage of a simpler path to a custom heap should one be desired.

CMakeLists.txt
```diff
@@ -273,22 +273,26 @@ target_compile_options(freertos_kernel PRIVATE
 
 
 ########################################################################
-add_subdirectory(include)
-add_subdirectory(portable)
+include(${CMAKE_CURRENT_LIST_DIR}/include/CMakeLists.txt)
+include(${CMAKE_CURRENT_LIST_DIR}/portable/CMakeLists.txt)
```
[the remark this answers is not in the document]
Skptak: Can you explain why the usage of `CMAKE_CURRENT_LIST_DIR` is an improper use of CMAKE?

The idea with this change is you no longer need to include the FreeRTOS-Kernel as a sub-directory, it can then be placed anywhere in your project as all of our cmake files aren't position dependent then. So instead of now where the kernel must be included in a sub-directory of your CMake File you would instead be able to include it from anywhere in your project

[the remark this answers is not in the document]
phelter: It's not that it is wrong, it is superfluous (not necessary) and is not required for what you were trying to achieve (the idea with this change).

In any cmake project, one can `add_subdirectory(free_rtos_kernel_dir)` with the appropriate configuration beforehand and that `free_rtos_kernel_dir` can be anywhere in your project.

In recent changes - the CMakeFiles.txt of the various FreeRTOS components were updated to allow a very common way of importing libraries via CMake itself:  `FetchContent`

```cmake
FetchContent_Declare( freertos_kernel
  GIT_REPOSITORY https://github.com/FreeRTOS/FreeRTOS-Kernel.git
  GIT_TAG        V10.6.1
)

...

set(FREERTOS_HEAP "4" CACHE STRING "" FORCE)
set(FREERTOS_PORT "GCC_POSIX" CACHE STRING "" FORCE)

FetchContent_MakeAvailable(freertos_kernel)
```

This will download the V10.6.1 version of the Freertos-Kernel into `build/_deps`
And then run the CMake from a local build directory in `build/_deps`.

To change the version - you then only change one line in the CMakeLists.txt that has the above example.

A great resource is [Professional CMake](https://crascit.com/professional-cmake/) for how to properly use the tool.

CMakeLists.txt
conara: In the CMake you do not have access to configSUPPORT_DYNAMIC_ALLOCATION and configSUPPORT_STATIC_ALLOCATION (from FreeRTOSConfig.h), right? I made a PR to change the default behaviour of FREERTOS_HEAP , see https://github.com/FreeRTOS/FreeRTOS-Kernel/pull/807 . (I was unaware that there were active pull requests attempting to address the same issue.).

chinglee-iot:
Close this PR due to it is included in #807.

PR #601, opened 2022-12-14
Draft: RFC: Enable granular locks for critical sections instead of a single kernel lock

sudeep-mohanty:
Enable granular locks for critical sections instead of a single kernel lock

Description
-----------
This commit updates all critical section APIs, viz., `task*_CRITICAL()`, to accept a spinlock variable. These spinlocks are defined locally to each of the kernel data structures. This change moves away from the single kernel lock design therefore ensuring lesser contentions for critical section locks. This effectively increases the performance of the SMP kernel.

The idea here is to have granularity in the critical section spinlocks. An example of how this improves performance is as below:

`Current Design:`
TaskA running on core#0 is in a critical section by calling a blocking function such as `uxTaskPriorityGet()`.
TaskB running on core#1 is performing a queue operation by calling `xQueueGenericSend()`. TaskB will have to contest for the kernel lock in order to complete the queue operation. This would mean that TaskB will be blocked until TaskA releases the lock.

`Proposed Changes:`
Under the new scheme with the above scenario, TaskB need not wait for TaskA to complete its operation as both the tasks will be contesting for separate locks which are local to the tasks module and the queue module respectively. This enables many kernel related operations to happen simultaneously across the cores, thereby helping in performance improvements in SMP systems.

Things to be considered
-----------
1. Currently the the `task*_CRITICAL()` macros are routed to the `vTaskEnterCritical()` function which ensures atomicity but also has the scope of checking for state change of a task via the function `prvCheckForRunStateChange()`. Under the proposed design change, the task state change may need a rethink. It could be abstracted out to the port layer.

2. Critical section nesting would need to be handled by the ports under the new design. This is currently handled by the kernel directly.

### DIFF
--- a/event_groups.c
+++ b/event_groups.c
@@ -71,6 +71,8 @@ typedef struct EventGroupDef_t
     #if ( ( configSUPPORT_STATIC_ALLOCATION == 1 ) && ( configSUPPORT_DYNAMIC_ALLOCATION == 1 ) )
         uint8_t ucStaticallyAllocated; /*< Set to pdTRUE if the event group is statically allocated to ensure no attempt is made to free the memory. */
     #endif
+
+    portMUX_TYPE xEventGroupLock;
 } EventGroup_t;
 
 /*-----------------------------------------------------------*/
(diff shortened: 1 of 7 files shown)


sudeep-mohanty:
cc: @ Dazza0 @zikalino

PR #611, opened 2023-01-15
Add `ulTaskGetRunTimeCounter` and `ulTaskGetRunTimePercent`

chrisnc:
Allow `ulTaskGetIdleRunTimeCounter` and `ulTaskGetIdleRunTimePercent` to be used whenever `configGENERATE_RUN_TIME_STATS` is enabled, as this is the only requirement for these functions to work.

### DIFF
--- a/.github/lexicon.txt
+++ b/.github/lexicon.txt
@@ -2321,6 +2321,8 @@ ulstoppedtimercompensation
 ultablebase
 ultaskgetidleruntimecounter
 ultaskgetidleruntimepercent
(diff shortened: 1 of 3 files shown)


chrisnc:
Hm, looks like the Kernel-Checker step is failing because it can't find `FreeRTOS/.github/scripts/posix_build_checker.sh`, which was removed here: https://github.com/FreeRTOS/FreeRTOS/pull/889.

archigup (approved):

n9wxu (approved):

PR #618, opened 2023-01-29
fix performance counting for SMP

rgrr:
Description
-----------
ulTaskSwitchedInTime must be an array, index is core number.

On single core devices this change has no effect, but on multicore devices (mine: RP2040) counting is currently wrong, because ulTaskSwitchedInTime is just a scalar type and counts only half of the actual CPU time.

Test Steps
-----------
Performance counting should count configNUM_CORES * (ticks of portALT_GET_RUN_TIME_COUNTER_VALUE()).
Without this change it only counts (ticks of portALT_GET_RUN_TIME_COUNTER_VALUE()).

Related Issue
-----------
None

### DIFF
--- a/tasks.c
+++ b/tasks.c
@@ -396,8 +396,8 @@ PRIVILEGED_DATA static volatile UBaseType_t uxSchedulerSuspended = ( UBaseType_t
 
 /* Do not move these variables to function scope as doing so prevents the
  * code working with debuggers that need to remove the static qualifier. */
-    PRIVILEGED_DATA static uint32_t ulTaskSwitchedInTime = 0UL;    /*< Holds the value of a timer/counter the last time a task was switched in. */
-    PRIVILEGED_DATA static volatile uint32_t ulTotalRunTime = 0UL; /*< Holds the total amount of execution time as defined by the run time counter clock. */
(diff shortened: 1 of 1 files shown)


rgrr:
As far as I understand the code, ulTotalRunTime[] can be a local dynamic variable.  No need to have it module global / static.  If you agree, I will add this to this PR.

chinglee-iot:
@rgrr 
Thank you for your PR. We will look into the modification and replay to you soon.

chinglee-iot:
@rgrr 

Thank you for this PR.

* ulTaskSwitchedInTime should be per core variables. Without this PR, ulTaskSwitchedInTime is overwritten when context switch on the other core and pxCurrentTCB->ulRunTimeCounter is wrong.

* ulTotalRunTime is declared as static global variable in order to work with debuggers. We can reference the comment above the declaration. So I suggest we keep this version PR.
```
/* Do not move these variables to function scope as doing so prevents the
 * code working with debuggers that need to remove the static qualifier. */
 ```

chinglee-iot (approved):
Without this PR, ulTaskSwitchedInTime is overwritten when context switch on the other core and pxCurrentTCB->ulRunTimeCounter is wrong.

shubnil (approved):

PR #619, opened 2023-02-01
Add ifndef for configASSERT

jasa66:
Description
-----------
FreeRTOS-Kernel/ThirdParty/GCC/RP2040/include/portmacro.h in smp-branch calls configASSERT macros, which should not be defined in release builds - macros will increase build size and cause unwanted release behaviour. 
I.e. in release builds this will break the build when configASSERT definition is commented out and macro is not defined.

By adding ifndef in the beginning of the portmacro.h file in case definition does not exist macro will be defined with empty macro.

Test Steps
-----------
Use smp branch of the FreeRTOS-Kernel.
comment out #define configASSERT from your FreeRTOSconfig.h
-->build should break without ifndef definition in portmacro.h

When ifndef added in the beginning of the portmacro.h, compilation will be passed no matter if configASSERT line is defined or commented out in FreeRTOSconfig.h file.

Related Issue
-----------

### DIFF
--- a/portable/ThirdParty/GCC/RP2040/include/portmacro.h
+++ b/portable/ThirdParty/GCC/RP2040/include/portmacro.h
@@ -60,6 +60,10 @@
     typedef int32_t           BaseType_t;
     typedef uint32_t          UBaseType_t;
 
+    #ifndef configASSERT
+        #define configASSERT( x )
+    #endif
+
     #if ( configUSE_16_BIT_TICKS == 1 )
         typedef uint16_t     TickType_t;
         #define portMAX_DELAY              ( TickType_t ) 0xffff


RichardBarry:
This should not be necessary as it's done globally for all ports here https://github.com/FreeRTOS/FreeRTOS-Kernel/blob/smp/include/FreeRTOS.h#L275  If you are having an issue with asserts not being defined away then you may be including header files in the wrong order.  FreeRTOS.h should be included first, then the header files that define the needed API (for example, queue.h).  FreeRTOSConfig.h should never be included explicitly.  Hope this helps.

jasa66:
I made a simple application to verify this. 
In here first include is FreeRTOS.h in main.c
https://github.com/jasa66/smptest

configASSERT is covered correctly everywhere else, but not in file FreeRTOS-Kernel/ThirdParty/GCC/RP2040/include/portmacro.h 

I still get following warnings which lead to linker errors in the end.
Warning:
[build] /Users/jasa/projects/smptest/FreeRTOS-Kernel/portable/ThirdParty/GCC/RP2040/include/portmacro.h: In function 'vPortRecursiveLock':
[build] /Users/jasa/projects/smptest/FreeRTOS-Kernel/portable/ThirdParty/GCC/RP2040/include/portmacro.h:178:9: warning: implicit declaration of function 'configASSERT' [-Wimplicit-function-declaration]
[build]   178 |         configASSERT(ulLockNum >= 0 && ulLockNum < portRTOS_SPINLOCK_COUNT );
[build]       |         ^~~~~~~~~~~~

Linker error:
[build] /Users/jasa/bin/gcc-arm-none-eabi-9-2019-q4-major/bin/../lib/gcc/arm-none-eabi/9.2.1/../../../../arm-none-eabi/bin/ld: CMakeFiles/smptest.dir/FreeRTOS-Kernel/tasks.c.obj: in function `vPortRecursiveLock':
[build] /Users/jasa/projects/smptest/FreeRTOS-Kernel/portable/ThirdParty/GCC/RP2040/include/portmacro.h:178: undefined reference to `configASSERT'

And each configASSERT will cause same warnings and linker errors.

What might be wrong in my application?

Issue #612, opened 2023-01-15
[BUG] xTaskNotifyWait & ulTaskNotifyTake perform a non-deterministic operation in a critical section.


karver8:
**Describe the bug**
The API functions `xTaskNotifyWait()`, `ulTaskNotifyTake()` and their indexed variants both make calls to `prvAddCurrentTaskToDelayedList()` from the inside of a critical section, which in turn calls `vListInsert()` on `pxOverflowDelayedTaskList` or `pxDelayedTaskList` depending on the current tick count(unless `xTicksToWait` equals `portMAX_DELAY`). This results in the kernel walking the list from inside the critical section, which does not happen anywhere else in the library.

This appears to break the rule that _"FreeRTOS never performs a non-deterministic operation, such as walking a linked list, from inside a critical section or interrupt. "_. I found it a little hard to believe something like this would be missed, due to how many eyes are on this code base, so I am half expecting there to be a reason that this rule does not apply. If there is a reason that this rule does not apply here, I think it should be documented.  

I searched around quite a bit for any information on this, and I finally came across [this post from 2018](https://www.freertos.org/FreeRTOS_Support_Forum_Archive/November_2018/freertos_Is_ulTaskNotifyTake_Non-Deterministic_a5bad9540bj.html) with the same concern, which gave me the confidence to make this bug report.

If this is in fact an error, there is one more related issue with a comment inside the function `vTaskPlaceOnEventList()` :
```
/* THIS FUNCTION MUST BE CALLED WITH EITHER INTERRUPTS DISABLED OR THE
 * SCHEDULER SUSPENDED AND THE QUEUE BEING ACCESSED LOCKED. */
```
However this function makes a call to `vListInsert()` and `prvAddCurrentTaskToDelayedList()`, making this function non-deterministic, although there does not appear to be anything that calls this function with interrupts disabled. The comment should be changed to: 
```
/* THIS FUNCTION MUST BE CALLED WITH THE SCHEDULER SUSPENDED AND THE QUEUE BEING ACCESSED LOCKED. */
```

This potentially impacts all systems that use task notifications with block times other than 0 and `portMAX_DELAY`, and were relying on bounded interrupt latency. This issue may also be present in SafeRTOS.

**Solution**
It seems simple to fix this by just suspending the scheduler instead, many other portions of the kernel add tasks to the delayed tasks list doing exactly this.

RichardBarry:
[...]

PR #625, opened 2023-02-13
Fix xTaskNotifyWait & ulTaskNotifyTake determinism.

karver8:
Description
-----------
Fixes the bug described in this issue:

One unorthodox thing being done here, which I feel should be pointed out, is that I am calling `vTaskSuspendAll()` from inside a critical section. This is done to minimize the performance impact of this change, since one of the main benefits of task notifications is that they are faster than the other primitives. The only worry I see here is that it may cause issues with certain ports if they do tricky things inside `portSOFTWARE_BARRIER()` or `portMEMORY_BARRIER()`, but I cannot think of any examples.

Test Steps
-----------
As suggested by Richard Barry, adding `configASSERT( ulCriticalNesting == 0UL )`, into `vListInsert()`, which is a condition that should always be met, will cause it to be triggered by a call to `xTaskNotifyWait()` or `ulTaskNotifyTake()` if the task blocks. This of course only works on ports that have the variable `ulCriticalNesting`.


karver8:
Let me know if there is anything I need to do to move this forward. It looks like unit tests need to be updated in the other repository, these should be pretty simple changes but I am not sure how that would link up to this PR.

n9wxu:
The code seems to work on an RP2040 but fails unit tests.  These failures need to be checked and updated as required.

karver8:
@n9wxu Yes I have looked at the failures and I know how to fix them, but how do I synchronize changes to the unit test repo with changes in this repository? I can submit a PR in the FreeRTOS repo to fix the unit tests, but it looks like that will break the tests until the FreeRTOS-kernel submodule is updated?

tasks.c
```diff
@@ -3137,8 +3137,8 @@ void vTaskPlaceOnEventList( List_t * const pxEventList,
 {
     configASSERT( pxEventList );
 
-    /* THIS FUNCTION MUST BE CALLED WITH EITHER INTERRUPTS DISABLED OR THE
-     * SCHEDULER SUSPENDED AND THE QUEUE BEING ACCESSED LOCKED. */
+    /* THIS FUNCTION MUST BE CALLED WITH THE SCHEDULER SUSPENDED AND THE QUEUE
```
RichardBarry, line 3140: Why was this comment changed?  The original "Interrupts disabled" should probably be "inside a critical section", but I think it's still valid.

karver8: I changed it because this function calls vListInsert(), which shouldn't be called while interrupts are disabled as far as I am aware. I noted this comment in my original issue. But I can change it back if you think this is not correct.

tasks.c
```diff
@@ -4744,22 +4744,25 @@ TickType_t uxTaskResetEventItemValue( void )
...
-                pxCurrentTCB->ucNotifyState[ uxIndexToWait ] = taskWAITING_NOTIFICATION;
+                traceTASK_NOTIFY_TAKE_BLOCK( uxIndexToWait );
 
-                if( xTicksToWait > ( TickType_t ) 0 )
-                {
-                    prvAddCurrentTaskToDelayedList( xTicksToWait, pdTRUE );
-                    traceTASK_NOTIFY_TAKE_BLOCK( uxIndexToWait );
+                /* Suspend the scheduler before enabling interrupts. */
+                vTaskSuspendAll();
+                taskEXIT_CRITICAL();
```
jefftenney, line 4760: Strictly speaking, the API doesn't allow calling vTaskSuspendAll() from inside a critical section.  It probably works, but it would be the only place it happens in the FreeRTOS codebase.

The intent behind the larger construct being proposed here seems to be to avoid suspending/resuming the scheduler if possible, which in turn keeps `ulTaskGenericNotifyTake()` as lightweight as possible.  Maybe a different construct could be used instead.  For example, the idle task uses a different construct to avoid suspending/resuming the scheduler for tickless idle.  A similar approach might work here -- specifically, checking the notification value while "unprotected" to determine whether to suspend the scheduler, and then verifying the notification value after suspending the scheduler if suspending is warranted.

kar-rahul-aws:
Hi @karver8 
We had to create a new PR #[833 ](https://github.com/FreeRTOS/FreeRTOS-Kernel/pull/833) because the kernel code has diverged too much since the time this PR was created. We have ensured to mark you as the commit author and mentioned this in PR [833 ](https://github.com/FreeRTOS/FreeRTOS-Kernel/pull/833)description. We are closing this PR as PR [833](https://github.com/FreeRTOS/FreeRTOS-Kernel/pull/833) has been merged. Thank you for your valuable contribution.
Regards.

karver8:
Thanks! @kar-rahul-aws

PR #626, opened 2023-02-17
POSIX port fixes

jacky309:
Fix POSIX port

Description
-----------
This change fixes a build failure in the POSIX port on my system (WSL). It also introduces an improvement which helps identifying each of the threads by identifying with a name, which is very useful for debugging.

### DIFF
--- a/portable/ThirdParty/GCC/Posix/port.c
+++ b/portable/ThirdParty/GCC/Posix/port.c
@@ -73,7 +73,7 @@
 typedef struct THREAD
 {
     pthread_t pthread;
-    pdTASK_CODE pxCode;
+    TaskFunction_t pxCode;
     void * pvParams;
     BaseType_t xDying;
     struct event * ev;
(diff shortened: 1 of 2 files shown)


portable/ThirdParty/GCC/Posix/port.c
```diff
@@ -461,6 +461,8 @@ static void * prvWaitForStart( void * pvParams )
     uxCriticalNesting = 0;
     vPortEnableInterrupts();
 
+    pthread_setname_np(pxThread->pthread, pcTaskGetName(NULL));
```
alfred2g, line 464: We do not allow *_np (non portable functions), as they would break some  Posix builds (mac os), that are not GNU specific systems
maybe we can surround it with some ifdefs / endif to be omitted on non gnu systems

jacky309: I have removed that change from this PR. I'll probably include it in another PR once I have found a portable way of implementing it.

aggarg (approved):

paulbartell (approved):

PR #641, opened 2023-03-10
Add functions to get the buffers of statically created objects

Dazza0:
# Add `...GetStaticBuffers()` Functions

Description
-----------

This PR adds `...GetStaticBuffers()` functions to get pointers to the buffers used to create a particular static FreeRTOS objects (i.e., objects created using `...CreateStatic()`).

These functions are helpful because some ports (such as ESP-IDF) use the `...CreateStatic()` when an object needs to be allocated from special memory (i.e., not included in the regular `pvPortMalloc()`/`vPortFree()`). "Special Memory" could include things such as memory from a "fast malloc memory pool" or memory from external (i.e., off chip) RAM, and this special memory is allocated/freed using a separate API.

Currently, we need to associate the static object handles and their buffers using some sort of list. For example:

```c
    // Create a task from special memory
    StackType_t *stack_buffer = pvPortSpecialMalloc(2048);
    StaticTask_t *task_obj = pvPortSpecialMalloc(sizeof(StaticTask_t));
    TaskHandle_t task_hdl;
    task_hdl = xTaskCreateStatic(task_func, "task", 2048, NULL, 5, stack_buffer, task_obj);
    // Associate the task_hdl with it's buffers using a linked list
    linked_list_entry_t entry = {
        .task_hdl = task_hdl,
        .stack = stack_buffer,
        .obj = task_obj,
    };
    AddToList(some_list, &entry);

    ...

    // Delete the task
    vTaskDelete(task_hdl);
    // Find the buffers associated with the task and free them
    linked_list_entry_t entry;
    FindFromList(some_list, task_hdl, &entry);
    vPortSpecialFree(entry.stack_buffer);
    vPortSpecialFree(entry.task_obj);
```

But having a `...GetStaticBuffers()` allows us to avoid the list overhead, as we can simply call `...GetStaticBuffers()` to get the associated buffers before the object is deleted. For example:

```c
    // Create a task from special memory
    StackType_t *stack_buffer = pvPortSpecialMalloc(2048);
    StaticTask_t *task_obj = pvPortSpecialMalloc(sizeof(StaticTask_t));
    TaskHandle_t task_hdl;
    task_hdl = xTaskCreateStatic(task_func, "task", 2048, NULL, 5, stack_buffer, task_obj);

    ...

    // Get the task's buffers before deleting it
    StackType_t *stack_buffer_ret;
    StaticTask_t *task_obj_ret;
    xTaskGetStaticBuffers(task_hdl, &stack_buffer_ret, &task_obj_ret);

    // Delete the task
    vTaskDelete(task_hdl);

    // Free the buffers
    vPortSpecialFree(stack_buffer_ret);
    vPortSpecialFree(task_obj_ret);
```

Test Steps
-----------

Checklist:
----------

Related Issue
-----------

### DIFF
--- a/.github/lexicon.txt
+++ b/.github/lexicon.txt
@@ -1464,13 +1464,24 @@ ppdc
 ppio
 ppitc
 ppmc
+ppucmessagebufferstoragearea
+ppucqueuestorage
+ppucstreambufferstoragearea
 ppudr
 ppuer
 ppusr
+ppuxstackbuffer
 ppvdestination
 ppwm
+ppxeventgroupbuffer
 ppxidletaskstackbuffer
 ppxidletasktcbbuffer
+ppxsemaphorebuffer
+ppxstaticmessagebuffer
+ppxstaticqueue
+ppxstaticstreambuffer
+ppxtaskbuffer
+ppxtimerbuffer
 ppxtimertaskstackbuffer
 ppxtimertasktcbbuffer
 pr
@@ -2723,6 +2734,7 @@ xeventgroupcreatestatic
 xeventgroupdelete
 xeventgroupgetbits
 xeventgroupgetbitsfromisr
+xeventgroupgetstaticbuffer
 xeventgroupsetbits
 xeventgroupsetbitsfromisr
 xeventgroupsync
@@ -2796,6 +2808,7 @@ xmessage
 xmessagebuffer
 xmessagebuffercreate
 xmessagebuffercreatestatic
+xmessagebuffergetstaticbuffers
 xmessagebufferisempty
 xmessagebufferisfull
 xmessagebuffernextlengthbytes
@@ -2865,6 +2878,7 @@ xqueuecreatestatic
 xqueuegenericsend
 xqueuegenericsendfromisr
 xqueuegetmutexholder
+xqueuegetstaticbuffers
 xqueuegivefromisr
 xqueuegivemutexrecursive
 xqueueorsemaphore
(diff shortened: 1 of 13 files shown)


paulbartell:
@Dazza0 Thanks for the contribution. The changes seem reasonable to me but I need to have a couple more people take a look at it.

Are you willing to update the unit tests to cover the new functions you've added or would you like some assistance with that?

Dazza0:
@paulbartell I've added some unit tests to https://github.com/FreeRTOS/FreeRTOS/pull/964

paulbartell:
@Dazza0 I made a small change in xTaskGetStaticBuffers to handle the tskSTATICALLY_ALLOCATED_STACK_ONLY case. Otherwise it looks good to me.

Dazza0:
@paulbartell The `tskSTATICALLY_ALLOCATED_STACK_ONLY` changes LGTM

paulbartell (approved):

n9wxu (approved):

aggarg (dismissed):

aggarg (approved):

urutva:
This is not a comment about this PR but on the use-case described.

If a task created with `xTaskCreateStatic` with buffers allocated dynamically deletes itself, that can have unintended side-effect of usage of freed memory. This is because, when a  task deletes itself, the task deletion is deferred to the idle task which still needs to access the TCB for cleanup activities.

Imagine a scenario, where there are two tasks Task A and Task B. Task A is created from main and Task B is created by Task A using `xTaskCreateStatic` with buffers allocated dynamically. When Task B deletes itself, assuming that Task A has higher priority than idle task then Task A is scheduled and if it frees up memory allocated for TCB then when idle task is scheduled, it'll access the TCB of Task B (usage of freed memory).

The use-case described works as expected if the task is not allowed to delete itself. Wondering if this should be documented?

paulbartell (approved):

PR #652, opened 2023-03-30
Add configRESET_TLS_BLOCK macro, adjust formatting.

paulbartell:
Description
-----------
Add configRESET_TLS_BLOCK macro to allow using newlibc functions from Interrupt / Exception handlers.

Repro Steps
-----------
Attempt to call printf from a fault handler / interrupt. 
Note that if the last task to run had not yet initialized stdio, the call may fail due to attempts to allocate heap memory.

Fix
-----------
Add a configRESET_TLS_BLOCK macro which may be called to reset the thread local storage to a particular context for use in interrupt handlers. For newlibc, this sets _impure_ptr = _global_impure_ptr.

Checklist:
----------

### DIFF
--- a/include/FreeRTOS.h
+++ b/include/FreeRTOS.h
@@ -117,19 +117,23 @@
 #if ( configUSE_C_RUNTIME_TLS_SUPPORT == 1 )
 
     #ifndef configTLS_BLOCK_TYPE
-        #error Missing definition:  configTLS_BLOCK_TYPE must be defined in FreeRTOSConfig.h when configUSE_C_RUNTIME_TLS_SUPPORT is set to 1.
+        #error Missing definition:  configTLS_BLOCK_TYPE must be defined when configUSE_C_RUNTIME_TLS_SUPPORT is set to 1.
     #endif
 
     #ifndef configINIT_TLS_BLOCK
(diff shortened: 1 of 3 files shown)


moninom1 (approved):

kar-rahul-aws (approved):

jefftenney:
If an ISR uses `configRESET_TLS_BLOCK()`, would there be an easy way for it to restore the TLS pointer (eg, _impure_ptr) before returning from the interrupt?  That would be important if the ISR returns to the interrupted task without an OS context switch.

The ISR could do the save/restore "by hand" like this:

```
__interrupt__ void myISR( void )
{
    struct _reent *reSave = _impure_ptr;
    configRESET_TLS_BLOCK();   // _impure_ptr = _global_impure_ptr;

    // Body of ISR

    _impure_ptr = reSave;
}
```

But then that begs the question why the macro configRESET_TLS_BLOCK() exists at all since the ISR writer has to know how everything works behind the scenes anyway, including that it works only for newlib.  And even if we can justify the existence of configRESET_TLS_BLOCK(), it would still be insufficient for nested interrupts, which cannot all share the one global reent struct at _global_impure_ptr.  Wondering if it would be better to leave this whole concept (TLS support in interrupt handlers) up to the developer.

PR #668, opened 2023-04-28
paranthesize expression-like macro

tcpluess:
fixed a lint warning

Description
-----------
Lint complains about the macro `configSET_TLS_BLOCK` not being parenthesized. I fixed that in `newlib-freertos.h`.
The macro is used in two places in `tasks.c` and parenthesizing the macro has no effect at all, but removes one lint warning.

Test Steps
-----------
see above.

Related Issue
-----------
none.

### DIFF
--- a/include/newlib-freertos.h
+++ b/include/newlib-freertos.h
@@ -52,7 +52,7 @@
 #endif
 
 #ifndef configSET_TLS_BLOCK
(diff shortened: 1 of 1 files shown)


archigup:
Hey, for information, could you describe which lint with which linter is flagging this?

archigup (approved):

PR #670, opened 2023-05-04
Updated tasks.c checks for scheduler suspension

sudeep-mohanty:
This PR updates the checks for scheduler suspension in `tasks.c` module.

Description
-----------
Currently in the FreeRTOS kernel, the variable `uxSchedulerSuspended` is used to keep track of scheduler suspension. The variable is incremented with every call to `vTaskSuspendAll()`, thereby keeping a track of nested scheduler suspensions. However, in many places in the `tasks.c` module, the check for scheduler suspension is not uniform. At times `uxSchedulerSuspended` is compared to `pdFALSE` giving an impression that `uxSchedulerSuspended` is a boolean.

To avoid non uniform checks of the scheduler state, this PR updates all checks involving `uxSchedulerSuspended` to be tested as follows:
```
uxSchedulerSuspended == ( UBaseType_t ) 0U -> Scheduler not suspended
uxSchedulerSuspended != ( UBaseType_t ) 0U -> Scheduler suspended
```

Checklist:
----------
* [x] I have tested my changes. No regression in existing tests.

### DIFF
--- a/tasks.c
+++ b/tasks.c
@@ -387,7 +387,7 @@ const volatile UBaseType_t uxTopUsedPriority = configMAX_PRIORITIES - 1U;
  * kernel to move the task from the pending ready list into the real ready list
  * when the scheduler is unsuspended.  The pending ready list itself can only be
  * accessed from a critical section. */
-PRIVILEGED_DATA static volatile UBaseType_t uxSchedulerSuspended = ( UBaseType_t ) pdFALSE;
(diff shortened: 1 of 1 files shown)


sudeep-mohanty:
Thank you for taking a look @RichardBarry. I agree with the deterministic test compliance that is needed. I have now updated the PR to not use a macro and the `uxSchedulerSuspended` variable is tested only against `== 0` or `!= 0`.

Updated the PR description accordingly.
PTAL again. Thanks!

aggarg (approved):

kar-rahul-aws (approved):

PR #675, opened 2023-05-09
Fix warnings from portext.asm while using LARGE_DATA_MODEL

moninom1:
Description
-----------
Warnings were observed in portext.asm from portSAVE_CONTEXT and portRESTORE_CONTEXT when using mov.w in large data model. 
Used mov_x  to handle it, on the basis of [data_model.h](https://github.com/FreeRTOS/FreeRTOS-Kernel/blob/1b8a4244bdd911f156ce5bc6f9f13dcde68228cb/portable/CCS/MSP430X/data_model.h#L34) 

Test Steps
-----------

Checklist:
----------

Related Issue
-----------

### DIFF
--- a/portable/CCS/MSP430X/portext.asm
+++ b/portable/CCS/MSP430X/portext.asm
@@ -48,7 +48,7 @@ portSAVE_CONTEXT .macro
 
     ;Save the remaining registers.
     pushm_x #12, r15
-    mov.w   &usCriticalNesting, r14
+    mov_x   &usCriticalNesting, r14
     push_x r14
     mov_x   &pxCurrentTCB, r12
     mov_x   sp, 0( r12 )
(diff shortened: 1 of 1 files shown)


portable/CCS/MSP430X/portext.asm
```diff
@@ -48,7 +48,7 @@ portSAVE_CONTEXT .macro
 
     ;Save the remaining registers.
     pushm_x #12, r15
-    mov.w   &usCriticalNesting, r14
+    mov_x   &usCriticalNesting, r14
```
RichardBarry, line 51: Both these changes look suspicious as after the update the two changed lines exchange data between a 20-bit register and a 16-bit variable.  How has this been tested?

moninom1: Thank you for the review, we have ordered a board for the validation part. I will move the PR to draft till then.

PR #683, opened 2023-05-30
Remove __NVIC_PRIO_BITS and configPRIO_BITS check in port

chinglee-iot:
Remove __NVIC_PRIO_BITS and configPRIO_BITS check in port.

Add a new check if the bits hardware does not implement are zero in configMAX_SYSCALL_INTERRUPT_PRIORITY.
Assertion here helps inform that the configuration is not set to hardware as expected, although hardware ignores these bits. Possible reason here is that __NVIC_PRIO_BITS or configPRIO_BITS is set greater than hardware implemented bits.

Description
-----------
* Remove __NVIC_PRIO_BITS and configPRIO_BITS check in ARMv8, coretex-M3 and cortex-M4
* Check hardware not implemented bits in configMAX_SYSCALL_INTERRUPT_PRIORITY. These bits should be zero.

Test Steps
-----------
Verify on STM32L4S5 platform. STM32L4S5 implments 4 prio bits.
* Define configMAX_SYSCALL_INTERRUPT_PRIORITY with bits not implemented by the hardware by setting __NVIC_PRIO_BITS to 5. The assert catch this error.

Checklist:
----------

Related Issue
-----------
#665 , #673 


ActoryOu (dismissed):

paulbartell (dismissed):
I would strongly prefer that we remove all checks on __NVIC_PRIO_BITS and configPRIO_BITS and instead validate configMAX_SYSCALL_INTERRUPT_PRIORITY with respect to the number of implemented priority bits.

The kernel's job is not to validate that CMSIS macros are set correctly. Rather the kernel just need to validate that configMAX_SYSCALL_INTERRUPT_PRIORITY can be run on the given processor.

RichardBarry:
Paul's suggestion definitely has merit.

aggarg (approved):

ActoryOu (approved):

chinglee-iot:
@paulbartell 
__NVIC_PRIO_BITS and configPRIO_BITS check are removed in this PR. Would you like to review this PR again?

PR #688, opened 2023-06-12
Use portTASK_FUNCTION_PROTO to replace portNORETURN

chinglee-iot:
* Use portTASK_FUNCTION_PROTO to replace portNORETURN for GCC port

Description
-----------

Test Steps
-----------
Consume the FreeRTOS-kernel/CMakeLists.txt with Posix-GCC port. Compile with clang should not have missing-noreturn error.

Checklist:
----------

Related Issue
-----------

### DIFF
--- a/include/FreeRTOS.h
+++ b/include/FreeRTOS.h
@@ -895,10 +895,6 @@
     #define portDONT_DISCARD
 #endif
 
-#ifndef portNORETURN
-    #define portNORETURN
-#endif
-
 #ifndef configUSE_TIME_SLICING
     #define configUSE_TIME_SLICING    1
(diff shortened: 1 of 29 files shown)


chinglee-iot:
The CI fail is because we don't have test case for prvIdleTask yet. In this commit, we added back the static qualifier.

ActoryOu (approved):

aggarg (approved):

billybednar:
It looks like this just deletes portNORETURN for most ports. Was that the intent? The noreturn attribute was only moved to  portTASK_FUNCTION_PROTO in the POSIX port.

PR #696, opened 2023-06-26
Add ePendingReady state for task in a pending ready list

chinglee-iot:
Add ePendingReady state.
The unit test is updated in another PR.

Description
-----------
* Add ePendingReady state for task in a pending ready list.

Test Steps
-----------
Run the unit test case [test_eTaskGetState_success_not_current_tcb_pending_ready](https://github.com/FreeRTOS/FreeRTOS/blob/8f3233e0a0730b16f4fed3669d89954cbdd39ea2/FreeRTOS/Test/CMock/tasks/tasks_1_utest.c#L1724).
* Before this PR, eReady is expected.
* After this PR, ePendingReady should be expected.

Checklist:
----------

Related Issue
-----------

### DIFF
--- a/.github/lexicon.txt
+++ b/.github/lexicon.txt
@@ -607,6 +607,7 @@ epage
 epc
 epclear
 epeds
+ependingready
 ependofwr
 epint
 epread
--- a/include/task.h
+++ b/include/task.h
@@ -95,13 +95,13 @@ typedef BaseType_t (* TaskHookFunction_t)( void * );
 /* Task states returned by eTaskGetState. */
 typedef enum
 {
-    eRunning = 0, /* A task is querying the state of itself, so must be running. */
-    eReady,       /* The task being queried is in a ready list. */
-    ePendingReady,/* The task being queried is in a pending ready list. */
-    eBlocked,     /* The task being queried is in the Blocked state. */
-    eSuspended,   /* The task being queried is in the Suspended state, or is in the Blocked state with an infinite time out. */
-    eDeleted,     /* The task being queried has been deleted, but its TCB has not yet been freed. */
-    eInvalid      /* Used as an 'invalid state' value. */
+    eRunning = 0,  /* A task is querying the state of itself, so must be running. */
+    eReady,        /* The task being queried is in a ready list. */
+    ePendingReady, /* The task being queried is in a pending ready list. */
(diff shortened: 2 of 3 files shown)


jefftenney:
>If a blocked or suspended task is readied by an ISR while the running task suspends the scheduler, the task's state will still be eBlocked or eSuspended with current logic.

@chinglee-iot It seems this issue was recently addressed in #679.  After that PR, the above statement is no longer true.  After #679, the new behavior is: "If a blocked or suspended task is readied by an ISR while the running task suspends the scheduler, the task's state will be eReady."

Is there a benefit to distinguishing `ePendingReady` from `eReady` in the API?  The pending-ready status is a scheduler implementation detail not currently exposed to the API.  From the user's perspective, maybe a task in the pending-ready list really is `eReady`.

RichardBarry:
Reading Jeff's comment, I agree it could be confusing to expose internal implementation detail.  The PR is responding to this scenario: https://github.com/FreeRTOS/FreeRTOS-Kernel/pull/577, so the question is, is that a valid scenario.  Is it valid to call eTaskGetState() when the scheduler is suspended when the generalised rule of thumb is not to call API functions from a critical section or with the scheduler suspended.

The head revision already returns eReady for this scenario.

chinglee-iot:
Agreed with Jeff's and Richard's comment. Let's close this PR and create another for minor update with eReady comment.

jefftenney:
@chinglee-iot Prior to #679, `eTaskGetState()` would return `eBlocked` or `eSuspended` under the conditions being discussed.  After #679, it returns `eReady`.  However, #679 did *not* make the corresponding change to `uxTaskGetSystemState()`, which still returns `eBlocked` or `eSuspended`.  Thus #679 introduced a minor inconsistency.  Addressing that inconsistency would probably make a good PR.

In my opinion returning `eReady` is ideal behavior, even considering that a developer making this API call with the scheduler suspended is suspect, as Richard says.  So I think changing `uxTaskGetSystemState()` to match the current `eTaskGetState()` behavior is best.

chinglee-iot:
@jefftenney I will create another PR for the uxTaskGetSystemState() function. Let's discuss in the new PR.

chinglee-iot:
We created another PR for uxTaskGetSystemState() function. #702

Issue #687, opened 2023-06-10
[BUG] circular dependency in CMake causes linker errors


CrustyAuklet:
**Describe the bug**
There is a circular dependency between CMake targets `freertos_kernel` and `freertos_kernel_port`, causing the final executable to have both archive files linked twice. This causes multiple definition errors if the consuming program uses `--whole-archive`

**Target**
Custom SAML21 board (ARM Cortex-M0) and UNIX port simulator.
Toolchain: gcc-12

**Host**
- Host OS: Ubuntu 20.04, Windows11
- Version: [e.g. Mojave 10.14.6]

**To Reproduce**
- Create a cmake static library that links against `freertos_kernel`
- Create an executable target that links with the static library using `--whole-archive`
- observe multiple definition errors, and in the linker command `freertos_kernel.a` and `freertos_kernel_port.a` are emitted twice each.

**Expected behavior**
each static library should only be in the final link command once

**Additional context**
This is important if creating a "framework" or "SDK" using FreeRTOS. in my static library is the `main()` and `vApplicationGetIdleTaskMemory` functions, reset vectors, etc. Some of these functions will not be linked into the final binary without using `--whole-archive`, but using that will cause errors due to the circular reference in FreeRTOS.

I solved this in my framework, which also has a backend/port structure, by splitting the main target into 2. One interface only library that has the include paths and defines, and the actual main target with all the source files that links against that interface library. `freertos_kernel_port` could now link against that interface library and get the include paths, without depending on the actual `freertos_kernel` static library.

cookpate:
Hello, our team will investigate your use case.

CrustyAuklet:
cool! :)

PR #700, opened 2023-07-05
Fix circular dependency in CMake project

cookpate:
Fix circular dependency in CMake project

Description
-----------

Move includes into an interface library.
The kernel and port link against this interface.
The port no longer links against the kernel, breaking the cycle.

In order for custom ports to also break the cycle, they must link against `freertos_kernel_include` instead of `freertos_kernel`.

Checklist:
----------

This PR introduces no code changes.

Related Issue
-----------

https://github.com/FreeRTOS/FreeRTOS-Kernel/issues/687

### DIFF
--- a/include/CMakeLists.txt
+++ b/include/CMakeLists.txt
@@ -4,8 +4,7 @@ add_library(freertos_kernel_include INTERFACE)
 
 target_include_directories(freertos_kernel_include
     INTERFACE
-        $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}>
-        $<INSTALL_INTERFACE:.>
+        .
         # Note: DEPRECATED but still supported, may be removed in a future release.
         $<$<NOT:$<TARGET_EXISTS:freertos_config>>:${FREERTOS_CONFIG_FILE_DIRECTORY}>
 )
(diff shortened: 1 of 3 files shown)


include/CMakeLists.txt
```diff
@@ -0,0 +1,16 @@
+# FreeRTOS internal cmake file. Do not use it in user top-level project
+
+add_library(freertos_kernel_include INTERFACE)
+
+target_include_directories(freertos_kernel_include
+    INTERFACE
+        $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}>
```
paulbartell, line 7: I'm not familiar with the usage of INSTALL_INTERFACE vs BUILD_INTERFACE. Is this necessary?

cookpate: If someone runs a CMake install script, the INSTALL_INTERFACE generator expression is used instead of the BUILD_INTERFACE generator.

I think the distinction is only needed when specifying absolute paths for INTERFACE include directories. So, I've simplified both to use a relative path.

paulbartell (approved):

AniruddhaKanhere (approved):

PR #719, opened 2023-07-24
Duplicate #625 : Fix xTaskNotifyWait & ulTaskNotifyTake determinism

kar-rahul-aws:
Description
-----------
Fixes the bug described in this issue:

One unorthodox thing being done here, which I feel should be pointed out, is that I am calling vTaskSuspendAll() from inside a critical section. This is done to minimize the performance impact of this change, since one of the main benefits of task notifications is that they are faster than the other primitives. The only worry I see here is that it may cause issues with certain ports if they do tricky things inside portSOFTWARE_BARRIER() or portMEMORY_BARRIER(), but I cannot think of any examples.

Test Steps
-----------
The code seems to work on an RP2040.

Checklist:
----------

Related Issue
-----------
#625 


karver8:
I am curious why the PR I submitted was copied? I am guessing it is just easier this way, which is fine, though I would appreciate if I was made author of the first commit. Some sort of record of my contribution on GitHub would be nice. Obviously you don't have to do this.

Anyways this looks like what I figured would need to happen to get those tests to pass, I have been waiting for a response to my questions about these tests but never received one. Thanks for doing this @kar-rahul-aws.

kar-rahul-aws:
Hi @karver8 
Yes you are right, this duplicate PR was opened , for easier merge conflict resolution since the source code for FreeRTOS Kernel underwent a lot of changes since your PR. I have kept this PR as a draft , and also your old [PR ](https://github.com/FreeRTOS/FreeRTOS-Kernel/pull/625/) open, since we are still working on which one to merge , once the unit tests are fixed. Rest assured, your name will added as the first author for this very commit . Thanks again for the contribution. 
For the unit test fail issue, I have opened a [PR ](https://github.com/FreeRTOS/FreeRTOS/pull/1049). You can have a look and see if this is what you had intended to do .

karver8:
@kar-rahul-aws Thanks for doing this. The addition of listLIST_IS_EMPTY_Expect() is what I wrote down in my notes a while back so that PR seems good to me.

kar-rahul-aws:
Closing since duplicate

PR #742, opened 2023-08-04
Prevent xExpectedIdleTime Underflow with Pre-computation Value Check

owentroe:
Description
-----------

In the case the port/user code introduces off-by-one errors when updating the tick count after `portSUPPRESS_TICKS_AND_SLEEP()`, this minor tweak prevents the `xExpectedIdleTime` from being excessively high due to underflow. This case is easy enough to avoid (by either checking the tick increment value, or making sure sleep happens for less than `xExpectedIdleTime` to ensure no off-by-one edge case), however this change improves the robustness of FreeRTOS making it more immune from poor port or user code. 

This change adds a value check before calculating the `xReturn` value to ensure underflow does not occur by returning `0` instead. 

Test Steps
-----------

Behavior can be reproduced by incrementing `xTickCount` by more than `xExpectedIdleTime` after wakeup in the implementation for `vPortSuppressTicksAndSleep()`. 

Checklist:
----------

Related Issue
-----------

### DIFF
--- a/tasks.c
+++ b/tasks.c
@@ -3485,7 +3485,7 @@ void vTaskSuspendAll( void )
         }
         else
         {
-            xReturn = xNextTaskUnblockTime - xTickCount;
+            xReturn = (xNextTaskUnblockTime > xTickCount)? 0 : xNextTaskUnblockTime - xTickCount;
         }
 
         return xReturn;
--- a/tasks.c
+++ b/tasks.c
@@ -3485,7 +3485,7 @@ void vTaskSuspendAll( void )
         }
         else
         {
-            xReturn = (xNextTaskUnblockTime > xTickCount)? 0 : xNextTaskUnblockTime - xTickCount;
+            xReturn = (xNextTaskUnblockTime > xTickCount)? xNextTaskUnblockTime - xTickCount : 0;
         }
 
         return xReturn;
--- a/tasks.c
+++ b/tasks.c
@@ -3485,7 +3485,7 @@ void vTaskSuspendAll( void )
         }
         else
         {
-            xReturn = (xNextTaskUnblockTime > xTickCount)? xNextTaskUnblockTime - xTickCount : 0;
+            xReturn = ( xNextTaskUnblockTime > xTickCount ) ? ( xNextTaskUnblockTime - xTickCount ) : 0;
         }
 
         return xReturn;


AniruddhaKanhere (dismissed):
This looks good to me. Thank you @owentroe for the contribution.

@aggarg what do you think?

AniruddhaKanhere (dismissed):
This looks good to me. Thank you for the contribution @owentroe.

What do you think @aggarg?

jefftenney:
One concern I have is that the condition being caught here (`xNextTaskUnblockTime < xTickCount`) is potentially deadly for the kernel (if I remember correctly), so it might be misleading to test for it and then imply that it is being handled.

And with as frequently as `prvGetExpectedIdleTime()` is being called, it would be a shame to make it slower unnecessarily.

One question for you @owentroe.  Is it possible for a user implementation of  `portSUPPRESS_TICKS_AND_SLEEP` to cause this condition?  I ask because that function is required to call `vTaskStepTick()` to increment xTickCount, and `vTaskStepTick()` has protections against this condition.

```
    void vTaskStepTick( TickType_t xTicksToJump )
    {
        /* Correct the tick count value after a period during which the tick
         * was suppressed.  Note this does *not* call the tick hook function for
         * each stepped tick. */
        configASSERT( ( xTickCount + xTicksToJump ) <= xNextTaskUnblockTime );
```

owentroe:
@jefftenney Really good catch, I didn't look closely at `vTaskSTepTick`. This change definitely worked around an issue introduced in user code. The behavior was similar to the following (after `xTickCount` surpassed `xNextTaskUnblockTime` without the fix implemented):

1. `xExpectedIdleTime` became higher than expected (need to investigate further). 
2. The user implementation for `portSUPPRESS_TICKS_AND_SLEEP()` would truncate this value and set a wakeup alarm with an onboard timer. 
   * Before returning, this function would step the ticks by the time passed according to an onboard RTC. 
   * Note: the truncated sleep time is much much less than the underflowed `xExpectedIdleTime` value. 
3.  The scheduler would continue on sending the system back to sleep and not waking up any unblocked tasks due to the value error (`xExpectedIdleTime`). 

The fix in this PR mitigated the issues causes by `xTickCount` rising too high (i.e. the scheduler would not stall, and tasks would run as normal). As an aside, it also fixed the system RTOS breaking when pausing in a debugger (I presumed this was due to the RTC continuing to advanced while paused at the WFI instruction, and then passing an excessively large value to `vTaskStepTick`. 

All of the behavior I saw indicates that the `vTaskStepTick` function call is allowing the `xTickCount` to be too large. I need to take a closer look to figure out what is _actually_ going on, but I am now away from the code for the weekend.

owentroe:
Curiosity got the best of me, so I _had_ to remote in. Turns out a peer changed the definition of `configASSERT` in our project to do nothing, so the assert falls through regardless of the condition. 

Still technically user code that broke the protections, but I'm not sure that _this_ PR would be the way to mitigate user code breaking the `configASSERT` macro.

RichardBarry:
If this change masks mistakes in port code then it should not be made, that is what asserts are for.  If we do make this change, then please ensure it complies with the FreeRTOS style guide, which doesn't permit ternary operators.

owentroe:
Agree with @RichardBarry. Closing.

Thank you everyone for the input.

PR #748, opened 2023-08-11
move non-critical code out of critical section.

Moral-Hao:
move non-critical code out of critical section.

Description
-----------

The size calculation in `pvPortMalloc` is non-critical code, we can move it out of critical section, make critical section as small as possible.

Test Steps
-----------

Checklist:
----------

Related Issue
-----------

### DIFF
--- a/portable/MemMang/heap_2.c
+++ b/portable/MemMang/heap_2.c
@@ -159,6 +159,45 @@ void * pvPortMalloc( size_t xWantedSize )
     void * pvReturn = NULL;
     size_t xAdditionalRequiredSize;
 
+    if( xWantedSize > 0 )
+    {
+        /* The wanted size must be increased so it can contain a BlockLink_t
+            * structure in addition to the requested amount of bytes. */
+        if( heapADD_WILL_OVERFLOW( xWantedSize, xHeapStructSize ) == 0 )
+        {
+            xWantedSize += xHeapStructSize;
+
+            /* Ensure that blocks are always aligned to the required number
+                * of bytes. */
+            if( ( xWantedSize & portBYTE_ALIGNMENT_MASK ) != 0x00 )
+            {
+                /* Byte alignment required. */
+                xAdditionalRequiredSize = portBYTE_ALIGNMENT - ( xWantedSize & portBYTE_ALIGNMENT_MASK );
+
+                if( heapADD_WILL_OVERFLOW( xWantedSize, xAdditionalRequiredSize ) == 0 )
+                {
+                    xWantedSize += xAdditionalRequiredSize;
+                }
+                else
+                {
+                    xWantedSize = 0;
+                }
+            }
+            else
+            {
+                mtCOVERAGE_TEST_MARKER();
+            }
+        }
+        else
+        {
+            xWantedSize = 0;
+        }
+    }
+    else
+    {
+        mtCOVERAGE_TEST_MARKER();
+    }
+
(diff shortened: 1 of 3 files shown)


RichardBarry:
> Separately, I feel like the entire body of `pvPortMalloc` does need to have the scheduler suspended.

On first look, I'm not sure that is the case.  There aren't statics or globals used in the section separated out from the code that has the scheduler suspended.

Moral-Hao:
> There aren't statics or globals used in the section separated out from the code that has the scheduler suspended.

Yes, this is why I thought those code should move out.

 

> I don't think _not allowing_ the user to require non-aligned sizes is a behavior we want to have in the API.

The API does allow the user to require non-aligned sizes, the code is just adding some padding space to satisfy the internal align requirement. And these code was already in heap_4 and heap_5, I just move them out of scheduler  suspended section.
For heap_2,  The original code is :
`xAdditionalRequiredSize = heapSTRUCT_SIZE + portBYTE_ALIGNMENT - ( xWantedSize & portBYTE_ALIGNMENT_MASK );`
Compare to heap_4/heap_5, these code may allocate additional unnecessary space when the user required size is already aligned, so I replace the original code with the same code in heap_4/heap_5.

ydhuang28:
Upon a second look I think both of you are correct @RichardBarry @Moral-Hao. However, in `heap_2.c`, the size of the heap struct is defined as `heapSTRUCT_SIZE`; @Moral-Hao could you update the PR to use that in `heap_2` instead of `xHeapStructSize`?

I can approve the PR after the change is made.

Moral-Hao:
@ydhuang28  Sorry about the mistake about `xHeapStructSize` in heap_2.c.
It looks like most of other variables are start with `x`, do you think change the variable name `heapSTRUCT_SIZE` in heap_2.c to `xHeapStructSize` is better? 
And do you have any suggest to help me find these kind of silly mistake in the early phase?

ydhuang28:
@RichardBarry Right, it's the `xWantedSize <= xFreeBytesRemaining` that guards the overflow.

@Moral-Hao I would leave it as is because there may be application code using `heap_2.c` that's using this macro as it's currently defined so that would break their code. Happy to approve after you make the change to `heapSTRUCT_SIZE` though.

ydhuang28:
@Moral-Hao The FreeRTOS source code naming convention is located [here](https://www.freertos.org/FreeRTOS-Coding-Standard-and-Style-Guide.html), FYI.

Moral-Hao:
@ydhuang28 Thank you. According to the naming conventions, the constant variable in heap_2.c should be named as `usHeapStructSize`, which is different  from the name in heap_4.c/heap_5.c, because the variable type in heap_2.c is `uint16_t`  and the variable type in heap_4.c and heap_5.c is `size_t`.
As the variable is declared as `static`, there is noway the application code can use this constant variable. I think it is safe to change the variable's name and type.

ydhuang28:
It should probably be changed to be a `size_t` as well, since `xWantedSize` is a `size_t`?

Moral-Hao:
> It should probably be changed to be a `size_t` as well, since `xWantedSize` is a `size_t`?

Yes, that what I'm going to do.

aggarg (approved):

tony-josi-aws (approved):

htibosch:
I totally agree with the proposed change: it reduces the time that the scheduler is suspended during every call to `pvPortMalloc()`. Thanks

PR #750, opened 2023-08-13
Initial SMP review - DO NOT MERGE

RichardBarry:
Description
-----------
DO NOT MERGE THIS PR – it is only intended as a code review

This PR contains multiple “`_RB_`” tags in comments highlighting review comments I left in the code.  Some are trivial and just stylistic, for example, where macro names are not consistent with the kernel’s style guide.  More fundamentally though, I would like us to experiment with reducing use of the pre-processor to split the code into single and multi-core versions.  I feel this can be done by:
1)	Using the same pxCurrentTCB[] array no matter the setting of configNUMBER_OF_CORES, just as the same xYieldPends[] array is already used in all cases.  The loops through the pxCurrentTCB[] array should still work even when there is only one position in the array.
2)	Implementing single core versions of the macros that currently only exist when configNUMBER_OF_CORES > 1.  Hopefully some of these can be defined as constants, or even just defined as nothing, when configNUMBER_OF_CORES == 1 to reduce code size impact.  If successful, this will reduce duplication, testing and maintenance, and improve readability.  Currently single core configurations are not exercising much of the same code as multicore configurations.
So far, I’ve reviewed the first quarter of the file to prevent adding the same comment in multiple places, and not commented on design.

### DIFF
--- a/tasks.c
+++ b/tasks.c
@@ -58,7 +58,7 @@
     #include <stdio.h>
 #endif /* configUSE_STATS_FORMATTING_FUNCTIONS == 1 ) */
 
-#if ( configNUMBER_OF_CORES == 1 )
+#if ( configNUMBER_OF_CORES == 1 ) /*_RB_ Not clear why this only applies to single core - please add an explanation. */
     #if ( configUSE_PREEMPTION == 0 )
 
 /* If the cooperative scheduler is being used then a yield should not be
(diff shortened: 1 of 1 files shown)


chinglee-iot:
Commits addresses the review feedback are merged into main branch. Let's close this PR and use new PR for other suggestion.

PR #769, opened 2023-08-31
Update taskYIELD_IF_USING_PREEMPTION macro

chinglee-iot:
Update `taskYIELD_IF_USING_PREEMPTION `to align single core and SMP implementation.

Description
-----------
`taskYIELD_IF_USING_PREEMPTION `is called in the following two scenarios in single core implementation:
* Request the current running task to yield.
* Request any running task to yield for the task.

In single core implementation both of the scenarios is achieved by calling `portYIELD_WITHIN_API()`. However, these scenarios are handled differently in SMP. `prvYieldCore()` is called when requesting a running task to yield and `prvYieldForTask()` is called when yielding a running task for higher priority task.

This PR tries to align these scenarios with the same macro in single core and SMP. A new macro `taskYIELD_ANY_CORE_IF_USING_PREEMPTION` is added. Existing `taskYIELD_IF_USING_PREEMPTION` macro is renamed to `taskYIELD_TASK_CORE_IF_USING_PREEMPTION` with one parameter.

Single core implementation
```c
#define taskYIELD_TASK_CORE_IF_USING_PREEMPTION( pxTCB ) portYIELD_WITHIN_API()
#define taskYIELD_ANY_CORE_IF_USING_PREEMPTION( pxTCB )  \
if( pxCurrentTCB->uxPriority < ( pxTCB )->uxPriority ) \
{ \
    portYIELD_WITHIN_API(); \
}
```

SMP implementation
```
#define taskYIELD_TASK_CORE_IF_USING_PREEMPTION( pxTCB ) prvYieldCore( ( pxTCB )->xTaskRunState )
#define taskYIELD_ANY_CORE_IF_USING_PREEMPTION( pxTCB )  prvYieldForTask( pxTCB )
```

Checklist:
----------

Related Issue
-----------
Address #750 comment in line 61.


tasks.c
```diff
@@ -58,16 +58,42 @@
...
-    #endif
-#endif /* if ( configNUMBER_OF_CORES == 1 ) */
+    #define taskYIELD_IF_USING_PREEMPTION( pxTCB )
+    #define taskYIELD_FOR_TASK_IF_USING_PREEMPTION( pxTCB )
+#else
+
+    #if ( configNUMBER_OF_CORES == 1 )
+
+/* pxTCB is not used since the task can only runs on core 0 in single core. */
+        #define taskYIELD_IF_USING_PREEMPTION( pxTCB )    portYIELD_WITHIN_API()
```
RichardBarry, line 72: Why is the parameter needed here?  If it is needed it may need to be defined as follows to prevent warnings about the parameter not being used: { portYIELD_WITHIN_API(); ( void ) pxTCB; }

RichardBarry (changes requested):
See inline comment.

tasks.c
chinglee-iot: This suggestion makes sense.
If taskYIELD_IF_USING_PREEMPTION is used in the following code, there will be compiler warning.
```c
void func( TCB_t * pxTCB )
{
    /* The function assumes pxTCB is used in this macro. However, this macro doesn't use pxTCB in
     * single core scheduler. There will be compiler warning. It is better taskYIELD_IF_USING_PREEMPTION
     * is defined as suggestion. */
    taskYIELD_IF_USING_PREEMPTION( pxTCB );
}
```
The macro is updated and more comments to explain why the pxTCB is not used in this macro.

aggarg (approved):

ActoryOu (approved):

kar-rahul-aws (approved):

chinglee-iot:
@RichardBarry This PR is updated. Would you like to review this PR again?

PR #774, opened 2023-09-02
Update block validate macro in heap_5

chinglee-iot:
Update block validate macro in heap_5

Description
-----------
When configENABLE_HEAP_PROTECTOR is not set in heap_5, `pucHeapHighAddress `and `pucHeapLowAddress` won't be updated.
The following assert is always triggered.
```
    #define heapVALIDATE_BLOCK_POINTER( pxBlock )                       \
    configASSERT( ( pucHeapHighAddress != NULL ) &&                     \
                  ( pucHeapLowAddress != NULL ) &&                      \
                  ( ( uint8_t * ) ( pxBlock ) >= pucHeapLowAddress ) && \
                  ( ( uint8_t * ) ( pxBlock ) < pucHeapHighAddress ) )
```
This PR upate the `heapVALIDATE_BLOCK_POINTER` macro when configENABLE_HEAP_PROTECTOR is not set to 1.

Test Steps
-----------
Run the WIN32-MSVC demo 
Before this pull request
```
Trace started.
The trace will be dumped to the file "Trace.dump" whenever a call to configASSERT()
fails or the 't' key is pressed.
Note that the trace output uses the ring buffer mode, meaning that the output trace
will only be the most recent data able to fit within the trace recorder buffer.

Starting the full demo.
ASSERT! Line 264, file C:\msys64\home\chinglee\kernel\heap_verify\FreeRTOS\FreeRTOS\S
ource\portable\MemMang\heap_5.c, GetLastError() 0

Trace output saved to Trace.dump
```

After this PR
```
Trace started.
The trace will be dumped to the file "Trace.dump" whenever a call to configASSERT()
fails or the 't' key is pressed.
Note that the trace output uses the ring buffer mode, meaning that the output trace
will only be the most recent data able to fit within the trace recorder buffer.

Starting the full demo.
```

Checklist:
----------

Related Issue
-----------

### DIFF
--- a/portable/MemMang/heap_5.c
+++ b/portable/MemMang/heap_5.c
@@ -143,29 +143,31 @@ typedef struct A_BLOCK_LINK
 /* Canary value for protecting internal heap pointers. */
     PRIVILEGED_DATA static portPOINTER_SIZE_TYPE xHeapCanary;
 
-
 /* Macro to load/store BlockLink_t pointers to memory. By XORing the
  * pointers with a random canary value, heap overflows will result
  * in randomly unpredictable pointer values which will be caught by
  * heapVALIDATE_BLOCK_POINTER assert. */
     #define heapPROTECT_BLOCK_POINTER( pxBlock )    ( ( BlockLink_t * ) ( ( ( portPOINTER_SIZE_TYPE ) ( pxBlock ) ) ^ xHeapCanary ) )
 
-#else /* if ( configENABLE_HEAP_PROTECTOR == 1 ) */
-
-    #define heapPROTECT_BLOCK_POINTER( pxBlock )    ( pxBlock )
-
-#endif /* configENABLE_HEAP_PROTECTOR */
-
-/* Highest and lowest heap addresses used for heap block bounds checking. */
-PRIVILEGED_DATA static uint8_t * pucHeapHighAddress = NULL;
-PRIVILEGED_DATA static uint8_t * pucHeapLowAddress = NULL;
-
 /* Assert that a heap block pointer is within the heap bounds. */
-#define heapVALIDATE_BLOCK_POINTER( pxBlock )                           \
+    #define heapVALIDATE_BLOCK_POINTER( pxBlock )                       \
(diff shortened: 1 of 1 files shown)


tony-josi-aws (approved):

sirnish (approved):

portable/MemMang/heap_5.c
```diff
@@ -143,30 +143,32 @@ typedef struct A_BLOCK_LINK
...
 /* Assert that a heap block pointer is within the heap bounds. */
-#define heapVALIDATE_BLOCK_POINTER( pxBlock )                           \
+    #define heapVALIDATE_BLOCK_POINTER( pxBlock )                       \
     configASSERT( ( pucHeapHighAddress != NULL ) &&                     \
                   ( pucHeapLowAddress != NULL ) &&                      \
                   ( ( uint8_t * ) ( pxBlock ) >= pucHeapLowAddress ) && \
                   ( ( uint8_t * ) ( pxBlock ) < pucHeapHighAddress ) )
 
+/* Highest and lowest heap addresses used for heap block bounds checking. */
+    PRIVILEGED_DATA static uint8_t * pucHeapHighAddress = NULL;
```
Skptak, line 160: Why were these variables moved below the macro?
Just worried this might lead to a compilation warning at some point as these variables the macro are referencing aren't defined when the macro is.

chinglee-iot: The macro will be used in code only. Therefore, it should not has compilation warning since the variables are defined before this macro is used in code.

Another problem related to your comment is that It looks like the declaration of macro, static variables and extern function are mixed together. Probably we can put them into independent section for readability.
I would like to create another PR to update this in heap_4.c and heap_5.c.

kar-rahul-aws (approved):

portable/MemMang/heap_5.c
chinglee-iot: Update for readability.

Skptak (approved):

PR #804, opened 2023-09-21
Kernel Checker CI Workflow File Updates

Skptak:
Description
-----------
Use echo groups on the header check
Perform sparse checkout of just the .github folder for the header check instead of all the files.
Update python checkout version being used.
Update the version of the get changed files action being used.

Test Steps
-----------

Checklist:
----------

Related Issue
-----------

### DIFF
--- a/.github/workflows/kernel-checks.yml
+++ b/.github/workflows/kernel-checks.yml
@@ -9,10 +9,7 @@ jobs:
     steps:
       # Install python 3
       - name: Tool Setup
-        uses: actions/setup-python@v2
-        with:
-          python-version: 3.7.10
-          architecture:   x64
+        uses: actions/setup-python@v3
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
 
@@ -21,6 +18,7 @@ jobs:
         uses: actions/checkout@v3
         with:
           repository: FreeRTOS/FreeRTOS
+          sparse-checkout: '.github'
           ref:  main
           path: tools
 
@@ -32,7 +30,7 @@ jobs:
 
       # Collect all affected files
       - name: Collecting changed files
-        uses: lots0logs/gh-action-get-changed-files@2.1.4
+        uses: lots0logs/gh-action-get-changed-files@2.2.2
         with:
           token: ${{ secrets.GITHUB_TOKEN }}
 
--- a/.github/workflows/kernel-checks.yml
+++ b/.github/workflows/kernel-checks.yml
@@ -35,10 +35,35 @@ jobs:
           token: ${{ secrets.GITHUB_TOKEN }}
 
       # Run checks
-      - name: Check File Headers
+      - env:
+          bashPass: \033[32;1mPASSED -
+          bashInfo: \033[33;1mINFO -
+          bashFail: \033[31;1mFAILED -
+          bashEnd:  \033[0m
+          stepName: Check File Headers
+        name: ${{ env.stepName }}
+        shell: bash
         run: |
+          # ${{ env.stepName }}
+          echo -e "::group::${{ env.bashInfo }} Install Dependencies ${{ env.bashEnd }}"
+
+          # Copy the common tools from the FreeRTOS/FreeRTOS repo.
           mv tools/.github/scripts/common inspect/.github/scripts
+
+          # Install the necessary python dependencies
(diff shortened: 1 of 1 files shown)


.github/workflows/kernel-checks.yml
```diff
@@ -9,10 +9,7 @@ jobs:
     steps:
       # Install python 3
       - name: Tool Setup
-        uses: actions/setup-python@v2
-        with:
-          python-version: 3.7.10
-          architecture:   x64
+        uses: actions/setup-python@v3
```
jasonpcarroll, line 12: Should we not set python version/architecture? Obv the default must work... so this is a nitpick more than anything. I guess I just worry that we'll update to a later version and this suddenly breaks if the newer version has a different default.

.github/workflows/kernel-checks.yml
```diff
@@ -32,15 +30,40 @@ jobs:
 
       # Collect all affected files
       - name: Collecting changed files
-        uses: lots0logs/gh-action-get-changed-files@2.1.4
+        uses: lots0logs/gh-action-get-changed-files@2.2.2
```
jasonpcarroll, line 33: I'm personally not opposed to using third party actions... but it does pose a security concern as the third party could modify their code to do nefarious things in theory - esp since you pass the github token.

Skptak: > I'm personally not opposed to using third party actions.
This already exists here. It's a version update since the current action version is soon to be deprecated
[Link](https://github.com/FreeRTOS/FreeRTOS-Kernel/actions/runs/6265515088)
```
<html><body>
<!--StartFragment-->

FreeRTOS Kernel Header Checks                                                                                                                           The following actions uses node12 which  is deprecated and will be forced to run on node16:  actions/setup-python@v2, lots0logs/gh-action-get-changed-files@2.1.4.  For more info:  https://github.blog/changelog/2023-06-13-github-actions-all-actions-will-run-on-node16-instead-of-node12-by-default/
--
 


                      <a class="d-block no-underline Link--secondary color-fg-default" href="https://github.com/FreeRTOS/FreeRTOS-Kernel/actions/runs/6265515088/job/17014521341#step:5:13">
                          <strong>FreeRTOS Kernel Header Checks</strong>
                        
                      </a>
                      <div class="color-fg-muted">
                        <div data-target="annotation-message.annotationContainer" class="text-small text-mono annotation--contracted">
                          <div>The `set-output` command is deprecated 
and will be disabled soon. Please upgrade to using Environment Files. 
For more information see: 
https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/</div>
                        </div></div><!--EndFragment-->
</body>
</html>
```
>  but it does pose a security concern as the third party could modify their code to do nefarious things in theory
This is a tagged commit.

> esp since you pass the github token.
Any PR from a fork is granted read only permissions

.github/workflows/kernel-checks.yml
Skptak: You're right being forced to a new version could lead to a breaking script

Much in the way that being tied to an old version that is deprecated and will be dropped could lead to a breaking script
```
The following actions uses node12 which is deprecated and will be forced to run on node16: actions/setup-python@v2, lots0logs/gh-action-get-changed-files@2.1.4. For more info: https://github.blog/changelog/2023-06-13-github-actions-all-actions-will-run-on-node16-instead-of-node12-by-default/
```

jasonpcarroll: Right but can you not specify the python-version field as was done before for setup-python@v2?

.github/workflows/kernel-checks.yml
jasonpcarroll: >This already exists here. It's a version update since the current action version is soon to be deprecated

I know it is already used - I can see the commit change. I'm still pointing it out. 

>This is a tagged commit.

Just because its tagged doesn't mean anything - an owner of a repository could just delete the tag and set it to a new commit...  

>Any PR from a fork is granted read only permissions on the token

Ok, but would if someone manually triggered it... suddenly third party action has an owner token.

.github/workflows/kernel-checks.yml
Skptak: Not that version.

bradleysmith23 (approved):

archigup (approved):

PR #805, opened 2023-09-22
Add option to set the core affinity for the Timer Svc Task on SMP systems

sudeep-mohanty:
Title
-----------
Make Timer Service Task core affinity configurable

Description
-----------
This PR introduces the `configTIMER_SERVICE_TASK_CORE_AFFINITY` config option which allows a user to set the core affinity of the Timer Service Task on SMP systems. The motivation to have this feature comes from requests such as [this](https://github.com/espressif/esp-idf/issues/12204#issuecomment-1709830113) and [this](https://github.com/espressif/esp-idf/issues/10315) on the ESP-IDF repository. It would be nice if the FreeRTOS-Kernel can support this natively.

Test Steps
-----------
N.A

Checklist:
----------

Related Issue
-----------
None

### DIFF
--- a/timers.c
+++ b/timers.c
@@ -60,11 +60,20 @@
     #define tmrNO_DELAY                    ( ( TickType_t ) 0U )
     #define tmrMAX_TIME_BEFORE_OVERFLOW    ( ( TickType_t ) -1 )
 
-/* The name assigned to the timer service task.  This can be overridden by
- * defining trmTIMER_SERVICE_TASK_NAME in FreeRTOSConfig.h. */
+/* The name assigned to the timer service task. This can be overridden by
+ * defining configTIMER_SERVICE_TASK_NAME in FreeRTOSConfig.h. */
     #ifndef configTIMER_SERVICE_TASK_NAME
         #define configTIMER_SERVICE_TASK_NAME    "Tmr Svc"
     #endif
+
+    #if ( ( configNUMBER_OF_CORES > 1 ) && ( configUSE_CORE_AFFINITY == 1 ) )
+
+/* The core affinity assigned to the timer service task on SMP systems.
+ * This can be overridden by defining configTIMER_SERVICE_TASK_CORE_AFFINITY in FreeRTOSConfig.h. */
+        #ifndef configTIMER_SERVICE_TASK_CORE_AFFINITY
+            #define configTIMER_SERVICE_TASK_CORE_AFFINITY    tskNO_AFFINITY
+        #endif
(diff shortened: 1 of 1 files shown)


timers.c
```diff
@@ -243,43 +251,89 @@
...
+            if( xTimerQueue != NULL )
+            {
+                #if ( configSUPPORT_STATIC_ALLOCATION == 1 )
+                {
+                    StaticTask_t * pxTimerTaskTCBBuffer = NULL;
+                    StackType_t * pxTimerTaskStackBuffer = NULL;
+                    uint32_t ulTimerTaskStackSize;
+
+                    vApplicationGetTimerTaskMemory( &pxTimerTaskTCBBuffer, &pxTimerTaskStackBuffer, &ulTimerTaskStackSize );
+                    xTimerTaskHandle = xTaskCreateStaticiAffinitySet( prvTimerTask,
```
archigup, line 305:
```suggestion
                    xTimerTaskHandle = xTaskCreateStaticAffinitySet( prvTimerTask,
```

archigup:
Hey, thanks for the PR! 

Could the `#if ( configNUMBER_OF_CORES == 1 )` be moved inwards to just around the task create functions? Seems the rest of the logic in the blocks are the same.

sudeep-mohanty:
Thanks for the review @archigup. Updated as suggested.

chinglee-iot:
@sudeep-mohanty 
Thank you for creating the feedback. The issue in your description indicates that timer task is pinned to core 0 and blocked by higher priority tasks on core 0. As a result, the application watchdog callback can not be called within timeout. The alternative solution also suggests that allowing tasks to run on any core could be a better solution, which is the current setting for timer task. Therefore, this should not be an issue in FreeRTOS SMP.

On the other hand, user can consider to create another task with core affinity to serve application specific requirement instead of changing the affinity of timer task cause it may have effect on software timer service.

sudeep-mohanty:
Thanks for the feedback @chinglee-iot. The idea here is to give users the flexibility of either punning the timer task to a core or letting it run unpinned. Ofcourse, all issues with a lower priority task getting blocked by a higher priority task remain even on an SMP system irrespective of the task's core affinity, so IMO this PR shouldn't cause change to how the timer task runs in practice. A counter argument could be to let the timer task run on a core uninterrupted for a potentially more responsive timer service which is not possible today.

aggarg (approved):

chinglee-iot (approved):

PR #833, opened 2023-10-13
Fix xTaskNotifyWait & ulTaskNotifyTake determinism.

kar-rahul-aws:
Description
-----------
Fixes the bug described in this issue: https://github.com/FreeRTOS/FreeRTOS-Kernel/issues/612. This was originally contributed in this PR: https://github.com/FreeRTOS/FreeRTOS-Kernel/pull/625.

The implementation suspends the scheduler before exiting the critical section (i.e. before enabling interrupts).  If we do not do so, a notification sent from an ISR, which happens after exiting the critical section and before suspending the scheduler, will get lost. The sequence of events will be:
1. Exit critical section.
2. Interrupt - ISR calls `xTaskNotifyFromISR` which adds the task to the Ready list.
3. Suspend scheduler.
4. `prvAddCurrentTaskToDelayedList` moves the task to the delayed or suspended list.
5. Resume scheduler does not touch the task (because it is not on the pendingReady list), effectively losing the notification from the ISR.

The same does not happen when we suspend the scheduler before exiting the critical section. The sequence of events in this
case will be:
1. Suspend scheduler.
2. Exit critical section.
3. Interrupt - ISR calls `xTaskNotifyFromISR` which adds the task to the pendingReady list as the scheduler is suspended.
4. `prvAddCurrentTaskToDelayedList` adds the task to delayed or suspended list. Note that this operation does not nullify
   the add to pendingReady list done in the above step because a different list item, namely xEventListItem, is used for
   adding the task to the pendingReady list. In other words, the task still remains on the pendingReady list.
5. Resume scheduler moves the task from pendingReady list to the Ready list.

Test Steps
-----------
Tested by adding a dummy loop in the implementation to cause a button press interrupt at the desired point and ensured that the notification is not lost. 

Checklist:
----------

Related Issue
-----------

### DIFF
--- a/tasks.c
+++ b/tasks.c
@@ -5127,7 +5127,7 @@ void vTaskPlaceOnEventList( List_t * const pxEventList,
 
     configASSERT( pxEventList );
 
-    /* THIS FUNCTION MUST BE CALLED WITH EITHER INTERRUPTS DISABLED OR THE
+    /* THIS FUNCTION MUST BE CALLED WITH THE
      * SCHEDULER SUSPENDED AND THE QUEUE BEING ACCESSED LOCKED. */
 
     /* Place the event list item of the TCB in the appropriate event list.
@@ -7415,7 +7415,6 @@ TickType_t uxTaskResetEventItemValue( void )
         configASSERT( uxIndexToWaitOn < configTASK_NOTIFICATION_ARRAY_ENTRIES );
 
         taskENTER_CRITICAL();
-        {
             /* Only block if the notification count is not already non-zero. */
             if( pxCurrentTCB->ulNotifiedValue[ uxIndexToWaitOn ] == 0UL )
             {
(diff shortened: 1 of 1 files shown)


tasks.c
```diff
@@ -7415,23 +7415,39 @@ TickType_t uxTaskResetEventItemValue( void )
...
-            /* Only block if the notification count is not already non-zero. */
-            if( pxCurrentTCB->ulNotifiedValue[ uxIndexToWaitOn ] == 0UL )
+            /* Mark this task as waiting for a notification. */
+            pxCurrentTCB->ucNotifyState[ uxIndexToWaitOn ] = taskWAITING_NOTIFICATION;
+
+            if( xTicksToWait > ( TickType_t ) 0 )
             {
-                /* Mark this task as waiting for a notification. */
-                pxCurrentTCB->ucNotifyState[ uxIndexToWaitOn ] = taskWAITING_NOTIFICATION;
+                prvAddCurrentTaskToDelayedList( xTicksToWait, pdTRUE );
```
htibosch, line 7427: Hi @kar-rahul-aws, it looks like you want to remove this call to `prvAddCurrentTaskToDelayedList()` and move it to the section where the scheduler is suspended?

kar-rahul-aws: Hi @htibosch , updated , thanks.

aggarg (approved):

chinglee-iot (approved):

PR #841, opened 2023-10-19
Fix MISRA_C_2012 rule 9.1 violation

kar-rahul-aws:
Fix MISRA C 2012 rule 9.1 violation

Description
-----------
**MISRA C 2012 rule 9.1**

> The value of an object with automatic storage duration shall not be read before it has been set

**MISRA violation**

The following is an example that the default value of an object is not set, before the value from that object is read.
The function prototype:
```
static void prvProcessReceivedCommands( void )
    {
        DaemonTaskMessage_t xMessage;
        Timer_t * pxTimer;
        BaseType_t xTimerListsWereSwitched;
        TickType_t xTimeNow;

        while( xQueueReceive( xTimerQueue, &xMessage, tmrNO_DELAY ) != pdFAIL )
        {
              // remaining function task
        }
    }
```

**Fix in this PR**

- Add default values for objects , which have not been set , before they are read from.

Test Steps
-----------
N/A

Checklist:
----------


tasks.c
```diff
@@ -7274,7 +7274,8 @@ static void prvResetNextTaskUnblockTime( void )
         int iSnprintfReturnValue;
         BaseType_t xOutputBufferFull = pdFALSE;
         UBaseType_t uxArraySize, x;
-        configRUN_TIME_COUNTER_TYPE ulTotalTime, ulStatsAsPercentage;
+        configRUN_TIME_COUNTER_TYPE ulTotalTime = 0;
+        configRUN_TIME_COUNTER_TYPE ulStatsAsPercentage;
```
Skptak, line 7278: Should this also receive an initial value?

kar-rahul-aws: ulStatsAsPercentage is assigned a value [here](https://github.com/FreeRTOS/FreeRTOS-Kernel/blob/main/tasks.c#L7335), before it is read, so it does not cause this MISRA violation even if it is not given an initial value.

chinglee-iot (approved):

shubnil (approved):

PR #842, opened 2023-10-19
Fix MISRA_C_2012 rule 7.2 violation

kar-rahul-aws:
Description
-----------
This PR adds fix for the Misra rule 7.2 violation
Misra Rule 7.2 states that
_A "u" or "U" suffix shall be applied to all integer constants that are represented in an unsigned type._

Test Steps
-----------
Run the Coverity tool to verify these errors are no longer reported.

Checklist:
----------

### DIFF
--- a/include/stack_macros.h
+++ b/include/stack_macros.h
@@ -89,7 +89,7 @@
     #define taskCHECK_FOR_STACK_OVERFLOW()                                                            \
     do {                                                                                              \
         const uint32_t * const pulStack = ( uint32_t * ) pxCurrentTCB->pxStack;                       \
-        const uint32_t ulCheckValue = ( uint32_t ) 0xa5a5a5a5;                                        \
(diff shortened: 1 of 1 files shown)


include/stack_macros.h
```diff
@@ -89,7 +89,7 @@
     #define taskCHECK_FOR_STACK_OVERFLOW()                                                            \
     do {                                                                                              \
         const uint32_t * const pulStack = ( uint32_t * ) pxCurrentTCB->pxStack;                       \
-        const uint32_t ulCheckValue = ( uint32_t ) 0xa5a5a5a5;                                        \
+        const uint32_t ulCheckValue = ( uint32_t ) 0xa5a5a5a5u;                                       \
```
moninom1, line 92: I think we are using capital u 'U' everywhere else, can we keep it same?

kar-rahul-aws: updated.

moninom1 (approved):

aggarg (approved):

PR #844, opened 2023-10-20
Fix MISRA_C_2012 rule 8.4 violation

kar-rahul-aws:
Fix MISRA C 2012 rule 8.4 violation

Description
-----------
MISRA rule 8.4 

> A compatible declaration shall be visible when an object or function with external linkage is defined.

**MISRA Rule Violation**

This rule states that if no available declaration is present for a variable in a header file, it should be either declared as static variable or an appropriate declaration be assigned.

2 errors are suppressed at [1](https://github.com/FreeRTOS/FreeRTOS-Kernel/blob/main/queue.c#L168) and [2](https://github.com/FreeRTOS/FreeRTOS-Kernel/blob/main/tasks.c#L439) by comment, because:

1. `xQueueRegistry` is utilized as an extern variable in queue unit tests.
2. A declaration in header file is not useful as the assembly code will still need to declare it separately from the port files.

Test Steps
-----------
N/A.

Checklist:
----------


tasks.c
```diff
@@ -490,7 +493,7 @@ PRIVILEGED_DATA static TaskHandle_t xIdleTaskHandles[ configNUMBER_OF_CORES ];
 /* Improve support for OpenOCD. The kernel tracks Ready tasks via priority lists.
  * For tracking the state of remote threads, OpenOCD uses uxTopUsedPriority
  * to determine the number of priority lists to read back from the remote target. */
-const volatile UBaseType_t uxTopUsedPriority = configMAX_PRIORITIES - 1U;
+static const volatile UBaseType_t uxTopUsedPriority = configMAX_PRIORITIES - 1U;
```
chinglee-iot, line 496: This one I am not so sure why we declare this as global instead of static global. Reference this thread
https://github.com/FreeRTOS/FreeRTOS-Kernel/pull/188

queue.c
```diff
@@ -166,7 +166,9 @@ typedef xQUEUE Queue_t;
  * The pcQueueName member of a structure being NULL is indicative of the
  * array position being vacant. */
     PRIVILEGED_DATA QueueRegistryItem_t xQueueRegistry[ configQUEUE_REGISTRY_SIZE ];
-
+    /* MISRA Ref 8.4.1 [Declaration shall be visible] */
```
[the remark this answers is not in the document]
kar-rahul-aws: This variable is getting used in Unit tests [here](https://github.com/FreeRTOS/FreeRTOS/blob/main/FreeRTOS/Test/CMock/queue/tracing/queue_registry_utest.c#L54), which result in CI check failure if declared static.

tasks.c
```diff
@@ -436,6 +436,9 @@ typedef tskTCB TCB_t;
 /*lint -save -e956 A manual analysis and inspection has been used to determine
  * which static variables must be declared volatile. */
 #if ( configNUMBER_OF_CORES == 1 )
+    /* MISRA Ref 8.4.1 [Declaration shall be visible] */
```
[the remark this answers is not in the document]
kar-rahul-aws: The rule is already stated [here](https://github.com/FreeRTOS/FreeRTOS-Kernel/blob/main/MISRA.md#rule-84).

queue.c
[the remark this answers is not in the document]
chinglee-iot: If that is the case, we need to update MISRA.md with 8.4.2 for different description.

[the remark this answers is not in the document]
kar-rahul-aws: updated.

queue.c
```diff
@@ -166,7 +166,9 @@ typedef xQUEUE Queue_t;
  * The pcQueueName member of a structure being NULL is indicative of the
  * array position being vacant. */
     PRIVILEGED_DATA QueueRegistryItem_t xQueueRegistry[ configQUEUE_REGISTRY_SIZE ];
-
+    /* MISRA Ref 8.4.2 [Declaration shall be visible] */
```
chinglee-iot, line 169: The annotation should be placed just above the declaration not below.

kar-rahul-aws: updated.

tasks.c
ActoryOu: Is it possible to check with @dachalco, the author of #188?
Or any other way to verify this before changing?

queue.c
```diff
@@ -165,6 +165,10 @@ typedef xQUEUE Queue_t;
 /* The queue registry is simply an array of QueueRegistryItem_t structures.
  * The pcQueueName member of a structure being NULL is indicative of the
  * array position being vacant. */
+
+/* MISRA Ref 8.4.2 [Declaration shall be visible] */
+/* More details at: https://github.com/FreeRTOS/FreeRTOS-Kernel/blob/main/MISRA.md#rule-84 */
+/* coverity[misra_c_2012_rule_8_4_violation] */
     PRIVILEGED_DATA QueueRegistryItem_t xQueueRegistry[ configQUEUE_REGISTRY_SIZE ];
```
swaldhoer, line 172: If ``xQueueRegistry`` is only needed to be visible for unit tests why not write this as

```c
#ifndef WHAT_EVER_THE_UNIT_TEST_MACRO_IS
static
#endif
PRIVILEGED_DATA QueueRegistryItem_t xQueueRegistry[ configQUEUE_REGISTRY_SIZE ];
```

then the visibility would be correct for both use cases.

My personal opinion in these *I need visibility for the unit test* cases is to rather have a ``ifdef/ifndef`` than an unnecessary visibility.

tasks.c
aggarg: Looking at the [OpenOCD code](https://sourceforge.net/p/openocd/code/ci/master/tree/src/rtos/FreeRTOS.c) mentioned in the 188 PR, it uses other symbols which are also static, so guess it should be fine.

chinglee-iot (approved):

PR #845, opened 2023-10-20
Fix MISRA_C_2012 rule 8.2 violation

kar-rahul-aws:
Fix MISRA_C_2012 rule 8.2 violation

Description
-----------

**Misra Rule 8.2**

> Function types shall be in prototype form with named parameters.

**MISRA violation**
The parameters in any function prototype should be named.  The parameter names can provide useful information regarding the function interface and a mismatch between a declaration and definition might be indicative of a programming error.
The function prototype:
```
typedef void (* crCOROUTINE_CODE)( CoRoutineHandle_t,
                                    UBaseType_t );
 ```
 
 **Fix**

-  Assign names to the parameters in the function prototype.

Test Steps
-----------
N/A.

Checklist:
----------


moninom1 (approved):

chinglee-iot (commented):
Can we also elaborate what could be the problem if parameter name is not specified in the prototype?
Suggest we still following the naming convention for the parameters.

PR #855, opened 2023-10-23
Fix MISRA_C_2012 rule 13.2 violation

kar-rahul-aws:
Fix MISRA C 2012 rule 13.2 violation

Description
-----------
**Misra Rule 13.2**

> The value of an expression and its persistent side effects shall be the same under all permitted evaluation orders.

**MISRA violation**
MISRA states that there shall be no more than one read access with volatile-qualified type within one sequence point.
The justification for the rule is:
If more than one volatile-qualified variable is read between sequence points, then it is unspecified which is read first. Reading a volatile variable is a side effect.

**Fix**
The solution is to explicitly order the reads, using a new set of static variables.

Test Steps
-----------
N/A.

Checklist:
----------


tasks.c
```diff
@@ -4477,13 +4479,15 @@ char * pcTaskGetName( TaskHandle_t xTaskToQuery ) /*lint !e971 Unqualified char
...
         /* Correct the tick count value after a period during which the tick
          * was suppressed.  Note this does *not* call the tick hook function for
          * each stepped tick. */
-        configASSERT( ( xTickCount + xTicksToJump ) <= xNextTaskUnblockTime );
+        configASSERT( ( xConstTickCount + xTicksToJump ) <= xConstNextTaskUnblockTime );
 
-        if( ( xTickCount + xTicksToJump ) == xNextTaskUnblockTime )
+        if( ( xConstTickCount + xTicksToJump ) == xConstNextTaskUnblockTime )
         {
             /* Arrange for xTickCount to reach xNextTaskUnblockTime in
```
ActoryOu, line 4492:
```suggestion
            /* Arrange for xConstTickCount to reach xNextTaskUnblockTime in
```

ActoryOu (commented):
One minor comment. Please take a look.

tasks.c
kar-rahul-aws: updated.

ActoryOu (dismissed):

ActoryOu (approved):

tasks.c
```diff
@@ -3854,7 +3856,7 @@ void vTaskSuspendAll( void )
         }
         else
         {
-            xReturn = xNextTaskUnblockTime - xTickCount;
+            xReturn = xConstNextTaskUnblockTime - xConstTickCount;
```
chinglee-iot, line 3859: Maybe xConstTickCount is enough to get ride of access multiple voilate variable in one statement. For example,
```c
xReturn = xNextTaskUnblockTime - xConstTickCount;
```

tasks.c
```diff
@@ -8366,14 +8373,14 @@ static void prvAddCurrentTaskToDelayedList( TickType_t xTicksToWait,
                 /* Wake time has overflowed.  Place this item in the overflow
                  * list. */
                 traceMOVED_TASK_TO_OVERFLOW_DELAYED_LIST();
-                vListInsert( pxOverflowDelayedTaskList, &( pxCurrentTCB->xStateListItem ) );
+                vListInsert( pxOverflowDelayedList, &( pxCurrentTCB->xStateListItem ) );
```
chinglee-iot, line 8376: `prvAddCurrentTaskToDelayedList` is called with scheduler suspended. Can we consider the following code to reduce the logic?
```
ListItem_t * pxCurrentTaskStateItem = pxCurrentTCB->xStateListItem;

vListInsert( pxDelayedTaskList, pxCurrentTaskStateItem );
vListInsert( pxOverflowDelayedList, pxCurrentTaskStateItem );
```

tasks.c
kar-rahul-aws: Updated.

tasks.c
kar-rahul-aws: This does not impact the MISRA violation, do you want this in a separate PR?

chinglee-iot: The comment above is suggestion only. The original implementation is okay to me.

chinglee-iot (approved):

tasks.c
kar-rahul-aws: Ok.

moninom1 (dismissed):

aggarg (approved):

PR #860, opened 2023-10-25
Fix MISRA C 2012 Rule 10.3 errors

chinglee-iot:
Fix MISRA C 2012 Rule 10.3 errors

Description
-----------
**MISRA C 2012 Rule 10.3**
> The value of an expression shall not be assigned to an object with a narrower essential type or of a different essential type category.

**MISRA violation**
The size of constant with suffix `UL` is different for different compiler.

For example, `sizeof( 0x01000000UL )` is 4 if using 32-bits GCC compiler. However, it is 8 if using 64-btis GCC compiler. Therefore, assigning a constant with suffix `UL` to a `uint32_t` can cause implicit conversion which may lead to unintended result.

The following is an example of the violation. Size of constant eventEVENT_BITS_CONTROL_BYTES is 8 with 64-bits GCC compiler. The statement `uxReturn &= ~eventEVENT_BITS_CONTROL_BYTES;` causes implicit conversion.
```c
#define eventEVENT_BITS_CONTROL_BYTES 0xff000000UL

UBaseType_t uxReturn;
...
uxReturn &= ~eventEVENT_BITS_CONTROL_BYTES;
```

**Fix In this PR**
* Use type casting for constant type.
* Use type casting to prevent assign value to an object with narrower essential type.

Test Steps
-----------
N/A

Checklist:
----------

- [ ] ~~I have tested my changes. No regression in existing tests.~~

Related Issue
-----------


tasks.c
```diff
@@ -7568,7 +7568,7 @@ TickType_t uxTaskResetEventItemValue( void )
             {
                 if( xClearCountOnExit != pdFALSE )
                 {
-                    pxCurrentTCB->ulNotifiedValue[ uxIndexToWaitOn ] = 0UL;
+                    pxCurrentTCB->ulNotifiedValue[ uxIndexToWaitOn ] = ( uint32_t ) 0UL;
```
moninom1, line 7571: Why not TickType_t here?

chinglee-iot: The ulNotifiedValue is of uint32_t type.
https://github.com/FreeRTOS/FreeRTOS-Kernel/blob/37678b06568fda7aaff090b77e8e34d6ab29ede4/tasks.c#L413

moninom1 (approved):

aggarg (approved):

PR #873, opened 2023-11-01
Fix ASSERT_SIDE_EFFECT for static type size check

chinglee-iot:
Fix ASSERT side effect for volatile variable

Description
-----------
Using volatile variable in assertion could result in different behavior in non-debug mode. Unused variable warning should also be considered when using assertion to check the size of static type.

In this PR:
* Remove the volatile prefix to prevent assert side effect.
* Add `( void ) xSize;` statement for unused parameter in configASSERT.

Test Steps
-----------
* Using GCC compiler with `-wunused-variable`. Define configASSERT with the following should have no error.
```
#define configASSERT( x )
```

* Assertion when static type has different size.

Checklist:
----------

Related Issue
-----------

### DIFF
--- a/event_groups.c
+++ b/event_groups.c
@@ -92,8 +92,13 @@ static BaseType_t prvTestWaitCondition( const EventBits_t uxCurrentEventBits,
             /* Sanity check that the size of the structure used to declare a
              * variable of type StaticEventGroup_t equals the size of the real
              * event group structure. */
-            volatile size_t xSize = sizeof( StaticEventGroup_t );
+            size_t xSize = sizeof( StaticEventGroup_t );
             configASSERT( xSize == sizeof( EventGroup_t ) );
+
+            /* The followwing line is added for unused parameter in configASSERT.
+             * There will be compiler warning when configASSERT is defined but the
+             * parameter is not used. For example, `#define configASSERT( x )`. */
+            ( void ) xSize;
         } /*lint !e529 xSize is referenced if configASSERT() is defined. */
         #endif /* configASSERT_DEFINED */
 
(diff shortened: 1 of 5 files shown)


urutva:
@chinglee-iot Please correct me If I'm wrong, based on https://github.com/FreeRTOS/FreeRTOS-Kernel/blob/83083a8a1371359e66af4a5aecb5e3110100f5c9/include/FreeRTOS.h#L349, wouldn't it be better to recommend not defining `configASSERT` in `FreeRTOSConfig.h` rather than defining it as `#define configASSERT( x )` which is already handled by `FreeRTOS.h`? I believe the intention behind the conditional compilation based on `configASSERT_DEFINED` is to achieve having no assertions at all in release/production build.

chinglee-iot:
> @chinglee-iot Please correct me If I'm wrong, based on
> 
> https://github.com/FreeRTOS/FreeRTOS-Kernel/blob/83083a8a1371359e66af4a5aecb5e3110100f5c9/include/FreeRTOS.h#L349
> , wouldn't it be better to recommend not defining `configASSERT` in `FreeRTOSConfig.h` rather than defining it as `#define configASSERT( x )` which is already handled by `FreeRTOS.h`? I believe the intention behind the conditional compilation based on `configASSERT_DEFINED` is to achieve having no assertions at all in release/production build.

@urutva 

Thank you for your feedback. The `( void ) xSize;` is added for backward compatibility. Existing demos won't have to change their FreeRTOSConfig.h file.  '#define configASSERT( x )` is one of the scenario. The following example is another one:
```c
/* assert.h. */
#ifdef NDEBUG           /* required by ANSI standard */
# define assert(__e) ((void)0)
#else
...
#endif

#define configASSERT( x ) assert( x )
```

I also agree with your point of view. Maybe I think too much about the volatile prefix in the static type size check. I will ask other members for their opinions and update in this thread again.

chinglee-iot:
@urutva 
This PR was created to fix the ASSERT_SIDE_EFFECT warning by removing the volatile qualifier. The volatile was add to prevent always true condition due to compare const to const. I'd like to close this PR until we have other solution for this.

PR #878, opened 2023-11-03
Suppress MISRA C:2012 rule 11.5 deviations

chinglee-iot:
Suppress MISRA C:2012 rule 11.5 deviations

Description
-----------

## MISRA deviations ##
The rule requires a pointer to void should not be converted into a pointer to object cause this may result in a pointer that is not correctly aligned, resulting in undefined behavior. Casting void * to a pointer to other type is used in the following scenarios:
* Casting the pointer to memory block returned by pvPortMalloc to other types.
* Opaque pointer for data hiding purposes
* List item owner for general use cases
* Callback function parameters for general use cases

## Suppress in this PR ##
* Elaborate the reason to suppress this rule by comment.

Test Steps
-----------
N/A

Checklist:
----------

- [ ] ~~I have tested my changes. No regression in existing tests.~~

Related Issue
-----------


MISRA.md
```diff
@@ -31,6 +31,44 @@ _Ref 8.4.1_
         a declaration in header file is not useful as the assembly code will
         still need to declare it separately.
 
+
+#### Rule 11.5
```
kar-rahul-aws, line 35: Since we are suppressing this Rule by comments, we should remove this rule from the deviations list ?

chinglee-iot: This rule will be removed from deviations list in coverity project PR.

aggarg (dismissed):

aggarg (approved):

ActoryOu (approved):

PR #900, opened 2023-11-24
Add SMP template port and example

chinglee-iot:
Add SMP template port and example

Description
-----------
The SMP template port shares of the same template port for single core. The port macros and implementation differences are separated by compile option `configNUMBER_OF_CORES`. Sharing the same port template helps user to identify the port differences between single core and SMP. A sample configuration file for SMP with minimal configuration change is also provided. The example project with cmake is updated to consume the template port and sample configuration to provide compilable project.

The configuration added for SMP is listed below
```c
/******************************************************************************/
/* Scheduling behaviour related definitions. **********************************/
/******************************************************************************/

/* Set configNUMBER_OF_CORES to greater than 1 to enable running one instance of
 * FreeRTOS kernel to schedule tasks across multiple identical processor cores. */
#define configNUMBER_OF_CORES    2

/******************************************************************************/
/* Hook and callback function related definitions. ****************************/
/******************************************************************************/

/* Set the following configUSE_* constants to 1 to include the named hook
 * functionality in the build.  Set to 0 to exclude the hook functionality from the
 * build.  The application writer is responsible for providing the hook function
 * for any set to 1.  See https://www.freertos.org/a00016.html */
#define configUSE_PASSIVE_IDLE_HOOK    0
```

As single core template port, the port is expected to be used for:
* Provide a starting point for a SMP port
* Provide a compilable project for static analysis tool. Since the sample configuration only enable a small set of features to reduce project complexity, a project for static analysis tool may use another configuration to enable more features for coverage. 

Test Steps
-----------
To build the example project for SMP template port:
```
cd examples\cmake_example
cmake -S . -B build -DFREERTOS_SMP_EXAMPLE=1
cd build
make
```
`smp_example` should be generated in the build folder.

Checklist:
----------

- [x] ~~I have tested my changes. No regression in existing tests.~~
- [x] ~~I have modified and/or added unit-tests to cover the code changes in this Pull Request.~~

Related Issue
-----------


ActoryOu (dismissed):
All current changes are good enough as SMP template. Approve in advance.
I'll suggest to have one readme document in examples/sample_configuration/smp to describe the configurations we should take care while setting SMP port.

ActoryOu (dismissed):

examples/sample_configuration/smp/FreeRTOSConfig.h
Skptak, line 1: I *really* disagree with adding this separate config file.
Doing a diff of the [sample_configuration/smp/FreeRTOSConfig.h](https://github.com/chinglee-iot/FreeRTOS-Kernel/blob/1ec40e532eb3d2bebc81539b13df112ff9587019/examples/sample_configuration/smp/FreeRTOSConfig.h#L76) and the existing [sample_configuration/FreeRTOSConfig.h](https://github.com/chinglee-iot/FreeRTOS-Kernel/blob/1ec40e532eb3d2bebc81539b13df112ff9587019/examples/sample_configuration/FreeRTOSConfig.h) the only change is the inclusion of `configNUMBER_OF_CORES 2` and `configUSE_PASSIVE_IDLE_HOOK 0`
```
# diff FreeRTOSConfig.h smp/FreeRTOSConfig.h
75a76,79
> /* Set configNUMBER_OF_CORES to greater than 1 to enable running one instance of
>  * FreeRTOS kernel to schedule tasks across multiple identical processor cores. */
> #define configNUMBER_OF_CORES                      2
>
290a295
> #define configUSE_PASSIVE_IDLE_HOOK           0
```

I feel like adding a completely different config file with just these two differences is just more confusing than adding something along the lines of this:
```
#ifndef DCMAKE_BUILD_SMP_EXAMPLE
    #error Please define in your build system if you wish to build the FreeRTOS-Kernel example project with or without SMP
#else
    #define configNUM_CORES DCMAKE_BUILD_SMP_EXAMPLE
#endif /* DCMAKE_BUILD_SMP_EXAMPLE */
```

Would be a better way to handle this. Especially seeing as the relevant code files are doing similar things to include the needed SMP code.

Skptak: My thought process with the above is that:
1. This becomes another config to maintain inside this repo.
    1. I was under the impression that the idea with adding the [CMake example](https://github.com/FreeRTOS/FreeRTOS-Kernel/tree/main/examples/cmake_example) was for an extremely barebones "Here's a rough idea of how the build system can work".
    1. Where adding this configuration is starting to break away from that.
    1. In my opinion, the difference between an SMP and non-SMP build (and demo) is probably best captured in a demo in the [FreeRTOS/FreeRTOS](https://github.com/FreeRTOS/FreeRTOS) repository.
1. I think that providing the example this way could actually be more confusing for new users?
    1. There's only a single difference between the two configs, (seeing as `configUSE_PASSIVE_IDLE_HOOK` is set to 0 in [FreeRTOS.h:476](https://github.com/FreeRTOS/FreeRTOS-Kernel/blob/a2a4485ed3050bcc2271688f0e253009be38751a/include/FreeRTOS.h#L476) when not using SMP). 
    1. I think some users may expect there to be more changes between the configs, which could lead to confusion? I know I was a little surprised to see that it was just the single difference.

jasonpcarroll: I agree that having another config is something else to maintain/check, but overall that's not a big deal if we just have automated CI checks. Could we add the CMake build for this to the CI checks if not done already? Arguably, it might be nicer for a customer to be able to just drag and drop the sample configuration for SMP or normal FreeRTOS as separate files, and would consider it more confusing to have to define something using -D than just using one file or the other.

Skptak: Thanks for the second opinion @jasonpcarroll!
In that case I'm fine with this proposal

Skptak (dismissed):

jasonpcarroll (approved):

jasonpcarroll (changes requested):
Add both cmake builds to the CI checks if this has not been done. Then I will approve.

examples/sample_configuration/smp/FreeRTOSConfig.h
chinglee-iot: @Skptak @jasonpcarroll 
Thanks for the opinion above.

My feeling about the template port is that It can be a starting point for developers who want to add a new port.
The example provides a cmake compilable project for developers. Based on this project, they can continue to implement the porting interface. It is probably not suitable to be put into FreeRTOS/Demo.

Also agree with the opinion that adding a new FreeRTOSConfig.h is confusing. I would like to suggest a compromise approach to update the FreeRTOSConfig.h with the following.
./sample_configuration/smp ( Reference file changed for complete version )
```c
#ifndef __FREERTOS_CONFIG_SMP_H__
#define __FREERTOS_CONFIG_SMP_H__

#include "../FreeRTOSConfig.h"
#define configNUMBER_OF_CORES    2
#define configUSE_PASSIVE_IDLE_HOOK    0

#endif /* __FREERTOS_CONFIG_SMP_H__ */
```
In this way, we only maintain one copy of common FreeRTOSConfig.h. Users can also know the configuration differences to enable FreeRTOS SMP support.

The CI demo build is also added. Please reference this [link](https://github.com/FreeRTOS/FreeRTOS-Kernel/actions/runs/7030308989/job/19129608729?pr=900).

kar-rahul-aws (approved):

shubnil (approved):

.github/workflows/kernel-demos.yml
```diff
@@ -111,6 +111,27 @@ jobs:
...
+          sudo apt-get -y update
+          sudo apt-get -y install build-essential
+
+      - name: Build CMake SMP Example Demo
+        shell: bash
+        working-directory: examples/cmake_example
+        run: |
+          cmake -S . -B build -DFREERTOS_SMP_EXAMPLE=1
+          cmake --build build
+
```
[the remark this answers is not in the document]
chinglee-iot: Merge the SMP example into CMake-Example job.

kar-rahul-aws (approved):

Skptak (approved):

examples/sample_configuration/smp/FreeRTOSConfig.h
Skptak, line 1: I think this is actually a great idea for how to handle this, thanks for doing this!

PR #902, opened 2023-11-24
Update coverity scan build and upload to single step

tony-josi-aws:
Description
-----------
This PR updates the coverity scan yml script to do the coverity scan build and upload to single step.

Test Steps
-----------
[Tested on fork](https://github.com/tony-josi-aws/FreeRTOS-Kernel/actions/runs/6979086011/job/18991676061)

Checklist:
----------

Related Issue
-----------

### DIFF
--- a/.github/workflows/coverity_scan.yml
+++ b/.github/workflows/coverity_scan.yml
@@ -63,18 +63,6 @@ jobs:
           cov-build --dir cov-int make -j
           tar czvf gcc_freertos_kernel_sample_build.tgz cov-int
 
-          echo "::endgroup::"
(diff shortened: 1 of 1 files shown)


Skptak (approved):

Skptak:
Either this or #903 should fix this issue
I introduced this bug when splitting the creation of the tar file and the uploading of it into separate steps. I had split these into separate steps so that if future changes were made to use the created coverity report that the creation and upload of it were seperated.

The tar file needs to be moved out of the build directory if used in a following step. Or the following step needed to move into the build directory, as is done in the first directory.

I approved this PR as this will also fix the current problem with this scan failing.

chinglee-iot (approved):

PR #924, opened 2023-12-14
Add code to allow building for x64 in MSVC

josesimoes:
Description
-----------
- Add code for x64 arch.
- Add initial value for local otherwise it won't get proper value in x64.

Test Steps
-----------
- Run MSVC solution in both x86 and x64 build.

Checklist:
----------

Related Issue
-----------
- Required for running on MSVC x64. Demo updated @ FreeRTOS/FreeRTOS#1139.


tasks.c
```diff
@@ -211,7 +211,7 @@
 
     #define taskSELECT_HIGHEST_PRIORITY_TASK()                                                  \
     do {                                                                                        \
-        UBaseType_t uxTopPriority;                                                              \
+        UBaseType_t uxTopPriority = 0;                                                          \
```
jefftenney, line 214: What do you think about moving this initialization into `portGET_HIGHEST_PRIORITY()`?  The initialization appears to be necessary only for some implementations of `portGET_HIGHEST_PRIORITY()`.  I am guessing in this case it is necessary because the x86-x64 implementations pass `&uxTopPriority` as a parameter to a function, and the compiler generates a warning without the initialization.

My thinking here is to avoid unnecessary CPU cycles in the task-switching path for most platforms.

josesimoes: Nothing against. 
This was required because when running the x64 build the upper bytes weren't set to 0 (ended up with something like `0xcccc0005` for example) which caused an exception for trying access wrong memory address.

josesimoes: Done!

portable/MSVC-MingW/portmacro.h
```diff
@@ -129,9 +129,13 @@ void vPortExitCritical( void );
...
 
 /* Store/clear the ready priorities in a bit map. */
+#if defined( __x86_64__ ) || defined( _M_X64 )
+    #define portRECORD_READY_PRIORITY( uxPriority, uxReadyPriorities )    ( uxReadyPriorities ) |= ( 1ULL << ( uxPriority ) )
+    #define portRESET_READY_PRIORITY( uxPriority, uxReadyPriorities )     ( uxReadyPriorities ) &= ~( 1ULL << ( uxPriority ) )
+#else
     #define portRECORD_READY_PRIORITY( uxPriority, uxReadyPriorities )    ( uxReadyPriorities ) |= ( 1UL << ( uxPriority ) )
     #define portRESET_READY_PRIORITY( uxPriority, uxReadyPriorities )     ( uxReadyPriorities ) &= ~( 1UL << ( uxPriority ) )
-
+#endif
```
jefftenney, line 139: Can we avoid adding conditional compilation here by using `( ( UBaseType_t ) 1 )` instead of `1UL`?

portable/MSVC-MingW/portmacro.h
```diff
@@ -143,7 +147,13 @@ void vPortExitCritical( void );
 
 /* BitScanReverse returns the bit position of the most significant '1'
  * in the word. */
+#if defined( __x86_64__ ) || defined( _M_X64 )
+        #define portGET_HIGHEST_PRIORITY( uxTopPriority, uxReadyPriorities )    uxTopPriority = 0; _BitScanReverse64( ( DWORD * ) &( uxTopPriority ), ( uxReadyPriorities ) )
```
jefftenney, line 151: Thanks for your earlier comment about setting `uxTopPriority` to 0.  The pointer casting performed here is the root cause of the issue and should be resolved directly -- ie, don't cast a uint64_t pointer as a uint32_t pointer for the call to `_BitScanReverse64()`.  Both `_BitScanReverse()` and `_BitScanReverse64()` write the result to a 32-bit integer. Something like this would resolve it:

```
#define portGET_HIGHEST_PRIORITY( uxTopPriority, uxReadyPriorities ) \
    do                                                               \
    {                                                                \
        DWORD ulTopPriority;                                         \
        _BitScanReverse64( &ulTopPriority, ( uxReadyPriorities ) );  \
        uxTopPriority = ulTopPriority;                               \
    } while( 0 )
```

Also best if you can indent the new conditionals following the pattern established above these changes.

portable/MSVC-MingW/portmacro.h
josesimoes: Done

portable/MSVC-MingW/portmacro.h
josesimoes: Done. Also, thank you for the explanation.

n9wxu (approved):

n9wxu:
@josesimoes did you forget to push your changes?  I see that you noted that you had completed the requested work, but I don't see the differences.

josesimoes:
@n9wxu oooops... you are correct, of course! 😅  embarassing...

portable/MSVC-MingW/portmacro.h
```diff
@@ -143,7 +142,18 @@ void vPortExitCritical( void );
...
+    #define portGET_HIGHEST_PRIORITY( uxTopPriority, uxReadyPriorities ) \
+        do                                                               \
+        {                                                                \
+            DWORD ulTopPriority;                                         \
+            _BitScanReverse64( &ulTopPriority, ( uxReadyPriorities ) );  \
+            uxTopPriority = ulTopPriority;                               \
+        } while( 0 )
+#else
+    #define portGET_HIGHEST_PRIORITY( uxTopPriority, uxReadyPriorities )    _BitScanReverse( ( DWORD * ) &( uxTopPriority ), ( uxReadyPriorities ) )
+#endif
```
jefftenney, line 155: @josesimoes Minor style issue - would you indent this whole block two tab's worth of spaces (should be 8 spaces I think).  That would match the existing style and current depth of the conditionals.

jefftenney (approved):

portable/MSVC-MingW/portmacro.h
josesimoes: Just committed those changes. I believe its correct. If not, let me know.

jefftenney: @josesimoes Apologies - I am just now seeing this comment.  The indenting still isn't quite right.  Please apply the attached.
[indent-suggestion.patch](https://github.com/FreeRTOS/FreeRTOS-Kernel/files/14073231/indent-suggestion.patch)

aggarg (approved):

kar-rahul-aws (approved):

PR #938, opened 2023-12-27
ARMv7-R MPU Port

Skptak:
Description
-----------
FreeRTOS Port for the [ARMv7-R Architecture](https://developer.arm.com/documentation/ddi0406/cd/?lang=en), tested on the [Cortex R4](https://developer.arm.com/Processors/Cortex-R4) and [Cortex R5](https://developer.arm.com/Processors/Cortex-R5) Processors

Test Steps
-----------

Checklist:
----------

Related Issue
-----------

### DIFF
--- a/portable/GCC/ARM_CRx_MPU/mpu_wrappers_v2_asm.S
+++ b/portable/GCC/ARM_CRx_MPU/mpu_wrappers_v2_asm.S
@@ -1,0 +1,1332 @@
+/*
+ * FreeRTOS V202112.00
+ * Copyright (C) 2020 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy of
+ * this software and associated documentation files (the "Software"), to deal in
+ * the Software without restriction, including without limitation the rights to
+ * use, copy, modify, merge, publish, distribute, sublicense, and/or sell copies of
+ * the Software, and to permit persons to whom the Software is furnished to do so,
+ * subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in all
+ * copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY, FITNESS
+ * FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR
+ * COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER
+ * IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN
+ * CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+ *
+ * https://www.FreeRTOS.org
+ * https://github.com/FreeRTOS
+ *
+ */
+
+/*---------------------------------------------------------------------------*/
+
+    .arm
+    .syntax unified
+    .section freertos_system_calls
+#define FREERTOS_ASSEMBLY
+    #include "FreeRTOSConfig.h"
+    #include "portmacro_asm.h"
+    #include "mpu_syscall_numbers.h"
+#undef FREERTOS_ASSEMBLY
+
+/* ------------------ Start of Port Specific System Calls ------------------ */
+
+/**
+ * Function: void vPortYield
+ * Inputs:      VOID
+*/
+.align 4
+.global vPortYield
+.type vPortYield, %function
+vPortYield:
+    /* Make the SVC to swap tasks */
+    SVC     #portSVC_YIELD
+    /* After yielding to another task, resume executing the calling task */
+    BX  LR
+
+
+/*-------------------------------------------------------------------------------*/
+/* vPortSystemCallExit */
+.align 4
+.global vPortSystemCallExit
+.type vPortSystemCallExit, %function
+vPortSystemCallExit:
+    /* Make the SVC to exit a system call */
+    SVC     #portSVC_SYSTEM_CALL_EXIT
+    /* After performing the requested FreeRTOS API, return to the calling task */
+    BX      LR
+
+/*-------------------------------------------------------------------------------*/
+/**
+ * Function: BaseType_t xPortIsPrivileged
+ * Inputs:      VOID
+*/
+.align 4
+.weak xPortIsPrivileged
+.type xPortIsPrivileged, %function
+xPortIsPrivileged:
+    /* Load value of SPSR into R0 */
+    MRS     R0, CPSR
+    /* Get the relevant mode bits */
+    AND     R0, R0, #0x1F
+    /* Check if we're in user mode */
+    CMP     R0, #USER_MODE
+    /* If we are in user mode set R0 to 0 for the return */
+    MOVEQ   R0, #0x0
+    /* Otherwise set R0 to 1 for the return */
+    MOVNE   R0, #0x01
+    BLX      LR
+
+/*-------------------------------------------------------------------------------*/
+/**
+ * Function: UBaseType_t ulPortCountLeadingZeros
+ * Inputs:      UBaseType_t ulBitmap
+*/
+.align 4
+.weak ulPortCountLeadingZeros
+.type ulPortCountLeadingZeros, %function
+ulPortCountLeadingZeros:
+    /* Count the leading zeros and return in R0 */
+    CLZ     R0, R0
+    BX      LR
+
+/* ------------------- End of Port Specific System Calls ------------------- */
+
+.macro processorInUserMode
+    /* Push R0 before using it */
+    PUSH    {R0}
+    /* Load value of SPSR into R0 */
+    MRS     R0, CPSR
+    /* Get the relevant mode bits */
+    AND     R0, R0, #0x1F
+    /* Determine if the CPU is in user mode */
+    CMP     R0, #USER_MODE
+    /* Restore the pushed register */
+    POP     {R0}
+.endm
+
+/*---------------------------------------------------------------------------*/
+
+.extern MPU_xTaskGetTickCountImpl
+/**
+ * Function: TickType_t MPU_xTaskGetTickCount
+ * Inputs:     void - No Inputs
+*/
+.align 4
+.global MPU_xTaskGetTickCount
+.type MPU_xTaskGetTickCount, function
+MPU_xTaskGetTickCount:
+    processorInUserMode
+    SVCEQ   #SYSTEM_CALL_xTaskGetTickCount
+    B       MPU_xTaskGetTickCountImpl
+
+/*---------------------------------------------------------------------------*/
+
+.extern MPU_uxTaskGetNumberOfTasksImpl
+/**
+ * Function: UBaseType_t MPU_uxTaskGetNumberOfTasks
+ * Inputs:     void - No Inputs
+*/
+.align 4
+.global MPU_uxTaskGetNumberOfTasks
+.type MPU_uxTaskGetNumberOfTasks, function
+MPU_uxTaskGetNumberOfTasks:
+    processorInUserMode
+    SVCEQ   #SYSTEM_CALL_uxTaskGetNumberOfTasks
+    B      MPU_uxTaskGetNumberOfTasksImpl
+
+/*---------------------------------------------------------------------------*/
+
+.extern MPU_vTaskSetTimeOutStateImpl
+/**
+ * Function: void MPU_vTaskSetTimeOutState
+ * Inputs:     TimeOut_t * const pxTimeOut
+*/
+.align 4
+.global MPU_vTaskSetTimeOutState
+.type MPU_vTaskSetTimeOutState, function
+MPU_vTaskSetTimeOutState:
+    processorInUserMode
+    SVCEQ   #SYSTEM_CALL_vTaskSetTimeOutState
+    B       MPU_vTaskSetTimeOutStateImpl
+
+/*---------------------------------------------------------------------------*/
+
+.extern MPU_xTaskCheckForTimeOutImpl
+/**
+ * Function: BaseType_t MPU_xTaskCheckForTimeOut
+ * Inputs:     TimeOut_t * const pxTimeOut
+ * Inputs:     TickType_t * const pxTicksToWait
+*/
+.align 4
+.global MPU_xTaskCheckForTimeOut
+.type MPU_xTaskCheckForTimeOut, function
+MPU_xTaskCheckForTimeOut:
+    processorInUserMode
+    SVCEQ   #SYSTEM_CALL_xTaskCheckForTimeOut
+    B       MPU_xTaskCheckForTimeOutImpl
+
+/*---------------------------------------------------------------------------*/
+
+.extern MPU_xQueueGenericSendImpl
+/**
+ * Function: BaseType_t MPU_xQueueGenericSend
+ * Inputs:     QueueHandle_t xQueue
+ * Inputs:     const void * const pvItemToQueue
+ * Inputs:     TickType_t xTicksToWait
+ * Inputs:     constBaseType_t xCopyPosition
+*/
+.align 4
+.global MPU_xQueueGenericSend
+.type MPU_xQueueGenericSend, function
+MPU_xQueueGenericSend:
+    processorInUserMode
+    SVCEQ   #SYSTEM_CALL_xQueueGenericSend
+    B       MPU_xQueueGenericSendImpl
+
+/*---------------------------------------------------------------------------*/
+
+.extern MPU_uxQueueMessagesWaitingImpl
+/**
+ * Function: UBaseType_t MPU_uxQueueMessagesWaiting
+ * Inputs:     const QueueHandle_t xQueue
+*/
+.align 4
+.global MPU_uxQueueMessagesWaiting
+.type MPU_uxQueueMessagesWaiting, function
+MPU_uxQueueMessagesWaiting:
+    processorInUserMode
+    SVCEQ   #SYSTEM_CALL_uxQueueMessagesWaiting
+    B       MPU_uxQueueMessagesWaitingImpl
+
+/*---------------------------------------------------------------------------*/
+
+.extern MPU_uxQueueSpacesAvailableImpl
+/**
+ * Function: UBaseType_t MPU_uxQueueSpacesAvailable
+ * Inputs:     const QueueHandle_t xQueue
+*/
+.align 4
+.global MPU_uxQueueSpacesAvailable
+.type MPU_uxQueueSpacesAvailable, function
+MPU_uxQueueSpacesAvailable:
+    processorInUserMode
+    SVCEQ   #SYSTEM_CALL_uxQueueSpacesAvailable
+    B       MPU_uxQueueSpacesAvailableImpl
+
+/*---------------------------------------------------------------------------*/
+
+.extern MPU_xQueueReceiveImpl
+/**
+ * Function: BaseType_t MPU_xQueueReceive
+ * Inputs:     QueueHandle_t xQueue
+ * Inputs:     void * const pvBuffer
+ * Inputs:     TickType_t xTicksToWait
+*/
+.align 4
+.global MPU_xQueueReceive
+.type MPU_xQueueReceive, function
+MPU_xQueueReceive:
+    processorInUserMode
+    SVCEQ   #SYSTEM_CALL_xQueueReceive
+    B       MPU_xQueueReceiveImpl
+
+/*---------------------------------------------------------------------------*/
+
+.extern MPU_xQueuePeekImpl
+/**
+ * Function: BaseType_t MPU_xQueuePeek
+ * Inputs:     QueueHandle_t xQueue
+ * Inputs:     void * const pvBuffer
+ * Inputs:     TickType_t xTicksToWait
+*/
+.align 4
+.global MPU_xQueuePeek
+.type MPU_xQueuePeek, function
+MPU_xQueuePeek:
+    processorInUserMode
+    SVCEQ   #SYSTEM_CALL_xQueuePeek
+    B       MPU_xQueuePeekImpl
+
+/*---------------------------------------------------------------------------*/
+
+.extern MPU_xQueueSemaphoreTakeImpl
+/**
+ * Function: BaseType_t MPU_xQueueSemaphoreTake
+ * Inputs:     QueueHandle_t xQueue
+ * Inputs:     TickType_t xTicksToWait
+*/
+.align 4
+.global MPU_xQueueSemaphoreTake
+.type MPU_xQueueSemaphoreTake, function
+MPU_xQueueSemaphoreTake:
+    processorInUserMode
+    SVCEQ   #SYSTEM_CALL_xQueueSemaphoreTake
+    B       MPU_xQueueSemaphoreTakeImpl
+
+/*---------------------------------------------------------------------------*/
+
+.extern MPU_xEventGroupWaitBitsImpl
+/**
+ * Function: EventBits_t MPU_xEventGroupWaitBitsEntry
+ * Inputs:     const xEventGroupWaitBitsParams_t * pxParams
+*/
+.align 4
+.global MPU_xEventGroupWaitBitsEntry
+.type MPU_xEventGroupWaitBitsEntry, function
+MPU_xEventGroupWaitBitsEntry:
+    processorInUserMode
+    SVCEQ   #SYSTEM_CALL_xEventGroupWaitBits
+    B       MPU_xEventGroupWaitBitsImpl
+
+/*---------------------------------------------------------------------------*/
+
+.extern MPU_xEventGroupClearBitsImpl
+/**
+ * Function: EventBits_t MPU_xEventGroupClearBits
+ * Inputs:     EventGroupHandle_t xEventGroup
+ * Inputs:     constEventBits_t uxBitsToClear
+*/
+.align 4
+.global MPU_xEventGroupClearBits
+.type MPU_xEventGroupClearBits, function
+MPU_xEventGroupClearBits:
+    processorInUserMode
+    SVCEQ   #SYSTEM_CALL_xEventGroupClearBits
+    B       MPU_xEventGroupClearBitsImpl
+
+/*---------------------------------------------------------------------------*/
+
+.extern MPU_xEventGroupSetBitsImpl
+/**
+ * Function: EventBits_t MPU_xEventGroupSetBits
+ * Inputs:     EventGroupHandle_t xEventGroup
+ * Inputs:     constEventBits_t uxBitsToSet
+*/
+.align 4
+.global MPU_xEventGroupSetBits
+.type MPU_xEventGroupSetBits, function
+MPU_xEventGroupSetBits:
+    processorInUserMode
+    SVCEQ   #SYSTEM_CALL_xEventGroupSetBits
+    B       MPU_xEventGroupSetBitsImpl
+
+/*---------------------------------------------------------------------------*/
+
+.extern MPU_xEventGroupSyncImpl
+/**
+ * Function: EventBits_t MPU_xEventGroupSync
+ * Inputs:     EventGroupHandle_t xEventGroup
+ * Inputs:     constEventBits_t uxBitsToSet
+ * Inputs:     constEventBits_t uxBitsToWaitFor
+ * Inputs:     TickType_t xTicksToWait
+*/
+.align 4
+.global MPU_xEventGroupSync
+.type MPU_xEventGroupSync, function
+MPU_xEventGroupSync:
+    processorInUserMode
+    SVCEQ   #SYSTEM_CALL_xEventGroupSync
+    B       MPU_xEventGroupSyncImpl
+
+/*---------------------------------------------------------------------------*/
+
+.extern MPU_xStreamBufferSendImpl
+/**
+ * Function: size_t MPU_xStreamBufferSend
+ * Inputs:     StreamBufferHandle_t xStreamBuffer
+ * Inputs:     const void * pvTxData
+ * Inputs:     size_t xDataLengthBytes
+ * Inputs:     TickType_t xTicksToWait
+*/
+.align 4
+.global MPU_xStreamBufferSend
+.type MPU_xStreamBufferSend, function
+MPU_xStreamBufferSend:
+    processorInUserMode
+    SVCEQ   #SYSTEM_CALL_xStreamBufferSend
+    B       MPU_xStreamBufferSendImpl
+
+/*---------------------------------------------------------------------------*/
+
+.extern MPU_xStreamBufferReceiveImpl
+/**
+ * Function: size_t MPU_xStreamBufferReceive
+ * Inputs:     StreamBufferHandle_t xStreamBuffer
+ * Inputs:     void * pvRxData
+ * Inputs:     size_t xBufferLengthBytes
+ * Inputs:     TickType_t xTicksToWait
+*/
+.align 4
+.global MPU_xStreamBufferReceive
+.type MPU_xStreamBufferReceive, function
+MPU_xStreamBufferReceive:
+    processorInUserMode
+    SVCEQ   #SYSTEM_CALL_xStreamBufferReceive
+    B       MPU_xStreamBufferReceiveImpl
+
+/*---------------------------------------------------------------------------*/
+
+.extern MPU_xStreamBufferIsFullImpl
+/**
+ * Function: BaseType_t MPU_xStreamBufferIsFull
+ * Inputs:     StreamBufferHandle_t xStreamBuffer
+*/
+.align 4
+.global MPU_xStreamBufferIsFull
+.type MPU_xStreamBufferIsFull, function
+MPU_xStreamBufferIsFull:
+    processorInUserMode
+    SVCEQ   #SYSTEM_CALL_xStreamBufferIsFull
+    B       MPU_xStreamBufferIsFullImpl
+
+/*---------------------------------------------------------------------------*/
+
+.extern MPU_xStreamBufferIsEmptyImpl
+/**
+ * Function: BaseType_t MPU_xStreamBufferIsEmpty
+ * Inputs:     StreamBufferHandle_t xStreamBuffer
+*/
+.align 4
+.global MPU_xStreamBufferIsEmpty
+.type MPU_xStreamBufferIsEmpty, function
+MPU_xStreamBufferIsEmpty:
+    processorInUserMode
+    SVCEQ   #SYSTEM_CALL_xStreamBufferIsEmpty
+    B       MPU_xStreamBufferIsEmptyImpl
+
+/*---------------------------------------------------------------------------*/
+
+.extern MPU_xStreamBufferSpacesAvailableImpl
+/**
+ * Function: size_t MPU_xStreamBufferSpacesAvailable
+ * Inputs:     StreamBufferHandle_t xStreamBuffer
+*/
+.align 4
+.global MPU_xStreamBufferSpacesAvailable
+.type MPU_xStreamBufferSpacesAvailable, function
+MPU_xStreamBufferSpacesAvailable:
+    processorInUserMode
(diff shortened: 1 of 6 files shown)


portable/GCC/ARM_CRx_MPU/mpu_wrappers_v2_asm.S
```diff
@@ -0,0 +1,1334 @@
...
+    /* Load value of SPSR into R0 */
+    MRS     R0, CPSR
+    /* Get the relevant mode bits */
+    AND     R0, R0, #0x1F
+    /* Check if we're in user mode */
+    CMP     R0, #USER_MODE
+    /* If we are in user mode set R0 to 0 for the return */
+    MOVEQ   R0, #0x0
+    /* Otherwise set R0 to 1 for the return */
+    MOVNE   R0, #0x01
```
kstribrnAmzn, line 85: Seems like it would be better to check if the R0 register is equal to the system mode (#0x1F) and return true if it is, this way all other modes are treated as user mode.

The update you would need:
```
CMP R0, #SYSTEM_MODE
MOVEQ R0, #0x1
MOVENE R0, #0x0
```

portable/GCC/ARM_CRx_MPU/mpu_wrappers_v2_asm.S
```diff
@@ -0,0 +1,1334 @@
...
+/**
+ * Function: BaseType_t xPortIsPrivileged
+ * Inputs:      VOID
+*/
+.align 4
+.weak xPortIsPrivileged
+.type xPortIsPrivileged, %function
+xPortIsPrivileged:
+    /* Load value of SPSR into R0 */
+    MRS     R0, CPSR
```
kstribrnAmzn, line 77: Why don't you need to store the R0 register value before this load?

portable/GCC/ARM_CRx_MPU/mpu_wrappers_v2_asm.S
```diff
@@ -0,0 +1,1334 @@
...
+ * Function: TickType_t MPU_xTaskGetTickCount
+ * Inputs:     void - No Inputs
+*/
+.align 4
+.global MPU_xTaskGetTickCount
+.type MPU_xTaskGetTickCount, function
+MPU_xTaskGetTickCount:
+    processorInUserMode
+    SVCEQ   #SYSTEM_CALL_xTaskGetTickCount
+    B       MPU_xTaskGetTickCountImpl
```
kstribrnAmzn, line 129: I am probably misreading this - but if the process in usermode then `processInUserMode` will set the condition status field to true which will in turn result in the `SVCEQ` call executing followed by the unconditional branch which calls the impl.

What happens here if you're in system mode? From what I can tell it wouldn't log the system call but would still take the branch. Is that the correct behavior?

portable/GCC/ARM_CRx_MPU/portASM.S
```diff
@@ -0,0 +1,624 @@
...
+
+    /* Call the interrupt handler. */
+    PUSH    {R0-R3, LR}
+    BL      vApplicationIRQHandler
+
+    /* Restore the registers */
+    POP     {R0-R3, LR}
+    ADD     SP, SP, R2
+
+    CPSID    i
```
kstribrnAmzn, line 568: Nit: Does this need to be capital `I`? It has been in the other disable/enable calls

portable/GCC/ARM_CRx_MPU/mpu_wrappers_v2_asm.S
kstribrnAmzn: Or is the goal here to treat all other modes as system modes?

portable/GCC/ARM_CRx_MPU/port.c
```diff
@@ -0,0 +1,799 @@
...
+        }
+        else
+        {
+            ulReturnValue++;
+        }
+    }
+
+    /* Shift the code by one before returning so it can be written directly
+     * into the the correct bit position of the attribute register. */
+    return ulReturnValue << 1UL;
```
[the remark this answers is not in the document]
kstribrnAmzn: I get this part after looking at the header file corresponding to the sizes. Resolving this one.

portable/GCC/ARM_CRx_MPU/mpu_wrappers_v2_asm.S
Skptak: R0 is the return register, so calling convention expects the return for this function to be in R0, link to a blog post explaining this is [here](https://community.arm.com/arm-community-blogs/b/architectures-and-processors-blog/posts/function-parameters-on-32-bit-arm)

portable/GCC/ARM_CRx_MPU/mpu_wrappers_v2_asm.S
Skptak: Yes that is correct. When the processor is in [User Mode](https://developer.arm.com/documentation/ddi0406/cb/System-Level-Architecture/The-System-Level-Programmers--Model/ARM-processor-modes-and-ARM-core-registers/ARM-processor-modes?lang=en) it will make a SVC, which will route through the [portASM.S:FreeRTOS_SVC_Handler](https://github.com/FreeRTOS/FreeRTOS-Kernel/blob/d965464f4e2a22d5bcd0d85ae63fac4356ff8b8d/portable/GCC/ARM_CRx_MPU/portASM.S#L288) to [portASM.S:svcSystemCallEnter](https://github.com/FreeRTOS/FreeRTOS-Kernel/blob/d965464f4e2a22d5bcd0d85ae63fac4356ff8b8d/portable/GCC/ARM_CRx_MPU/portASM.S#L397)

When an unprivileged task goes through [portASM.S:svcSystemCallEnter](https://github.com/FreeRTOS/FreeRTOS-Kernel/blob/d965464f4e2a22d5bcd0d85ae63fac4356ff8b8d/portable/GCC/ARM_CRx_MPU/portASM.S#L397) the stack pointer of the task is temporarily moved to [ulSystemCallStackBuffer](https://github.com/FreeRTOS/FreeRTOS-Kernel/blob/d965464f4e2a22d5bcd0d85ae63fac4356ff8b8d/portable/GCC/ARM_CRx_MPU/portmacro.h#L628), and the task's link register is set to [vPortSystemCallExit](https://github.com/FreeRTOS/FreeRTOS-Kernel/blob/d965464f4e2a22d5bcd0d85ae63fac4356ff8b8d/portable/GCC/ARM_CRx_MPU/mpu_wrappers_v2_asm.S#L61):

```
    /* Store the task's SP and LR to xSYSTEM_CALL_STACK_INFO */
    STM     R11, {R13-R14}^
    /* It's a compilation warning to use the ! for writeback, so manually move R11 */
    ADD     R11, R11, 0x8
    /* Load pulSystemCallStackPointer and pulSystemCallLinkRegister now */
    LDM     R11, {R13-R14}^
```

The handler also loads the function from the table declared in [portable/Common/mpu_wrappers_v2](https://github.com/FreeRTOS/FreeRTOS-Kernel/blob/V11.0.1/portable/Common/mpu_wrappers_v2.c#L4896):

```
    /* Load the base address of the uxSystemCallImplementations[] table into R14 */
    LDR     R14, =uxSystemCallImplementations

    /** Shift the value of R11, the SVC number, left by two to get the jump offset
     * Add this offset to R14, which holds the jump table address. This is the address
     * of the SVC that the relevant function is trying to complete.
     * Now when the Link Register is loaded as the Program Counter at the end of this
     * handler, the caller will immediately execute the requested function */
    LDR     R14, [R14, R11, lsl #2]
```

Lastly the task is set to operate in SYSTEM_MODE for the duration of the [FreeRTOS System Call](https://github.com/FreeRTOS/FreeRTOS-Kernel/blob/d965464f4e2a22d5bcd0d85ae63fac4356ff8b8d/portable/GCC/ARM_CRx_MPU/portASM.S#L432C1-L439C25):

```
    /* Swap the SPSR to SYS_MODE for the System Call. Move SPSR into R12 */
    MRS     R12, SPSR
    /* Set the MODE bits to SYS_MODE */
    ORR     R12, R12, #SYS_MODE
    /* Assign the new value to SPSR */
    MSR     SPSR_cxsf, R12
    /* Leave through the SVC Exit */
    B   SVC_Handler_Exit
```

This way if an unprivileged task makes a FreeRTOS System Call all information related to that call is kept in privileged data. Then at the end of the system call the function will pop the LR off and exit through [vPortSystemCallExit](https://github.com/FreeRTOS/FreeRTOS-Kernel/blob/d965464f4e2a22d5bcd0d85ae63fac4356ff8b8d/portable/GCC/ARM_CRx_MPU/mpu_wrappers_v2_asm.S#L61). 

If a caller is already in a privileged operating mode they will fail this check and branch directly to the function, as they are already privileged and have the ability to execute the FreeRTOS-Kernel functions placed in [PRIVILEGED_FUNCTIONS](https://github.com/FreeRTOS/FreeRTOS-Kernel/blob/V11.0.1/include/mpu_wrappers.h#L270)

portable/GCC/ARM_CRx_MPU/port.c
```diff
@@ -0,0 +1,799 @@
...
+{
+    /** Setup the initial context of the task. The context is set exactly as
+     * expected by the portRESTORE_CONTEXT() macro. */
+    UBaseType_t ulContextIndex = MAX_CONTEXT_SIZE - 1U;
+
+    if( xRunPrivileged == pdTRUE )
+    {
+        /* Current Program Status and Control Register */
+        xMPUSettings->ulTaskFlags |= portTASK_IS_PRIVILEGED_FLAG;
+        xMPUSettings->ulTaskFlags |= 0x1F000000;
```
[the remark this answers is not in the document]
Skptak: [ulContext](https://github.com/FreeRTOS/FreeRTOS-Kernel/blob/d965464f4e2a22d5bcd0d85ae63fac4356ff8b8d/portable/GCC/ARM_CRx_MPU/portmacro_asm.h#L501) holds the saved context of a FreeRTOS-Task when it is moved from the running to non-running state.
```
/** @brief Mark if the Floating Point Registers will be saved.
 * @ingroup Task Context
 * @note When using the FPU, we must save additional registers into the task's context
 * These consist of the Floating Point Status and Control Register (FPSCR),
 * As well as the Floating Point Registers (FPRs)
 * Task Context which is stored in ulContext in order, consists of:
 * ulContext[ 0 ]:      Critical Nesting Count: ulCriticalNesting
 * ulContext[ 1 ]:      Floating Point Status and Control Register
 * ulContext[ 2 - 33 ]: Floating Point Registers: S0-S31
 * ulContext[ 34 - 46 ]:  General Purpose Registers: R0-R12
 * ulContext[ 47 ]:     Stack Pointer
 * ulContext[ 48 ]:     Link Register
 * ulContext[ 49 ]:     Program Counter
 * ulContext[ 50 ]:     Current Program Status and Control Register
 * ulContext[ 51 ]:     Supervisor Mode SRS PC Scratch Space
 * ulContext[ 52 ]:     Supervisor Mode SRS CPSR Scratch Space
 */
 ```
While [ulContext](https://github.com/FreeRTOS/FreeRTOS-Kernel/blob/d965464f4e2a22d5bcd0d85ae63fac4356ff8b8d/portable/GCC/ARM_CRx_MPU/portmacro_asm.h#L501) holds the [CPSR](https://developer.arm.com/documentation/ddi0363/e/programmer-s-model/program-status-registers), ulTaskFlags is used for different operations by the FreeRTOS-Kernel.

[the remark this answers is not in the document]
Comment (Skptak) [reply portable/GCC/ARM_CRx_MPU/port.c:110]: Believe you are correct here? I think I had added the actual system mode flag to this for functionality I didn't add, or have since removed?

File: portable/GCC/ARM_CRx_MPU/port.c [hunk 7]
```diff
@@ -0,0 +1,799 @@
...
+    const uint32_t ulRegionLength,
+    const uint32_t ulAccessRequested
+)
+{
+    BaseType_t xAccessGranted;
+    uint32_t ulRegionEnd = ulRegionStart + ulRegionLength;
+
+    /* Convert the MPU Region Size value to the actual size */
+    uint32_t ulTaskRegionLength = 1 << ( ( xTaskMPURegion->ulRegionSize >> 1 ) + 1U );
+    // uint32_t ulTaskRegionLength = 2 << ( xTaskMPURegion->ulRegionSize >> 1 );
```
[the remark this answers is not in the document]
Skptak: Actual should be the other way around, the commented out version is the "slightly more efficient version"

portable/GCC/ARM_CRx_MPU/portASM.S
```diff
@@ -0,0 +1,624 @@
...
+        /* Load values of struct MPU_REGION_REGISTERS into R2-R4 */
+        LDMIA   R1!, {R2-R4}
+        /* Load the values set in xMPU_REGION_REGISTERS
+         * R2 Will hold ulRegionSize
+         * R3 will hold ulRegionAttribute
+         * R4 will hold ulRegionBaseAddress
+         * R5 will hold the MPU Region number  */
+
+        /* Select the MPU Region using R5 */
+        MCR     p15, #0, R5, c6, c2, #0
```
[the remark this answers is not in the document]
Skptak: It's used for a bunch of things, I'm just going to link the ARM manual page about it, which can be found [here](https://developer.arm.com/documentation/ddi0363/e/system-control-coprocessor)

portable/GCC/ARM_CRx_MPU/portASM.S
Skptak: I don't think so, but thanks for pointing this out. I intentionally was trying to use capitals for ASM work.
Can you please add it as a suggestion to ensure this gets fixed?

.github/workflows/ci.yml
```diff
@@ -15,6 +15,15 @@ jobs:
         with:
           exclude-dirs: portable
 
+  port-formatting:
```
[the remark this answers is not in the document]
kstribrnAmzn: >The use of additional ASM files allows this port to easily be formatted since there is no inline ASM in the .c and .h files

I'm a big fan of this too. Keeps the assembly centrally located.

.github/workflows/kernel-demos.yml
```diff
@@ -154,8 +159,8 @@ jobs:
       - name: Checkout the FreeRTOS/FreeRTOS Repository
         uses: actions/checkout@v3
         with:
-          ref: main
-          repository: FreeRTOS/FreeRTOS
+          ref: ARM_CRx_MPU
+          repository: skptak/FreeRTOS
```
kstribrnAmzn, line 163: Remember to switch this

kstribrnAmzn (changes requested):

portable/GCC/ARM_CRx_MPU/mpu_wrappers_v2_asm.h
```diff
@@ -0,0 +1,514 @@
...
+} xTimerGenericCommandParams_t;
+
+typedef struct xEventGroupWaitBitsParams
+{
+    EventGroupHandle_t xEventGroup;
+    EventBits_t uxBitsToWaitFor;
+    BaseType_t xClearOnExit;
+    BaseType_t xWaitForAllBits;
+    TickType_t xTicksToWait;
+} xEventGroupWaitBitsParams_t;
```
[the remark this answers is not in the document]
kstribrnAmzn: Thanks for the explanation. Makes sense that this is a by product of declaring the functions in here.

portable/GCC/ARM_CRx_MPU/mpu_wrappers_v2_asm.S
kstribrnAmzn: Thank you for the detailed explanation here.

portable/GCC/ARM_CRx_MPU/portASM.S
Skptak: Changed this to a capital

.github/workflows/kernel-demos.yml
Skptak: I'll remove the build check from this PR and we can add it in after the [FreeRTOS/FreeRTOS PR](https://github.com/FreeRTOS/FreeRTOS/pull/1149) gets merged

kstribrnAmzn (approved):

portable/GCC/ARM_CRx_MPU/mpu_wrappers_v2_asm.S
```diff
@@ -0,0 +1,1333 @@
...
+    CLZ     R0, R0
+    BX      LR
+
+/* ------------------- End of Port Specific System Calls ------------------- */
+
+.macro processorInUserMode
+    /* Push R0 before using it */
+    PUSH    {R0}
+    /* Load value of SPSR into R0 */
+    MRS     R0, CPSR
```
joshzarr, line 106: Typo?
Difference between comment and code

portable/GCC/ARM_CRx_MPU/mpu_wrappers_v2_asm.S
```diff
@@ -0,0 +1,1333 @@
...
+    .extern MPU_uxEventGroupGetNumberImpl
+    /**
+     * Function: UBaseType_t MPU_uxEventGroupGetNumber
+     * Inputs:     void * xEventGroup
+    */
+    .align 4
+    .global MPU_uxEventGroupGetNumber
+    .type MPU_uxEventGroupGetNumber, function
+    MPU_uxEventGroupGetNumber:
+    processorInUserMode
```
joshzarr, line 619: Missing indent

portable/GCC/ARM_CRx_MPU/portmacro_asm.h
```diff
@@ -0,0 +1,515 @@
...
+
+/** @brief Flag uses to mark that a FreeRTOS Task is privileged
+ * @ingroup Port Privilege
+ */
+#define portTASK_IS_PRIVILEGED_FLAG      ( 1UL << 1UL )
+
+/** @brief SVC Number to use when requesting a context swap.
+ * @ingroup Scheduler
+ */
+#define portSVC_YIELD                    0x0100
```
joshzarr, line 119: Its probably worth at least leaving a comment that these values cannot clash with the system call numbers defined in mpu_syscall_numbers.h

portable/GCC/ARM_CRx_MPU/portmacro_asm.h
```diff
@@ -0,0 +1,515 @@
...
+ *
+ * Additional information related to the GPRs:
+ * https://developer.arm.com/documentation/ddi0406/cb/System-Level-Architecture/The-System-Level-Programmers--Model/ARM-processor-modes-and-ARM-core-registers/ARM-core-registers?lang=en
+ *
+ */
+
+/** @brief The length in ulContext for the General Purpose Registers in bytes
+ * @note There are 13 GPRs, R0-R12 each is 32 bits, so 13 registers * 4 Bytes each
+ */
+#define portGPR_LENGTH                       ( 15U * 4U )
```
joshzarr, line 458: Mismatch between the #define and the comment (13 vs 15). Same below for portREGISTER_CONTEXT_LENGTH

portable/GCC/ARM_CRx_MPU/portmacro.h
```diff
@@ -0,0 +1,690 @@
...
+    #define portGET_HIGHEST_PRIORITY( uxTopPriority, uxTopReadyPriority ) \
+        uxTopPriority = ( 31UL - ulPortCountLeadingZeros( ( uxTopReadyPriority ) ) )
+
+#endif /* configUSE_PORT_OPTIMISED_TASK_SELECTION */
+
+#ifndef configSETUP_TICK_INTERRUPT
+    #error configSETUP_TICK_INTERRUPT() must be defined in FreeRTOSConfig.h to call the function that sets up the tick interrupt.
+#endif /* configSETUP_TICK_INTERRUPT */
+
+#ifndef configCLEAR_TICK_INTERRUPT
```
joshzarr, line 149: I believe configCLEAR_TICK_INTERRUPT and configEOI_ADDRESS can be removed.

portable/GCC/ARM_CRx_MPU/portmacro.h
```diff
@@ -0,0 +1,690 @@
...
+    #error configCLEAR_TICK_INTERRUPT must be defined in FreeRTOSConfig.h to clear which ever interrupt was used to generate the tick interrupt.
+#endif /* configCLEAR_TICK_INTERRUPT */
+
+#ifndef configEOI_ADDRESS
+    #error Did not define a memrory address to mark the end of an Interrupt
+#endif /* configEOI_ADDRESS */
+
+#ifdef configUSE_TICKLESS_IDLE
+    #if( configUSE_TICKLESS_IDLE != 0 )
+        #error This port does not yet support tickless idle
```
joshzarr, line 159: Minor nitpick, i would remove the word 'yet'

portable/GCC/ARM_CRx_MPU/portmacro.h
```diff
@@ -0,0 +1,690 @@
...
+ * @note This must be installed as the handler for whichever peripheral is used
+ * to generate the RTOS tick if using a VIM. */
+
+/** @brief Decleration of the FreeRTOS Tick Handler.
+ *
+ * @ingroup Scheduler
+ *
+ * @note This must be installed as the handler for whichever peripheral is used
+ * to generate the RTOS tick if using a VIM. */
+void FreeRTOS_Tick_Handler( void ) __attribute__( ( weak, interrupt( "IRQ" ) ) );
```
joshzarr, line 314: I believe FreeRTOS_Tick_Handler() and vPortYieldWithinAPI() can be removed

portable/GCC/ARM_CRx_MPU/portmacro_asm.h
```diff
@@ -0,0 +1,515 @@
...
+ * BE CAREFUL DISABLING THIS: Certain APIs will try and optimize themselves
+ * by using the FPRs. If the FPU context is not saved and this happens it could be
+ * exceedingly difficult to debug why a strcpy() or other similar function
+ * seems to randomly fail.
+ */
+#ifndef configENABLE_FPU
+    #define configENABLE_FPU 0
+#endif
+
+/** @brief Mark if the Floating Point Registers will be saved.
```
joshzarr, line 477: Just as a suggestion, do you think it would add clarity if you had two version of this table, one which showed the packing when the FPU is used, and one without?
Or some other way to indicate how ulContext[] is sized/used based on the FPU configuration.

portable/GCC/ARM_CRx_MPU/portmacro_asm.h
```diff
@@ -0,0 +1,515 @@
...
+ * ulContext[ 47 ]:     Stack Pointer
+ * ulContext[ 48 ]:     Link Register
+ * ulContext[ 49 ]:     Program Counter
+ * ulContext[ 50 ]:     Current Program Status and Control Register
+ */
+#define portENABLE_FPU configENABLE_FPU
+
+#if( portENABLE_FPU == 1 )
+    /** @brief Length of a Task's Register Context when using an FPU. */
+    #define MAX_CONTEXT_SIZE 50U
```
joshzarr, line 496: Looks like there is a mismatch between the table above and the MAX_CONTEXT_SIZE

portable/GCC/ARM_CRx_MPU/mpu_wrappers_v2_asm.h
```diff
@@ -0,0 +1,514 @@
+/*
+ * FreeRTOS Kernel <DEVELOPMENT BRANCH>
+ * Copyright (C) 2021 Amazon.com, Inc. or its affiliates.  All Rights Reserved.
```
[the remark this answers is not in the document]
Skptak: I believe that we intentionally set new files to these dates? I can change it if we want to though

portable/GCC/ARM_CRx_MPU/portmacro.h
Skptak:
```suggestion
        #error This port does not support tickless idle
```

portable/GCC/ARM_CRx_MPU/portmacro_asm.h
Skptak: I think it would!

The only reason I did it this way was so that if you hover over the variable in Visual Studio Code you then see that table

I've got no issue with adding the second one, I'll think of a way to reformat it to make that work

portable/GCC/ARM_CRx_MPU/port.c
```diff
@@ -0,0 +1,799 @@
...
+/** @brief Variable to track whether or not the scheduler has been started.
+ * @ingroup Scheduler
+ * @note This variable is set to pdTRUE when the scheduler is started.
+ */
+PRIVILEGED_DATA static BaseType_t xSchedulerRunning = pdFALSE;
+
+/** @brief Used in portASM.S's IRQ Handler to clear an interrupt.
+ * @ingroup Interrupt Management
+ */
+PRIVILEGED_DATA volatile uint32_t ulICCEOIR = configEOI_ADDRESS;
```
joshzarr, line 78: This can be removed

portable/GCC/ARM_CRx_MPU/port.c
```diff
@@ -0,0 +1,799 @@
...
+        /* This function is called automatically when the task is created - in
+         * which case the stack region parameters will be valid. At all other
+         * times the stack parameters will not be valid and it is assumed that the
+         * stack region has already been configured. */
+
+        if( ulStackDepth > 0 )
+        {
+            uint32_t ulSmallestRegion = prvGetMPURegionSizeSetting( ulStackDepth * 0x4 );
+            /* Define the region that allows access to the stack. */
+            xMPUSettings->xRegion[ 0 ].ulRegionBaseAddress = ( uint32_t ) pxBottomOfStack;
```
joshzarr, line 348: It looks like there is a mismatch where xRegion[ 0 ] is being used as the stack MPU region where portmacro_asm.h states this should be the first user configurable region. This region ordering could allow the per task user defined regions to override stack protections.

portable/GCC/ARM_CRx_MPU/port.c
```diff
@@ -0,0 +1,799 @@
...
+    /* Load the context of the first task, starting the FreeRTOS-Scheduler's control. */
+    vPortStartFirstTask();
+
+    /* Will only get here if vTaskStartScheduler() was called with the CPU in
+     * a non-privileged mode or the binary point register was not set to its lowest
+     * possible value. prvTaskExitError() is referenced to prevent a compiler
+     * warning about it being defined but not referenced in the case that the user
+     * defines their own exit address. */
+    ( void ) prvTaskExitError();
+    return 0;
```
joshzarr, line 659: Minor nitpick, should return pdFALSE to match the comment above

portable/GCC/ARM_CRx_MPU/port.c
```diff
@@ -0,0 +1,799 @@
...
+    {
+        xAccessGranted = pdFALSE;
+    }
+
+    return xAccessGranted;
+}
+
+/* ------------------------------------------------------------------------- */
+
+/* PRIVILEGED_FUNCTION */ BaseType_t xPortIsAuthorizedToAccessBuffer(
```
joshzarr, line 554: There are several functions with the  PRIVILEGED_FUNCTION decorator commented out. Should they in fact be in the privileged region? If not, can this comment be deleted

portable/GCC/ARM_CRx_MPU/portASM.S
```diff
@@ -0,0 +1,580 @@
...
+#define FREERTOS_ASSEMBLY
+    #include "portmacro_asm.h"
+    #include "mpu_syscall_numbers.h"
+#undef FREERTOS_ASSEMBLY
+
+    /* External FreeRTOS-Kernel Variables */
+    .extern pxCurrentTCB
+    .extern uxSystemCallImplementations
+    .extern ulPortInterruptNesting
+    .extern ulICCEOIR
```
joshzarr, line 44: Can be removed

portable/GCC/ARM_CRx_MPU/mpu_wrappers_v2_asm.S
```diff
@@ -0,0 +1,1333 @@
...
+    .arm
+    .syntax unified
+    .section freertos_system_calls
+#define FREERTOS_ASSEMBLY
+    #include "FreeRTOSConfig.h"
+    #include "portmacro_asm.h"
+    #include "mpu_syscall_numbers.h"
+#undef FREERTOS_ASSEMBLY
+
+/* ------------------ Start of Port Specific System Calls ------------------ */
```
joshzarr, line 40: It looks like the majority of the functionality provided by this file (and associated header) are to support the mpu v2 wrappers. However, there are a few functions in this file which seem like they would be better placed in portASM.s such as vPortYied, vPortSystemCallExit, xPortIsPrivileged, ulPortCountLeadingZeros for consistency.

portable/GCC/ARM_CRx_MPU/port.c
```diff
@@ -0,0 +1,799 @@
...
+ */
+PRIVILEGED_FUNCTION static BaseType_t prvTaskCanAccessRegion(
+    const xMPU_REGION_REGISTERS * xTaskMPURegion,
+    const uint32_t ulRegionStart,
+    const uint32_t ulRegionLength,
+    const uint32_t ulAccessRequested
+)
+{
+    BaseType_t xAccessGranted;
+    uint32_t ulRegionEnd = ulRegionStart + ulRegionLength;
```
joshzarr, line 515: Looks like we can potentially wrap around memory here, are there any checks at higher levels to ensure this condition wont happen? If not, might be best to add them here

portable/GCC/ARM_CRx_MPU/port.c
```diff
@@ -0,0 +1,799 @@
...
+    /* Convert the MPU Region Size value to the actual size */
+    uint32_t ulTaskRegionLength = 1 << ( ( xTaskMPURegion->ulRegionSize >> 1 ) + 1U );
+    // uint32_t ulTaskRegionLength = 2 << ( xTaskMPURegion->ulRegionSize >> 1 );
+    uint32_t ulTaskRegionEnd = xTaskMPURegion->ulRegionBaseAddress + ulTaskRegionLength;
+    if( ( ulRegionStart >= xTaskMPURegion->ulRegionBaseAddress ) &&
+        ( ulRegionEnd <= ulTaskRegionEnd ) )
+    {
+        /* Unprivileged read is MPU Ctrl Access Bit Value bX1X */
+        if( ( tskMPU_READ_PERMISSION == ulAccessRequested ) &&
+            ( ( portMPU_PRIV_RW_USER_RO_NOEXEC ) &xTaskMPURegion->ulRegionAttribute ) )
```
joshzarr, line 526: Please add the missing space between the bitwise operands.

portable/GCC/ARM_CRx_MPU/portASM.S
```diff
@@ -0,0 +1,580 @@
...
+    /* Push the 32 Floating Point Registers (FPRs) onto the stack */
+    VSTM    LR!, { D0-D15 }
+#endif /* ( portENABLE_FPU == 1 ) */
+
+    /* Restore the saved register */
+    POP     { R0 }
+    /* Save the pre-exception Registers, Stack Pointer (SP), and LR */
+    STM     LR, { R0-R14 }^
+    /* Increment the LR after the popped registers */
+    ADD     LR, LR, #portGPR_LENGTH
```
joshzarr, line 95: Can you expand the comment here to explain why this is required? (I think its touched upon in functions further below).

Skptak: Comment adjusted to be more clear

portable/GCC/ARM_CRx_MPU/port.c
Skptak: Wrap around check added by ensuring the end address is greater than the start address of the region

portable/GCC/ARM_CRx_MPU/mpu_wrappers_v2_asm.S
Skptak: The portASM.S is marked  as PRIVILEGED_FUNCTIONS to ensure that they placed in the correct section of the linker script

Skptak: Due to this any system calls need to be placed into mpu_wrappers_v2.S to ensure they are accessible to unprivileged tasks

portable/GCC/ARM_CRx_MPU/portASM.S
```diff
@@ -0,0 +1,580 @@
...
+    /* Get the starting address for FreeRTOS System Calls */
+    LDR     R12, =__syscalls_flash_start__
+    /* Subtract the start point from the Program Counter of the caller */
+    SUB     R11, LR, R12
+    /* Now check if it is less than the length of the section */
+    LDR     R12, =__syscalls_flash_length__
+    /* Check if an SVC was raised after the end of FreeRTOS System Calls */
+    CMP     R11, R12
+    /* If the SVC was raised from outside FreeRTOS System Calls exit now */
+    BGE     SVC_Handler_Exit
```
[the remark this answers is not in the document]
Skptak: Yes, there is no error code for if the check fails. It is not possible to return an error code for a function like `vTaskDelay()` without an API change.

portable/GCC/ARM_CRx_MPU/port.c
Skptak: This comment is here to make it clear that the function is in privileged functions.

The declarations have the attributes that place them into PRIVILEGED_FUNCTIONS. This is the standard used in this repository.

portable/GCC/ARM_CRx_MPU/port.c
Skptak: Return value adjusted

portable/GCC/ARM_CRx_MPU/port.c
```diff
@@ -0,0 +1,799 @@
...
+/** @brief Interrupt nesting depth, used to count the number of interrupts to unwind.
+ * @ingroup Interrupt Management
+ */
+PRIVILEGED_DATA volatile uint32_t ulPortInterruptNesting = 0UL;
+
+/** @brief Variable to track whether or not the scheduler has been started.
+ * @ingroup Scheduler
+ * @note This variable is set to pdTRUE when the scheduler is started.
+ */
+PRIVILEGED_DATA static BaseType_t xSchedulerRunning = pdFALSE;
```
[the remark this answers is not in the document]
Skptak: This variable follows the precedent set in the existing MPU ports, such as the [CM3_MPU](https://github.com/FreeRTOS/FreeRTOS-Kernel/blob/V11.0.1/portable/GCC/ARM_CM3_MPU/port.c#L272) and [CM33_MPU](https://github.com/FreeRTOS/FreeRTOS-Kernel/blob/V11.0.1/portable/GCC/ARM_CM33/non_secure/port.c#L504)

portable/GCC/ARM_CRx_MPU/port.c
Skptak: Region number fixed, thank you for catching this

portable/GCC/ARM_CRx_MPU/port.c
Skptak: Removed

portable/GCC/ARM_CRx_MPU/portmacro_asm.h
Skptak: Value increased

portable/GCC/ARM_CRx_MPU/portmacro_asm.h
Skptak: Section has been re-formatted

portable/GCC/ARM_CRx_MPU/mpu_wrappers_v2_asm.S
Skptak: Comment has been corrected

portable/GCC/ARM_CRx_MPU/port.c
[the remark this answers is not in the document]
joshzarr: Thank you for renaming this variable. It should at least add clarity that it is distinct from the kernel state managed in tasks.c

portable/GCC/ARM_CRx_MPU/portmacro.h
Skptak: Removed both of these

joshzarr (dismissed):
Looks like all of my feedback has been addressed.
Verified the MPU demo runs on my TI Hercules HDK as expected.
Marking as approved

kstribrnAmzn (dismissed):

kstribrnAmzn (approved):

aggarg (dismissed):

aggarg (approved):

PR #950, opened 2024-01-12
Posix port - set name for threads

Mixaill:
Description
-----------

Pass the FreeRTOS task name to pthread via `pthread_setname_np()` for prettier output in debuggers. 

Tested on Linux, may need some fixes for macOS / *BSD

image

Test Steps
-----------
build any FreeRTOS-based app with posix port and run with attached debug

Checklist:
----------

-----------

### DIFF
--- a/portable/ThirdParty/GCC/Posix/port.c
+++ b/portable/ThirdParty/GCC/Posix/port.c
@@ -51,6 +51,10 @@
 *----------------------------------------------------------*/
 #include "portmacro.h"
 
+#ifdef __linux__
+    #define __USE_GNU
+#endif
+
 #include <errno.h>
 #include <pthread.h>
 #include <signal.h>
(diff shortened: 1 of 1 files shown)


Skptak:
Hey @Mixaill thanks for raising this PR!
This does seem useful for debugging, and it seems like a good idea to add this! However, due to a difference in the API for `pthread_setname_np()` between MacOS and Linux this PR would break MacOS support.

The MacOS version of the function, [pthread_setname_np(const char*)](https://opensource.apple.com/source/libpthread/libpthread-301.30.1/pthread/pthread.h.auto.html), doesn't take in a handler to the thread to name. While the Linux verison of the function, [pthread_setname_np(pthread_t thread, const char *name)](https://man7.org/linux/man-pages/man3/pthread_setname_np.3.html), takes in this second argument. This is talked about a bit more in this [Github issue](https://github.com/dotnet/runtime/issues/47087). As the POSIX port is meant to be available for MacOS and Linux I'd like this to be something that works and compiles for both. 

Maybe an "easy" way to handle this would be to make a wrapper function for `pthread_setname_np()`? Something like `prvPortSetCurrentThreadName(char * pxThreadName)`, which could then have the necessary check for if the target is apple or linux?

Mixaill:
@Skptak  added support for macOS

jasonpcarroll (approved):

PR #953, opened 2024-01-14
Make configSUPPORT_STATIC_ALLOCATION==1 an error for MPU ports

IsaacDynamo:
Make configSUPPORT_STATIC_ALLOCATION==1 an error for MPU ports

Description
-----------
The `vApplicationGet*TaskMemory()` functions are not generated for MPU ports even when `configSUPPORT_STATIC_ALLOCATION` is set.  This PR makes setting `configSUPPORT_STATIC_ALLOCATION` for MPU ports an error.  Copied explanation from https://github.com/FreeRTOS/FreeRTOS-Kernel/pull/790#discussion_r1329997898

Checklist:
----------

### DIFF
--- a/tasks.c
+++ b/tasks.c
@@ -45,7 +45,7 @@
  * reason is that the stack alignment requirements vary for different
  * architectures.*/
 #if ( ( configSUPPORT_STATIC_ALLOCATION == 1 ) && ( configKERNEL_PROVIDED_STATIC_MEMORY == 1 ) && ( portUSING_MPU_WRAPPERS != 0 ) )
-#error configKERNEL_PROVIDED_STATIC_MEMORY cannot be set to 1 when using an MPU port. The vApplicationGet*TaskMemory() functions must be provided manually.
+    #error configKERNEL_PROVIDED_STATIC_MEMORY cannot be set to 1 when using an MPU port. The vApplicationGet*TaskMemory() functions must be provided manually.
 #endif
 
 /* The MPU ports require MPU_WRAPPERS_INCLUDED_FROM_API_FILE to be defined
--- a/tasks.c
+++ b/tasks.c
@@ -41,6 +41,13 @@
 #include "timers.h"
 #include "stack_macros.h"
 
+/* The default definitions are only available for non-MPU ports. The
+ * reason is that the stack alignment requirements vary for different
+ * architectures.*/
+#if ( ( configSUPPORT_STATIC_ALLOCATION == 1 ) && ( configKERNEL_PROVIDED_STATIC_MEMORY == 1 ) && ( portUSING_MPU_WRAPPERS != 0 ) )
+#error configKERNEL_PROVIDED_STATIC_MEMORY cannot be set to 1 when using an MPU port. The vApplicationGet*TaskMemory() functions must be provided manually.
+#endif
+
 /* The MPU ports require MPU_WRAPPERS_INCLUDED_FROM_API_FILE to be defined
  * for the header files above, but not in this file, in order to generate the
  * correct privileged Vs unprivileged linkage and placement. */


Skptak:
Hey there and thanks for raising this PR!
Setting `configSUPPORT_STATIC_ALLOCATION` to 1 when using an MPU port is allowed, the end user just needs to supply the [vApplicationGetIdleTaskMemory](https://github.com/FreeRTOS/FreeRTOS-Kernel/blob/V11.0.1/tasks.c#L8645) and [vApplicationGetTimerTaskMemory](https://github.com/FreeRTOS/FreeRTOS-Kernel/blob/V11.0.1/tasks.c#L8686) functions. There is a statically allocated MPU demo that shows how to do this at [FreeRTOS/Demo/CORTEX_MPU_Static_Simulator_Keil_GCC/main.c](https://github.com/FreeRTOS/FreeRTOS/blob/154dee1c86195d88c49ba5c58c8725c16541cc16/FreeRTOS/Demo/CORTEX_MPU_Static_Simulator_Keil_GCC/main.c#L1083).

However! As you've pointed out, using an MPU port with `configKERNEL_PROVIDED_STATIC_MEMORY` set to 1 could lead to data aborts, or undefined behaviour, as the stacks and TCBs provided are not MPU Region aligned. As such I would remove the check for `( configSUPPORT_STATIC_ALLOCATION == 1 )` from your config check, as it is not the actual problem with the kernel provided functions.

IsaacDynamo:
> Hey there and thanks for raising this PR! Setting `configSUPPORT_STATIC_ALLOCATION` to 1 when using an MPU port is allowed, the end user just needs to supply the [vApplicationGetIdleTaskMemory](https://github.com/FreeRTOS/FreeRTOS-Kernel/blob/V11.0.1/tasks.c#L8645) and [vApplicationGetTimerTaskMemory](https://github.com/FreeRTOS/FreeRTOS-Kernel/blob/V11.0.1/tasks.c#L8686) functions. There is a statically allocated MPU demo that shows how to do this at [FreeRTOS/Demo/CORTEX_MPU_Static_Simulator_Keil_GCC/main.c](https://github.com/FreeRTOS/FreeRTOS/blob/154dee1c86195d88c49ba5c58c8725c16541cc16/FreeRTOS/Demo/CORTEX_MPU_Static_Simulator_Keil_GCC/main.c#L1083).

I agree.

> However! As you've pointed out, using an MPU port with `configKERNEL_PROVIDED_STATIC_MEMORY` set to 1 could lead to data aborts, or undefined behaviour, as the stacks and TCBs provided are not MPU Region aligned. 

This should not happen because the generated default implementations are already conditional on `#if ( ( configSUPPORT_STATIC_ALLOCATION == 1 ) && ( configKERNEL_PROVIDED_STATIC_MEMORY == 1 ) && ( portUSING_MPU_WRAPPERS == 0 ) )`, [source](https://github.com/FreeRTOS/FreeRTOS-Kernel/blob/cb900a2b7dc643130134eb5db95da037f7f5e60b/tasks.c#L8633) and [source](https://github.com/FreeRTOS/FreeRTOS-Kernel/blob/cb900a2b7dc643130134eb5db95da037f7f5e60b/tasks.c#L8674).

This PR only add a compile error when the request for default implementations cannot be fulfilled. This way the user gets a descriptive compiler error, instead of an unhelpful "undefined symbol" link error without context. 

> As such I would remove the check for `( configSUPPORT_STATIC_ALLOCATION == 1 )` from your config check, as it is not the actual problem with the kernel provided functions.

I think this would make the check unnecessarily strict, when `configSUPPORT_STATIC_ALLOCATION == 0` the `vApplicationGet*TaskMemory()` functions are not needed, so the "misconfiguration" is IMHO harmless.

On the other hand any configuration with `configSUPPORT_STATIC_ALLOCATION == 0` and `configKERNEL_PROVIDED_STATIC_MEMORY == 1` is illogical and should perhaps be reported. But this would be applicable to both MPU and non-MPU ports. So it should be a seperate check with its own error message.

Skptak (approved):

jasonpcarroll (approved):

PR #944, opened 2024-01-04
Support reset kernel state for restarting scheduler

chinglee-iot:
Support reset kernel state

Description
-----------

This PR is originated by cmorganBE in #914.

Application may restart the scheduler for each test case when it is running with a test framework.
This requirement is useful for simulator port like posix. Some of dynamic analysis tools are difficult to run on an embedded target. valgrind is one of the example. With this feature, application writers can better utilize the test framework and valgrind to help diagnosis memory and thread bugs.

FreeRTOS kernel initialize some of the internal variables at declaration time. Therefore, init function is not required in kernel.
To restart the scheduler, kernel state needs to be reset by setting internal variables to its default value before next time starting the scheduler.

In this PR:
* Add functions to reset kernel state. These functions are only required for source file with initialized internal variables, including tasks.c, timer.c, croutine.c and heap_x.c.
  * `void vTaskResetState( void )`
  * `void vTimerResetState( void )`
  * `void vPortHeapResetState( void )`
  * `void vCoRoutineResetState( void )`

The change in the PR only re-initialize internal variables. Memory or other resource allocated during the time scheduler is running should still be freed by the port or application.

Test Steps section provides an example usage of these functions.

Test Steps
-----------
Testing with the following application on posix port.

```c
#include <stdio.h>

#include "FreeRTOS.h"
#include "task.h"

void prvTestTask( void * pvParameters )
{
    vTaskDelay( pdMS_TO_TICKS( 1000 ) );
    printf( "End scheduler\r\n" );
    vTaskEndScheduler();
}

void main_test( void )
{
    for(;;)
   {
        xTaskCreate( prvTestTask, "Test", configMINIMAL_STACK_SIZE, NULL, tskIDLE_PRIORITY + 2, NULL );
        vTaskStartScheduler();
        vTaskResetState();   /* Without this function call, there will be exception. */
    }
}

```

Checklist:
----------

- [x] I have tested my changes. No regression in existing tests.
- [ ] I have modified and/or added unit-tests to cover the code changes in this Pull Request.

Related Issue
-----------
#914

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #728, opened 2023-07-28
Enable MSVC Port to leave the prvProcessSimulatedInterrupts loop when scheduler is stopped

Ju1He1:
Enable MSVC Port to leave the prvProcessSimulatedInterrupts loop when scheduler is stopped

Description
-----------
At the moment, the prvProcessSimulatedInterrupts is stuck in an endless loop even when the scheduler is stopped. This change pulls the xPortRunning Flag (which is set to false by vPortEndScheduler() within the loop. If set to false the loop is exited.
However this is needed when FreeRTOS based modules are tested using GoogleTest (otherwise the test won't stop)

Checklist:
----------

- [x] I have tested my changes. No regression in existing tests.
- [ ] I have modified and/or added unit-tests to cover the code changes in this Pull Request.

Related Issue
-----------

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #957, opened 2024-01-18
Fix MacOS Posix port

Skptak:
Description
-----------

Changes in #914 have caused the Posix GCC FreeRTOS Port to fail to build on MacOS
This PR fixes this issue by wrapping the resetting of a variable used when stopping the scheduler in an OS type check.
This PR also adds a Matrix configuration to the GitHub kernel-demo workflow to build the Posix Demos on MacOS

The exact commit that introduced the breaking change, per the git blame, can be found [here](https://github.com/FreeRTOS/FreeRTOS-Kernel/commit/7ba4124c785e400cb0e8c3c17a49f390e90ba77e)

Test Steps
-----------

Checklist:
----------

Related Issue
-----------

### DIFF
--- a/portable/ThirdParty/GCC/Posix/port.c
+++ b/portable/ThirdParty/GCC/Posix/port.c
@@ -266,9 +266,10 @@ BaseType_t xPortStartScheduler( void )
      */
     xSchedulerEnd = pdFALSE;
 
-    /* Reset the pthread_once_t structure. This is required if the port
-     * starts the scheduler again. */
-    hSigSetupThread = PTHREAD_ONCE_INIT;
+    /* Reset pthread_once_t, needed to restart the scheduler again.
+     * memset the internal struct members for MacOS/Linux Compatability */
+    memset( ( void * ) &hSigSetupThread.__sig, _PTHREAD_ONCE_SIG_init, sizeof(_PTHREAD_ONCE_SIG_init));
+    memset( ( void * ) &hSigSetupThread.__opaque, 0, sizeof(hSigSetupThread.__opaque));
 
     /* Restore original signal mask. */
     ( void ) pthread_sigmask( SIG_SETMASK, &xSchedulerOriginalSignalMask, NULL );
--- a/portable/ThirdParty/GCC/Posix/port.c
+++ b/portable/ThirdParty/GCC/Posix/port.c
@@ -285,8 +285,6 @@ BaseType_t xPortStartScheduler( void )
     /* Reset pthread_once_t, needed to restart the scheduler again.
      * memset the internal struct members for MacOS/Linux Compatability */
     #if __APPLE__
-        /* Restarting the scheduler is currently unsupported on MacOS. */
-        configASSERT( pdFALSE );
         hSigSetupThread.__sig = _PTHREAD_ONCE_SIG_init;
         memset( ( void * ) &hSigSetupThread.__opaque, 0, sizeof(hSigSetupThread.__opaque));
     #else /* Linux PTHREAD library*/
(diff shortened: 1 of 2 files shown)


chinglee-iot:
@Skptak 
Thanks for the fix and adding mac OS build with Posix port.

chinglee-iot (approved):

portable/ThirdParty/GCC/Posix/port.c
```diff
@@ -281,10 +282,17 @@ BaseType_t xPortStartScheduler( void )
...
-     * starts the scheduler again. */
-    hSigSetupThread = PTHREAD_ONCE_INIT;
-
+    /* Reset pthread_once_t, needed to restart the scheduler again.
+     * memset the internal struct members for MacOS/Linux Compatability */
+    #if __APPLE__
+        /* Restarting the scheduler is currently unsupported on MacOS. */
+        configASSERT( pdFALSE );
+        hSigSetupThread.__sig = _PTHREAD_ONCE_SIG_init;
+        memset( ( void * ) &hSigSetupThread.__opaque, 0, sizeof(hSigSetupThread.__opaque));
```
chinglee-iot, line 291: A posix port running on mac-OS can still call `vTaskEndScheduler()` to stop the scheduler. Suggest we just keep the comment and remove these lines since restarting scheduler is not supported in mac-OS. We can consider to support the feature in mac-OS with future commit.

Skptak: Calling `vTaskEndScheduler()/vTaskStartScheduler()`in the [Posix_GCC Demo](https://github.com/FreeRTOS/FreeRTOS/tree/main/FreeRTOS/Demo/Posix_GCC) was leading to a segfault when I called it on Linux. Where when calling it in MacOS I was seeing the scheduler thread immediately dying after the call to `( void ) pthread_sigmask( SIG_SETMASK, &xSchedulerOriginalSignalMask, NULL );` on the following line.

Since I didn't have a working example to compare to I added the `configASSERT()`. I believe this would be a more "user friendly" way of learning that this feature is not supported on MacOS compare to having a seemingly paused system due to the scheduler thread being killed.

chinglee-iot: Restarting FreeRTOS kernel by calling `vTaskStartScheduler()` again after `vTaskEndScheduler()` is a new feature. Resetting kernel state is required before next time calling `vTaskStartScheduler()` again to restart scheduler. A PR https://github.com/FreeRTOS/FreeRTOS-Kernel/pull/944 to add reset state function is created for discussion. The problem to restart scheduler in Linux or MacOS may be related to kernel state is not reset to their default value.

Posix port should not have problem to stop scheduler with `vTaskEndScheduler()` for the first time. User may call `vTaskEndScheduler()` for their testing purpose. For example, in this PR https://github.com/FreeRTOS/FreeRTOS-Kernel/pull/728 user is fixing MSVC-port end scheduler function to run this port with GoogleTest. The `configASSERT` added may cause problem to users who want to stop the scheduler.

Skptak: I have removed the lines
I was running the [examples/tests](https://github.com/cmorganBE/freertos_posix) linked in #914 and was seeing segfaults on Linux. This lead me to believe this feature was unsupported, and why I added a configASSERT() instead of segfaulting.

I have run the [POSIX_GCC Blinky Demo](https://github.com/FreeRTOS/FreeRTOS/blob/d603c9b1ba39a5f1c1ce8ea42397c72cc5bfcaa3/FreeRTOS/Demo/Posix_GCC/main_blinky.c#L1) and called `vTaskEndScheduler()` on MacOS to validate these changes.

chinglee-iot (approved):

aggarg (approved):

PR #969, opened 2024-01-30
Remove Trace Facility Requirement for Several Event Groups Functions

HTRamsey:
Remove Trace Facility Requirement for Several Event Groups Functions

Description
-----------
There is no configUSE_TRACE_FACILITY requirement for these functions, and no mention in documentation of it only being used for debugging.

Test Steps
-----------

Checklist:
----------

Related Issue
-----------

### DIFF
--- a/event_groups.c
+++ b/event_groups.c
@@ -495,7 +495,7 @@ EventBits_t xEventGroupClearBits( EventGroupHandle_t xEventGroup,
 }
 /*-----------------------------------------------------------*/
 
-#if ( ( configUSE_TRACE_FACILITY == 1 ) && ( INCLUDE_xTimerPendFunctionCall == 1 ) && ( configUSE_TIMERS == 1 ) )
+#if ( ( INCLUDE_xTimerPendFunctionCall == 1 ) && ( configUSE_TIMERS == 1 ) )
 
     BaseType_t xEventGroupClearBitsFromISR( EventGroupHandle_t xEventGroup,
                                             const EventBits_t uxBitsToClear )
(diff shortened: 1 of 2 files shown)


paulbartell:
@HTRamsey thanks for the PR. I'll work on getting this reviewed and merged.

AniruddhaKanhere (commented):
I am not sure I understand this change completely.

If we are removing the dependence of the function definitions and declarations from `configUSE_TRACE_FACILITY` then should we not use the #defined definitions instead of the ones which have tracing macro calls (i.e. `traceENTER_xEventGroupClearBitsFromISR`) in them?

Am I missing something? Can you please take a look at my question @HTRamsey?

HTRamsey:
@AniruddhaKanhere I guess that's preference on if you still want to be able to use the traceENTER, traceEVENT, and traceEXIT macros for event groups separately from a generic call to `xTimerPendFunctionCallFromISR`. I kept it that way because there are other functions that could have been macros instead as well (i.e. `vEventGroupClearBitsCallback` & `vEventGroupSetBitsCallback`, which by the way I am not sure these are ever used anywhere).

RichardBarry:
Clearing event group bits form an ISR isn't done within the ISR itself because its not deterministic - the operation is instead deferred to the timer task so it occurs outside of the ISR context.  If you are not tracing, then xEventGroupsClearBitsFromISR() calls the function in the timer task directly (https://github.com/FreeRTOS/FreeRTOS-Kernel/blob/main/include/event_groups.h#L443), but doing that when tracing obfuscates the reason for calling the timer task function.  Hence, when tracing, xEventGroupsClearBitsFromISR() is a function in its own right.  So the original code, before this PR, is the intentional behaviour.  I recommend changing the PR to instead add a comment to the existing code explaining why it is so.

PR #972, opened 2024-01-31
Fix MISRA C 2012 Rule 10.4 Violations

bradleysmith23:
Description
-----------
Fix violations of MISRA C 2012 rule 10.4 - "Both operands of an operator in which the usual arithmetic conversions are performed shall have the same essential type category"

Test Steps
-----------

Checklist:
----------

Related Issue
-----------

### DIFF
--- a/portable/Common/mpu_wrappers_v2.c
+++ b/portable/Common/mpu_wrappers_v2.c
@@ -2197,7 +2197,7 @@
                     if( ( !( ( pvItemToQueue == NULL ) && ( uxQueueItemSize != ( UBaseType_t ) 0U ) ) ) &&
                         ( !( ( xCopyPosition == queueOVERWRITE ) && ( uxQueueLength != ( UBaseType_t ) 1U ) ) )
                         #if ( ( INCLUDE_xTaskGetSchedulerState == 1 ) || ( configUSE_TIMERS == 1 ) )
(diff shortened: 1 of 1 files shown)


portable/Common/mpu_wrappers_v2.c
```diff
@@ -485,7 +485,7 @@
 
             lIndex = ( int32_t ) xTask;
 
-            if( IS_EXTERNAL_INDEX_VALID( lIndex ) != pdFALSE )
```
bradleysmith23, line 488: Operands of != had different essential types (bool and int)

kstribrnAmzn (approved):

portable/Common/mpu_wrappers_v2.c
chinglee-iot: `pdTRUE` and `pdFALSE` are regarded as boolean type by coverity because of the name contain true or false. In previous MISRA fix, `pdTRUE` is replaced with `pdPASS` and `pdFALSE` is replaced with with `pdFAIL` to avoid these false positive warnings. [Reference link](https://github.com/FreeRTOS/FreeRTOS-Kernel/blob/ba1e2dad3cc6fcb4209e33093efba83d189aa836/examples/coverity/CMakeLists.txt#L8). Suggest we can adapt the same method to run coverity here instead of fixing the code for test tool.

chinglee-iot (approved):

aggarg (approved):

PR #982, opened 2024-02-05
Update task notification scheduler suspension usage

Dazza0:
Description
-----------

https://github.com/FreeRTOS/FreeRTOS-Kernel/issues/612 fixed an issue with `ulTaskGenericNotifyTake()` and `xTaskGenericNotifyWait()` being non deterministic by ensuring that those functions call `prvAddCurrentTaskToDelayedList()` while the scheduler was suspended. The merged fix would suspend the scheduler while still in a critical section, then exit the critical section while the scheduler was suspended:

```c
taskENTER_CRITICAL();
vTaskSuspendAll();
taskEXIT_CRITICAL();
prvAddCurrentTaskToDelayedList( xTicksToWait, pdTRUE );
xTaskResumeAll();
```

The method above technically works but is a bit unintuitive due to interleaving critical sections and scheduler suspension.

This PR updates the behavior described above by wrapping the critical section in a scheduler suspension block:

```c
vTaskSuspendAll();
taskENTER_CRITICAL();
// Check if block required
taskEXIT_CRITICAL();
if (blocking) {
    prvAddCurrentTaskToDelayedList( xTicksToWait, pdTRUE );
}
xTaskResumeAll();
```

This method is more intuitive. As an added bonus, it also allows the SMP implementation of `vTaskSuspendAll()` to be simplified (as we can assume the `vTaskSuspendAll()` is now never called in a critical section). As such, an extra assert has been added to `vTaskSuspendAll()`  to ensure that it is never called in a critical section.

Related Issue
-----------

- See https://github.com/FreeRTOS/FreeRTOS/pull/1177 for CMock test updates
- Related to https://github.com/FreeRTOS/FreeRTOS-Kernel/issues/612


Dazza0:
@chinglee-iot I've updated the CMock unit tests as well. PTAL.

paulbartell (dismissed):

tasks.c
```diff
@@ -3808,6 +3808,8 @@ void vTaskSuspendAll( void )
 
         /* This must only be called from within a task. */
         portASSERT_IF_IN_ISR();
+        /* This must never be called from inside a critical section */
+        configASSERT( portGET_CRITICAL_NESTING_COUNT() == 0 );
```
Skptak, line 3812: @aggarg @RichardBarry

Is this change alright? I know some ports such as the [ARM_CRx_No_GIC](https://github.com/FreeRTOS/FreeRTOS-Kernel/blob/V11.0.1/portable/GCC/ARM_CRx_No_GIC/portASM.S#L60) push the critical nesting count as part of the task context:
```
    /* Push the critical nesting count. */
    LDR     R2, ulCriticalNestingConst
    LDR     R1, [R2]
    PUSH    {R1}
```
I'm worried that this change might be a breaking change for some ports/applications and was hoping for a deeper look into this PR from you two

chinglee-iot: portGET_CRITICAL_NESTING_COUNT() is okay to me. Sharing my observation here.
Critical nesting count should only be modified in critical section and context switch inside a critical section is not allowed in SMP. Use the following scenarios for discussion:
1. portGET_CRITICAL_NESTING_COUNT() is greater than 1. This task is in a critical section and can read or update the critical nesting count without problem.
2. portGET_CRITICAL_NESTING_COUNT() is 0. The task A may get preempted by task B while reading critical nesting count. Critical nesting count will be restore to 0 when the task B left critical section.

chinglee-iot (approved):

chinglee-iot (dismissed):

aggarg (dismissed):

chinglee-iot (approved):

aggarg (approved):

PR #985, opened 2024-02-06
Fix MISRA C 2012 Rule 13.3 Violations

bradleysmith23:
Description
-----------
MISRA C 2012 Rule 13.3 states that the increment/decrement(++/--) operators should contain the only side-effect in an expression. As a result, using ++/-- on a volatile results in a violation as reading a volatile qualified variable is a side-effect itself. To remediate this issue, the value of the volatile is assigned to temporary variables in these cases, then the temporary variable is incremented or decremented before being written back to the volatile qualified variable. 

Test Steps
-----------

Checklist:
----------

Related Issue
-----------


include/list.h
```diff
@@ -326,7 +327,9 @@ typedef struct xLIST
         }                                                                        \
                                                                                  \
         ( pxItemToRemove )->pxContainer = NULL;                                  \
-        ( pxList->uxNumberOfItems )--;                                           \
+        uxNumberOfItems = ( pxList->uxNumberOfItems );                           \
```
joshzarr, line 330: I would add a comment explaining why this temp variable is being used.
It might be tempting in the future to revert this change if it is not understood why this is being done.

bradleysmith23: Agreed, will do that now

PR #988, opened 2024-02-08
Fix MISRA C 2012 Rule 13.3 Violations

bradleysmith23:
Description
-----------
Rule 13.3 states that a full expression containing the increment (++) or decrement (--) operator should have no other potential side effects other than that caused by the increment or decrement operator. These were flagged as violations of the rule since the read of a volatile variable is considered to be a side effect in itself. As a result, using `+= 1` and `-= 1` in place of `++` and `--` brings these lines into compliance with the rule.

Test Steps
-----------

Checklist:
----------

Related Issue
-----------


include/list.h
```diff
@@ -326,7 +326,7 @@ typedef struct xLIST
         }                                                                        \
                                                                                  \
         ( pxItemToRemove )->pxContainer = NULL;                                  \
-        ( pxList->uxNumberOfItems )--;                                           \
+        pxList->uxNumberOfItems -= 1U;                                           \
```
kar-rahul-aws, line 329: pxList->uxNumberOfItems has the type _UBaseType_t_ . This operation leads to -Wconversion error , conversion from ‘unsigned int’ to ‘UBaseType_t’ . It can be changed to 
`pxList->uxNumberOfItems = ( UBaseType_t )( pxList->uxNumberOfItems - 1U );`

tasks.c
```diff
@@ -255,7 +255,7 @@
         pxTemp = pxDelayedTaskList;                                               \
         pxDelayedTaskList = pxOverflowDelayedTaskList;                            \
         pxOverflowDelayedTaskList = pxTemp;                                       \
-        xNumOfOverflows++;                                                        \
+        xNumOfOverflows += 1;                                                     \
```
kar-rahul-aws, line 258: **xNumOfOverflows** has the type BaseType_t . This operation leads to -Wconversion error , conversion from ‘unsigned int’ to ‘BaseType_t’ . It can be changed to
`xNumOfOverflows = ( BaseType_t )( xNumOfOverflows + 1U );`

tasks.c
```diff
@@ -3807,7 +3807,7 @@ void vTaskSuspendAll( void )
 
         /* The scheduler is suspended if uxSchedulerSuspended is non-zero.  An increment
          * is used to allow calls to vTaskSuspendAll() to nest. */
-        ++uxSchedulerSuspended;
+        uxSchedulerSuspended += 1U;
```
kar-rahul-aws, line 3810: **uxSchedulerSuspended** has the type UBaseType_t . This operation leads to -Wconversion error , conversion from ‘unsigned int’ to ‘UBaseType_t’ .

bradleysmith23: I am happy to change it to this implementation, but went with leaving the compound assignment operators in and casting the value on the right side of the operator to be the appropriate type as the C standard states they differ in that "A compound assignment of the form E1 op= E2 differs from the simple assignment expression E1 = E1 op (E2) only in that the lvalue E1 is evaluated only once."

tasks.c
bradleysmith23: I am happy to change it to this implementation, but went with leaving the compound assignment operators in and casting the value on the right side of the operator to be the appropriate type as the C standard states they differ in that "A compound assignment of the form E1 op= E2 differs from the simple assignment expression E1 = E1 op (E2) only in that the lvalue E1 is evaluated only once."

kar-rahul-aws (approved):

include/list.h
```diff
@@ -326,7 +326,7 @@ typedef struct xLIST
         }                                                                        \
                                                                                  \
         ( pxItemToRemove )->pxContainer = NULL;                                  \
-        ( pxList->uxNumberOfItems )--;                                           \
+        ( ( pxList )->uxNumberOfItems ) -= ( UBaseType_t ) 1U;                   \
```
htibosch, line 329: One simplification:
~~~diff
-    ( ( pxList )->uxNumberOfItems ) -= ( UBaseType_t ) 1U; 
+    ( ( pxList )->uxNumberOfItems ) -= 1U; 
~~~
I think that the cast is not necessary. Have a try.

Also try:
~~~diff
-    xNumOfOverflows += ( BaseType_t ) 1;   
+    xNumOfOverflows += 1;   
~~~

aggarg (approved):

tasks.c
cobusve: I am curious why this used to be a pre-increment, that seems slightly odd, at least unexpected in this case.

include/list.h
cobusve: The problem is that BaseType_t can be a different size from "unsigned", which means this will work fine with some types as BaseType_t but not others. These cases are tricky because compiling only with one type for BaseType_t will not expose all the problems !

tasks.c
cobusve: And of course like the other one I am curious why this variable is volatile, because if it is indeed volatile then the read/modify/write we are doing here creates a potential race condition where we read it, then add 1, then the value gets incremented from an ISR e.g. , and then we overwrite with an incorrect value here ?

tasks.c
cobusve: I am also curious here why this variable is volatile, because if it can change (and the post-increment causes a rmw race) then doing the +=1 will create the exact same race right?

PR #995, opened 2024-02-14
Revert "Fix MISRA C 2012 Rule 13.3 Violations"

bradleysmith23:
Reverts FreeRTOS/FreeRTOS-Kernel#988 
After further discussion, this PR does not appropriately address this issue and that the use of the volatile qualifier with uxNumberOfItems should be updated to be configLIST_VOLATILE to be inline with the use of this macro across the rest of the file which will be done in a PR following this one.

### DIFF
--- a/include/list.h
+++ b/include/list.h
@@ -326,7 +326,7 @@ typedef struct xLIST
         }                                                                        \
                                                                                  \
         ( pxItemToRemove )->pxContainer = NULL;                                  \
-        ( ( pxList )->uxNumberOfItems ) -= ( UBaseType_t ) 1U;                   \
+        ( pxList->uxNumberOfItems )--;                                           \
     } while( 0 )
 
 /*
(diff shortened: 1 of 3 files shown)


cobusve:
Check also the other cases for similar race conditions. I think the post increments and pre-increments should be fine if these variables are in fact not race conditions, and if volatile is required then we need to carefully evaluate what happens when we do a read/modify/write in a non-atomic way on these variables?

kar-rahul-aws:
We can revert the changes made in the merged PR  [988](https://github.com/FreeRTOS/FreeRTOS-Kernel/pull/988) , since these variables are not in race condition. These variables are only modified in a functionally atomic way ( **within critical sections or with the scheduler suspended** ) and are either passed by reference into a function or indexed via a volatile variable. The `configLIST_VOLATILE` macro was added with the motivation for advanced compiler technology , where a compiler might remove essential code leading to compilation error. In those cases, then the volatile qualifier can be inserted in the relevant places within the list structures by simply defining configLIST_VOLATILE to volatile in FreeRTOSConfig.h .

chinglee-iot (approved):

kar-rahul-aws:
Since the PR #988 does not essentially change the code generated, and it addresses the coverity issues, it works as expected without reverting the said PR. Hence closing this PR #995

PR #996, opened 2024-02-15
Add default core affinity config value.

### LINKED ISSUE #993: [Feature Request] Default Affinity Set
**Is your feature request related to a problem? Please describe.**
We created a freeRTOS SMP port for the Infineon TC3x and TC4x family, but the system design is not really symmetrical. Especially the Task context ist bound to the CPU executing the task. This makes switching tasks between different CPUs nearly impossible. But the SMP design still makes some sense to distribute work across cores more easily. For example in conjunction with FreeRTOS-Plus-TCP to distribute sockets across cores.

**Describe the solution you'd like**
Create a new config flag to allow setting a default affinity for a task, where no affinity set is provided during task creation. This is especially useful for legacy code. This could either be a fixed value or the current core, who creates the task.

**Describe alternatives you've considered**
Non

**How many devices will this feature impact?**
All TC3x and TC4x devices with SMP enabled.

**What are your project timelines?**
A patch is already prepared internally.

go2sh:
Description
-----------
Add a config define to set the default core affinity mask for tasks created without an affinity mask specified via arguments.

Test Steps
-----------
Set configTASK_DEFAULT_CORE_AFFINITY to 1 and check if all tasks created without an affinity mask specified run on core 0.

Checklist:
----------

Related Issue
-----------
Closes #993.

### DIFF
--- a/include/FreeRTOS.h
+++ b/include/FreeRTOS.h
@@ -486,7 +486,7 @@
 
 #if ( ( configNUMBER_OF_CORES > 1 ) && ( configUSE_CORE_AFFINITY == 1 ) )
     #ifndef configTASK_DEFAULT_CORE_AFFINITY
-        #define configTASK_DEFAULT_CORE_AFFINITY tskNO_AFFINITY
+        #define configTASK_DEFAULT_CORE_AFFINITY    tskNO_AFFINITY
     #endif
 #endif
 
--- a/include/FreeRTOS.h
+++ b/include/FreeRTOS.h
@@ -484,6 +484,12 @@
     #define configUSE_CORE_AFFINITY    0
 #endif /* configUSE_CORE_AFFINITY */
 
+#if ( ( configNUMBER_OF_CORES > 1 ) && ( configUSE_CORE_AFFINITY == 1 ) )
+    #ifndef configTASK_DEFAULT_CORE_AFFINITY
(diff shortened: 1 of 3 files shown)


include/FreeRTOS.h
```diff
@@ -484,6 +484,12 @@
     #define configUSE_CORE_AFFINITY    0
 #endif /* configUSE_CORE_AFFINITY */
 
+#if ( ( configNUMBER_OF_CORES > 1 ) && ( configUSE_CORE_AFFINITY == 1 ) )
+    #ifndef configTASK_DEFAULT_CORE_AFFINITY
+        #define configTASK_DEFAULT_CORE_AFFINITY tskNO_AFFINITY
+    #endif
+#endif
```
rawalexe, line 491: How does this help? The check is the same and the assign conditions look the same

rawalexe:
There is a formatting patch generated at the artifacts section of ci/cd can you also please apply that to your code.

include/FreeRTOS.h
go2sh: Can leave the check out. The define is not required if the two options are not set in that way. Defining it always doesn't harm either

rawalexe (approved):

aggarg (approved):

PR #998, opened 2024-02-17
Fix small bugs in Kernel

AniruddhaKanhere:
- Move the default value of `configUSE_MALLOC_FAILED_HOOK` as it is used in "portable.h"
- Add extra parenthesis in the definition of `portMAX_DELAY`.
- ~~Add check in the function `prvCreateIdleTasks` so that when the length of `configIDLE_TASK_NAME` is less than `configMAX_TASK_NAME_LEN` we do not go over `configIDLE_TASK_NAME`.~~

Description
-----------

Test Steps
-----------

Checklist:
----------

Related Issue
-----------

### DIFF
--- a/tasks.c
+++ b/tasks.c
@@ -3507,7 +3507,7 @@ static BaseType_t prvCreateIdleTasks( void )
     BaseType_t xIdleTaskNameIndex;
 
     /* Make sure that the below for loop doesn't go over the configIDLE_TASK_NAME macro. */
-    configASSERT( sizeof( configIDLE_TASK_NAME ) >= configMAX_TASK_NAME_LEN );
+    //configASSERT( sizeof( configIDLE_TASK_NAME ) >= configMAX_TASK_NAME_LEN );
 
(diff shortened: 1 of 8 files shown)


tasks.c
```diff
@@ -3506,6 +3506,9 @@ static BaseType_t prvCreateIdleTasks( void )
     TaskFunction_t pxIdleTaskFunction = NULL;
     BaseType_t xIdleTaskNameIndex;
 
+    /* Make sure that the below for loop doesn't go over the configIDLE_TASK_NAME macro. */
+    //configASSERT( sizeof( configIDLE_TASK_NAME ) >= configMAX_TASK_NAME_LEN );
```
htibosch, line 3510: Why is this new assert commented with a `//`?

AniruddhaKanhere: Ah no, I was trying to figure out why are the checks failing.

Skptak:
Do we know what version of coverity is used in the [coverity_scan.yml](https://github.com/FreeRTOS/FreeRTOS-Kernel/blob/main/.github/workflows/coverity_scan.yml) file and if the changes to the config are compatible with that version?

aggarg (approved):

kar-rahul-aws (approved):

PR #1005, opened 2024-03-01
ARM CM0+ MPU Port

Skptak:
Description
-----------
This pull request creates an ARM Cortex M0 Plus MPU Enabled Port
These changes have been overlayed onto the ARM_CM23_NTZ/non_secure port to create a minimal diff for the review
When the changes have been reviewed and approved this will then be moved into a separate directory.

Test Steps
-----------

Checklist:
----------

Related Issue
-----------


portable/GCC/ARM_CM23_NTZ/non_secure/mpu_wrappers_v2_asm.c
Skptak, line 1: The changes in this file are because the M0+ only allows +- 2KB jumps
As such we need to push the LR and then use it for the jump
https://developer.arm.com/documentation/dui0497/a/the-cortex-m0-instruction-set/branch-and-control-instructions/b--bl--bx--and-blx

portable/GCC/ARM_CM23_NTZ/non_secure/port.c
```diff
@@ -185,9 +186,11 @@ typedef void ( * portISR_t )( void );
 #define portMPU_MAIR0_REG                       ( *( ( volatile uint32_t * ) 0xe000edc0 ) )
 #define portMPU_MAIR1_REG                       ( *( ( volatile uint32_t * ) 0xe000edc4 ) )
 
-#define portMPU_RBAR_ADDRESS_MASK               ( 0xffffffe0 ) /* Must be 32-byte aligned. */
+#define portMPU_RBAR_ADDRESS_MASK               ( 0xffffff00 ) /* Must be 32-byte aligned. */
```
kstribrnAmzn, line 189: Was this a typo in the original port? Why do we need to now change the RBAR mask? This seems to imply the region is larger than the minimum size - https://developer.arm.com/documentation/dui0662/b/Cortex-M0--Peripherals/Memory-Protection-Unit/MPU-Region-Base-Address-Register?lang=en

portable/GCC/ARM_CM23_NTZ/non_secure/port.c
```diff
@@ -185,9 +186,11 @@ typedef void ( * portISR_t )( void );
...
 #define portMPU_RLAR_ADDRESS_MASK               ( 0xffffffe0 ) /* Must be 32-byte aligned. */
+#define portMPU_RASR_LENGTH_MASK                ( 0x1F )       /* Must be [7:31]. */
```
kstribrnAmzn, line 191: Why not just make this `0x3E` and avoid the later shifting in `portEXTRACT_REGION_LENGTH_FROM_RASR`

portable/GCC/ARM_CM23_NTZ/non_secure/portmacrocommon.h
```diff
@@ -144,13 +144,13 @@ extern void vClearInterruptMask( uint32_t ulMask ) /* __attribute__(( naked )) P
 #endif
 
 /* MPU regions. */
-#define portPRIVILEGED_FLASH_REGION                   ( 0UL )
-#define portUNPRIVILEGED_FLASH_REGION                 ( 1UL )
+#define portPRIVILEGED_FLASH_REGION                   ( 6UL )
+#define portUNPRIVILEGED_FLASH_REGION                 ( 5UL )
 #define portUNPRIVILEGED_SYSCALLS_REGION              ( 2UL )
```
kstribrnAmzn, line 149: Should this lie smack dab in the middle of the configurable regions? Why not have it at either the beginning or end given it has to be configured if the MPU is enabled?

portable/GCC/ARM_CM23_NTZ/non_secure/portmacrocommon.h
```diff
@@ -178,14 +178,25 @@ extern void vClearInterruptMask( uint32_t ulMask ) /* __attribute__(( naked )) P
...
 /*-----------------------------------------------------------*/
 
+/* Attributes used in MPU_RASR registers. */
+#define portMPU_REGION_PRIVILEGED_READ_WRITE_UNPRIV_READ_ONLY    ( 0x2UL << 24UL )
+#define portMPU_REGION_STRONGLY_ORDERED                          ( 0x0UL << 16UL )
+#define portMPU_REGION_DEVICE_MEMORY                             ( 0x1UL << 16UL )
+#define portMPU_REGION_CACHEABLE_BUFFERABLE                      ( 0x7UL << 16UL )
+
+#define portMPU_REGION_VALID                      ( 0x10UL )
+#define portMPU_RASR_TEX_S_C_B_MASK               ( 0x3FUL )
```
kstribrnAmzn, line 196: Should this be `0x7` which is then shifted by 16 bits?

portable/GCC/ARM_CM23_NTZ/non_secure/portasm.c
```diff
@@ -498,16 +491,24 @@ void vClearInterruptMask( __attribute__( ( unused ) ) uint32_t ulMask ) /* __att
             "    ldr r3, [r0, #24]          \n"
             "    subs r3, #2                \n"
             "    ldrb r2, [r3, #0]          \n"
-            "    cmp r2, %0                 \n"
+            "    ldr r3, =%0                \n"
```
kstribrnAmzn, line 494: What does the `=%` notation mean?

portable/GCC/ARM_CM23_NTZ/non_secure/port.c
kstribrnAmzn: Both will work here however it is done this way more to align with conventions.

portable/GCC/ARM_CM23_NTZ/non_secure/portmacrocommon.h
kstribrnAmzn: This was done as the CM0+ allows overlapping regions with the highest numbered region taking precedence. The V8M does not allow overlapping regions.

aggarg (approved):

joshzarr (approved):
Verified I was able to compile the port and run using the M0+ demo application.
Deferring a full review to Gaurav who has already approved this PR

PR #1031, opened 2024-04-13
Implemented heap_6 for C++ real-time applications

svenbieg:
Description
-----------
I've adjusted my heap to FreeRTOS like posted in the forum:

https://forums.freertos.org/t/real-time-memory-manager/19685

This is my very first commit to FreeRTOS, i don't even know how to compile the kernel. Maybe You get it running and let me know what You think about it.

Test Steps
-----------
Because i don't know how to compile FreeRTOS, i did my tests in a simple console application.
I allocated variable sized blocks, deleted every second entry and allocated again.

You can find my test here: 
[HeapTest.zip](https://github.com/FreeRTOS/FreeRTOS-Kernel/files/14968482/HeapTest.zip)

By submitting this pull request, I confirm that You can use, modify, copy, and redistribute this contribution, under the terms of Your choice.

Best regards,

Sven Bieg


svenbieg:
I've made a test with the WIN32-MSVC demo:

full_demo

Looking good!

svenbieg:
There is a mistake when the heap is full. Adding a block to the map can fail, i have to check for this and add the block to the cache if so. I'm going to fix this when i'm at home.

svenbieg:
I also thought about this, but none of Your heaps is really working.

heap_1 doesn't allow deletion
heap_2 is worse than heap_4, both won't get certified
heap_3 redirects to the standard-heap, also not certified
heap_5 heap_4 with multiple regions
heap_6 multiple regions and almost ready for certification -> Heap

Certification absolutely makes sense, there are a lot of people in the forum interested in this. Without a directory You get lost in parsing endless chains of gaps every time allocating. In my impelentation there is one pyramidal directory with all the different sizes of the gaps. When the smallest gap is found, there is another pyramidal directory with the offsets if there are multiple. I thought about chains, but it is necessary to remove a gap when combining free blocks.

The only readme i have right now is my GitHub-wiki, there is a link in the heap source-files. The benchmark above shows there is only little less memory available, this is the price for constant low time allocation. On my development-machine it was about 60µs per allocation/deallocation, You would need 10 times more gaps to double this duration.

There is still something wrong in my code when having multiple levels in the offset_list, this means 10^5 gaps of the same size. I'm going to fix this, i have to deal with empty groups. It may take some time to get this done, i'm working in electric industry all day. It is just too much for me in the evening after work. Maybe i'll find it out on Saturday, there is nothing more interesting to me.

To tell You the truth, i found this solution by the way while developing a database.

See You, my friend!

svenbieg:
To reproduce my error You can allocate 100,000 offset_index_item_groups. When i fixed this, i'm going to test 100,000 gaps of differnt sizes to overdrive the block_map, maybe i have the same problem there. That's it, then we can add the Amazon-header without any warning.

svenbieg:
To reproduce my error You can allocate 100,000 offset_index_item_groups and delete every second. If You can fix this, You can add the Amazon-header and buy any cerificate on the market.

svenbieg:
The error is in offset_index_parent_group when splitting a child. This child can get empty while creating a new group, then an item is moved that doesn't exist anymore. It's not really hard to fix.

svenbieg:
I'm going to take a look into the standard-implementation, i guess i can add this function too.

jnorre:
Hi Sven,

Johnny from Denmark. Nice to see you provided an example. Thanks!


> I also thought about this, but none of Your heaps is really working.

What do you mean? All of the heaps are working as described! FreeRTOS is used right now with these heaps in production software, safety critical software and in aviation software.


> heap_1 doesn't allow deletion heap_2 is worse than heap_4, both won't get certified heap_3 redirects to the standard-heap, also not certified heap_5 heap_4 with multiple regions heap_6 multiple regions and almost ready for certification -> Heap
> 

What certification(s) are you talking about (do you have a specific ISO, EN or DIN standard in mind)? 

When **danielglasser** (https://forums.freertos.org/u/danielglasser) talk about **DO-178B** not allowing deallocation, then it has nothing to do with the existing heap managers or with malloc/free as provided by the compiler - it is a requirement by the standard because this leave less points where the program can fail. Same with MISRA: Analysing what causes programs to crash and then making rules for how to make your program to avoid these pitfalls. Your heap changes **nothing** in this regard.


With your example I tried to test the code for the two things I value for embedded projects: Speed and memory usage/overhead. The test (after setting **configUSE_MALLOC_FAILED_HOOK** to 0) using 100 bytes memory block was as follows:
```
heap_5.c   Can allocate 405 blocks 
           1000000 malloc/free took 87 ms.
heap_6.c   Can allocate 451 blocks 
           1000000 malloc/free took 356 ms.
```
So your heap wins on number of allocations, but lose on time. Thats actually nice! However, heap_5.c is normally not used that much as we often can allocate static variables in the non-contiguous blocks, so here is the results for a contiguous memory area of **mainREGION_1_SIZE+mainREGION_2_SIZE+mainREGION_3_SIZE** (configTOTAL_HEAP_SIZE set to 50913 bytes):
```
heap_2.c   Can allocate 454 blocks 
           1000000 malloc/free took 79ms.
heap_4.c   Can allocate 454 blocks 
           1000000 malloc/free took 85ms.
heap_1.c   Can allocate 509 blocks 
           1000000 malloc/free took 73ms. (modified to allow test)
```
As you can see these three heaps are faster and has less overhead.

An observation: Your code does not respect the **configUSE_MALLOC_FAILED_HOOK** flag - it indicates that a failed allocation should call the provided **vApplicationMallocFailedHook** function instead of returning NULL.

So from my point of view, your memory manager seems to have a nice use case for multiple non-contiguous memory blocks, not at least because it can provide more allocated blocks than heap_5, but only **if** it is okay that it takes more time to allocate/deallocate. 

Best regards,
Johnny Norre

svenbieg:
Good evening Johnny,

thanks for Your reply. Please don't get angry because i think Your heaps don't work, they just don't work for me. As You can imagine, i have a lot more memory and multiple cores in my modern C++ application. I just want to be sure the heap won't slow down after a few days of operation.

I'm a hobby-programmer without any restrictions, i compile my C++ application right into the kernel. Probably there is no certification for this, this would be nice. VisualStudio debugging, IntelliSense and modern C++ should be enough to get an application running reliably with some experience.

Thank You again for the benchmarks! Your problem is getting more and more unusable gaps, slowing down the heap over time. Maybe it is just my imagination, but Tesla seemed to have problems nobody knows how to solve but me. Their AI is an illusion in my eyes. It was trained with modern C++ and will never get my insurance like You.

Your heaps may be faster, but not for every application on the long run. I'm going to put my implementation into MemMang today, for future use.

Thank You very much for Your notice on configUSE_MALLOC_FAILED_HOOK, i'm going to implement this. I just couldn't find it in heap_5.c

See You, Johnny! :-)

svenbieg:
The readme is up to You, i don't have access to it.

Shall i insert the Amazon-header with a link to my original source?

jnorre:
Hi,

> thanks for Your reply. Please don't get angry because i think Your heaps don't work, they just don't work for me. As You can imagine, i have a lot more memory and multiple cores in my modern C++ application. I just want to be sure the heap won't slow down after a few days of operation.

I don't get mad, but I get mildly annoyed when you can't see that the most common use of FreeRTOS are for embedded programming, where the resources (processsing power, flash memory and RAM) are very limited. The Win32 example is not even a good demonstration of FreeRTOS tasks or memory usage, as it maps to Windows threads and uses Windows stacks, which are two of the more memory consuming things in a embedded FreeRTOS based program. The Windows port is good for simulating an embedded program to try out concepts in an unlimited environment, but it is not used to make Windows multi-threaded programs.

> I'm a hobby-programmer without any restrictions, i compile my C++ application right into the kernel. Probably there is no certification for this, this would be nice. VisualStudio debugging, IntelliSense and modern C++ should be enough to get an application running reliably with some experience.

And many of us using FreeRTOS are professional programmers which make actual programs for actual devices. The certification we talk about is e.g. software for heavy machinery, where peoples health or lives is at risk if a controlling device or the software running it crashes. For this reason, there are some international standards we have to adhere to and often an actual certification process of the system (hardware, software and the system it controls) performed by certification bodies. Please respect, that some people has used FreeRTOS for a long time and maybe, just maybe have a little bit of more knowledge of how it works and what it needs, than other people.

> Thank You again for the benchmarks! Your problem is getting more and more unusable gaps, slowing down the heap over time. Maybe it is just my imagination, but Tesla seemed to have problems nobody knows how to solve but me. Their AI is an illusion in my eyes. It was trained with modern C++ and will never get my insurance like You.

I've heard Mr. Musk talk about software stuff from time to time (AI included) indicating he has little to no actual knowledge about the science or programming behind it. He is an excellent business man, however - respect for that.

> Your heaps may be faster, but not for every application on the long run. I'm going to put my implementation into MemMang today, for future use.

Thank you! As mentioned, after you made the example and used time to implement it for FreeRTOS, some people might find use for it. From my experience with the thing I have worked on (heavy machinery like cranes and foresting machines, earbuds, multiple surveillance systems) heap_2 is mostly used (maybe heap_4 for newer projects), followed by heap_1 and heap_3. Heap_5 is a niche as most embedded systems does not have the discontinuous memory that it is used for. However, more and more new microcontrollers seems to have more complex memory maps, so it might be more useable in the future. 

Best regards,
Johnny Norre

svenbieg:
A small step for You, but a great step for humanity.

svenbieg:
ISO-2024

svenbieg:
To reproduce the error You can allocate 100,000 offset_index_item_groups and delete every second. Something is missing in the first offset_index_parent_group on level 1. There may be an empty item-group when trying to split. I'm going to take a closer look on the weekend.

Best regards,

Sven Bieg

svenbieg:
Johnny,

You have all my respect. The point is that heap_6 won't affect Your heavy machinery at all. With the time going on You won't be able to get Your limited computing hardware anymore. What we need is human-readable code, so You can start over writing Your code in C++ again. My point of view is Your firmware getting smaller, easier to read, to adopt and to reproduce.
I understand Your need for certification, i won't live without it. You're just not able to proof my software has security issues, because You don't know it. I've been using ESP32 to drive my father's thermal solar power-plant, this is where i found FreeRTOS. My controller was working for two years until the solar-cells exploded. I still believe it was a hacker-attack, i didn't have SSL at this time. You can checkout my facebook-profile, there are some photos.

See You! :-)

svenbieg:
I don't have time for this, keep it or leave it!

svenbieg:
Please take with care, there is still something missing. I'm going to fix this on the weekend.

Best regards,

Sven Bieg

svenbieg:
I know You have an ARMv8 implementation with gigabytes of RAM.
I'm sure Your customers won't get angry next week.

svenbieg:
Johnny,

i really don't want to stress You. Just because You are in it, can You measure peak performance with thousands of too small gaps? You can reach this by allocating 8 bytes in a loop, deleting every second entry and allocating 4 bytes again. When You have enough gaps You can start measuring.

It's annoying me to see people who didn't even try to get the point and have to put thoughts here.

Not You, Johnny.

svenbieg:
aggarg,

if You can't read You are wrong in this place.

Can You fix my problem and measure peak performance?

svenbieg:
To destroy heap_4, Johnny?

svenbieg:
I've made my training in chemical industry, maybe You know Evonik. When there is something critical they have hardware locking.

svenbieg:
heap_6 is nothing more than heap_5 with a dictioanry. A logical consequence, an addition.

Never change a running system!

RichardBarry:
Thanks for your PR, and enthusiasm for the topic.  Having reviewed the code, I agree with [aggarg@’s post](https://github.com/FreeRTOS/FreeRTOS-Kernel/pull/1031#issuecomment-2061436987) – hence closing the PR here in the hope you create a PR in the community contributions repo as the best way of making this code available.  I feel the code is a little more complex than users expect for FreeRTOS, and from some of your comments, may contain issues.  Plus, some of the novelty would require signing of a Contributor License Agreement as mentioned at the bottom of the [CONTRIBUTING.md](https://github.com/FreeRTOS/FreeRTOS-Kernel/blob/bbc058967be6446f5e4cf486cad736856b53dd69/.github/CONTRIBUTING.md) file, which would then trigger escalated reviews on our side.

svenbieg:
Good bye

svenbieg:
Please delete my FreeRTOS-account and everything i have posted.

Amateurs!

svenbieg:
I am friendly.

svenbieg:
I'm Your new broke-vulture as You want.

svenbieg:
"WARNING without any warranty"

Amateurs!

svenbieg:
We all make mistakes :-;

svenbieg:
We all make mistakes ;-)

svenbieg:
I fixed my issue, it is ready for certification.

[FreeRTOS.zip](https://github.com/FreeRTOS/FreeRTOS-Kernel/files/15049950/FreeRTOS.zip)

Good bye!

PR #1046, opened 2024-04-29
Update ARM_CM0 to ARM_CM0_MPU in GCC

kar-rahul-aws:
Description
-----------
This PR updates ARM_CM0 to ARM_CM0_MPU for GCC port.

Test Steps
-----------
NA

Checklist:
----------


kar-rahul-aws:
The reason we have _MPU is because we have 2 separate ports - one MPU and one without MPU. For example, CM4F and CM4_MPU.
In this case, there is only one port and the MPU support is enabled using configENABLE_MPU . Therefore, this change adds no value other than breaking the existing applications.

PR #1065, opened 2024-05-20
Improve multiple accessing pxCurrentTCB in a function

chinglee-iot:
Description
-----------
Address Richard's feedback to update the usage of `pxCurrentTCB` in a function.
Since `pxCurrentTCB` is a volatile variable, accessing it multiple times can be slow if this pointer is not changed.

In this PR:
* Use a const variable to access pxCurrentTCB in a function to prevent multiple accessing a volatile variable.
* The `prvGetCurrentTaskTCB()` and `prvGetCurrentTaskTCBUnsafe()` functions are provided. `prvGetCurrentTaskTCBUnsafe()` can improve performance in SMP environments as it does not disable/enable interrupts. However, it should only be called in contexts where race conditions cannot occur.

Test Steps
-----------
Performance comparison test.

Checklist:
----------

Related Issue
-----------

### DIFF
--- a/tasks.c
+++ b/tasks.c
@@ -84,16 +84,17 @@
         portYIELD_WITHIN_API();                                  \
     } while( 0 )
 
-        #define taskYIELD_ANY_CORE_IF_USING_PREEMPTION( pxTCB ) \
-    do {                                                        \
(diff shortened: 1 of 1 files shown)


felixvanoost:
Hi @chinglee-iot, are there still plans to merge these optimizations?

PR #1079, opened 2024-06-05
Fix compilation warning about undelcared pthread_setname_np

maxiaogood:
Description
-----------
Use `_GNU_SOURCE` macro instead of `__USE_GNU` and define it before including portmacro.h. The reason is that portmacro.h includes limits.h which in-turn includes features.h - this results in `__USE_GNU` getting incorrectly undefined. 

Test Steps
-----------
1. use glibc2.24 and gcc 6.3.0 build sucess
2. use glibc2.35 and gcc 11.3.0 build sucess

Checklist:
----------

Related Issue
-----------
#1077 

### DIFF
--- a/portable/ThirdParty/GCC/Posix/port.c
+++ b/portable/ThirdParty/GCC/Posix/port.c
@@ -49,12 +49,10 @@
 * only or serialized with a FreeRTOS primitive such as a binary
 * semaphore or mutex.
 *----------------------------------------------------------*/
-#include "portmacro.h"
-
 #ifdef __linux__
-    #define __USE_GNU
+    #define _GNU_SOURCE
 #endif
-
+#include "portmacro.h"
 #include <errno.h>
 #include <pthread.h>
 #include <limits.h>
--- a/portable/ThirdParty/GCC/Posix/port.c
+++ b/portable/ThirdParty/GCC/Posix/port.c
@@ -61,9 +61,7 @@
 #include <string.h>
 #include <sys/time.h>
 #include <sys/times.h>
-#ifndef __APPLE__
-    #include <sys/prctl.h>
-#endif
+#include <sys/prctl.h>
 #include <time.h>
 #include <unistd.h>
 
--- a/portable/ThirdParty/GCC/Posix/port.c
+++ b/portable/ThirdParty/GCC/Posix/port.c
@@ -51,6 +51,9 @@
 *----------------------------------------------------------*/
 #include "portmacro.h"
 
+#ifdef __linux__
+    #define __USE_GNU
+#endif
 
 #include <errno.h>
 #include <pthread.h>
@@ -61,7 +64,6 @@
 #include <string.h>
 #include <sys/time.h>
 #include <sys/times.h>
-#include <sys/prctl.h>
 #include <time.h>
 #include <unistd.h>
 
@@ -139,7 +141,7 @@ static void prvPortSetCurrentThreadName(char * pxThreadName)
 #ifdef __APPLE__
     pthread_setname_np(pxThreadName);
 #else
-    prctl( PR_SET_NAME, pxThreadName );
+    pthread_setname_np(pthread_self(), pxThreadName);
 #endif
 }
 /*-----------------------------------------------------------*/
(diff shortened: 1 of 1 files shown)


portable/ThirdParty/GCC/Posix/port.c
```diff
@@ -51,9 +51,6 @@
 *----------------------------------------------------------*/
 #include "portmacro.h"
 
-#ifdef __linux__
-    #define __USE_GNU
```
aggarg, line 55: Did you try changing this to `#define _GNU_SOURCE`?

maxiaogood: I tried two compilation environments, but they both reported warnings
image

aggarg: Does it happen even if you remove this `#ifdef __linux__` check?

maxiaogood: yes
image
image

aggarg: And which OS are you using? Is it Ubuntu?

maxiaogood: yes，ubuntu:22.04

maxiaogood: Another is a docker

aggarg: That is strange because I do not see this warning on Ubuntu 22.04. Can you please do the following:

1. Locate pthread.h:
    ```
    locate pthread.h 
    ```
2. Open it and find `pthread_setname_np` in it.

My pthread.h is located at `/usr/include/pthread.h` and I see the following declaration in it:
```
#ifdef __USE_GNU
/* Get thread name visible in the kernel and its interfaces.  */
extern int pthread_getname_np (pthread_t __target_thread, char *__buf,
                               size_t __buflen)
     __THROW __nonnull ((2));

/* Set thread name visible in the kernel and its interfaces.  */
extern int pthread_setname_np (pthread_t __target_thread, const char *__name)
     __THROW __nonnull ((2));
#endif
```

maxiaogood: My file ptthread. h is the same as yours. I tried a pure ubuntu22.04 Docker, but it still has this warning. The test log is as follows
[test_ubuntu22.04.txt](https://github.com/user-attachments/files/15688487/test_ubuntu22.04.txt)

aggarg: This does not make sense - Either `__USE_GNU` is not getting defined correctly or you are using a different pthread.h file. Can you dump the preprocessed port.c file?

maxiaogood: [port.zip](https://github.com/user-attachments/files/15730120/port.zip)
image
This is my "port. c" and compiled. o files, as well as the "ptthread. h" and "features. h" in the system library

aggarg: Would you please use the slightly modified port.c: [port.zip](https://github.com/user-attachments/files/15756812/port.zip). Also, please dump the pre-processed port.c file ans share - you can use gcc `-E` option for that.

maxiaogood: [port.zip](https://github.com/user-attachments/files/15757234/port.zip)
this is use gcc -E

aggarg: Did you use the modified file I shared? I changed `_GNU_SOURCE` to `__USE_GNU` in port.c:
```
#define __USE_GNU
```

If you want to keep using `_GNU_SOURCE`, please move that to before including `portmacro.h`:

```c
#define _GNU_SOURCE

#include "portmacro.h"
#include <errno.h>
#include <pthread.h>
#include <limits.h>
#include <signal.h>
#include <stdio.h>
#include <stdlib.h>
#include <string.h>
#include <sys/time.h>
#include <sys/times.h>
#include <time.h>
#include <unistd.h>
```

maxiaogood: i move _GNU_SOURCE that to before including portmacro.h, No warning is reported, and it can be compiled on gcc 6.3 and glibc 2.24

aggarg: Thank you for following up and confirming.

aggarg (approved):

kar-rahul-aws (approved):

PR #1081, opened 2024-06-05
Clarify documentation of xTaskNumber in TaskStatus_t.

schilkp:
Description
-----------

Adds a note to the (confusingly named) xTaskNumber member of TaskStatus_t explaining that the value is different to the TaskNumber that can be accessed using the vTaskSetTaskNumber and vTaskGetTaskNumber functions.

The value returned is actually the value of uxTCBNumber in the TCB.

Note that documentation on the website should also be updated (https://www.freertos.org/vTaskGetInfo.html#TaskStatus_t). Is this done automatically based on code comments or does this need to be done manually?

Thanks!

Test Steps
-----------

N/A

Checklist:
----------

N/A

Related Issue
-----------

N/A

### DIFF
--- a/include/task.h
+++ b/include/task.h
@@ -161,7 +161,7 @@ typedef struct xTASK_STATUS
 {
     TaskHandle_t xHandle;                         /* The handle of the task to which the rest of the information in the structure relates. */
     const char * pcTaskName;                      /* A pointer to the task's name.  This value will be invalid if the task was deleted since the structure was populated! */
-    UBaseType_t xTaskNumber;                      /* A number unique to the task. */
+    UBaseType_t xTaskNumber;                      /* A number unique to the task. Note that this is not the task number that may be modified using vTaskSetTaskNumber() and uxTaskGetTaskNumber(), but a separate TCB-specific and unique identifier automatically assigned on task generation. */
     eTaskState eCurrentState;                     /* The state in which the task existed when the structure was populated. */
     UBaseType_t uxCurrentPriority;                /* The priority at which the task was running (may be inherited) when the structure was populated. */
     UBaseType_t uxBasePriority;                   /* The priority to which the task will return if the task's current priority has been inherited to avoid unbounded priority inversion when obtaining a mutex.  Only valid if configUSE_MUTEXES is defined as 1 in FreeRTOSConfig.h. */


kar-rahul-aws:
Hi @schilkp 
Thank you for the PR. The documentation on the website will be updated by us, once this PR is reviewed and merged.

include/task.h
```diff
@@ -161,7 +161,7 @@ typedef struct xTASK_STATUS
 {
     TaskHandle_t xHandle;                         /* The handle of the task to which the rest of the information in the structure relates. */
     const char * pcTaskName;                      /* A pointer to the task's name.  This value will be invalid if the task was deleted since the structure was populated! */
-    UBaseType_t xTaskNumber;                      /* A number unique to the task. */
+    UBaseType_t xTaskNumber;                      /* A number unique to the task. Note that this is not the task number that may be modified using vTaskSetTaskNumber() and vTaskGetTaskNumber, but a separate TCB-specific and unique identifier automatically assigned on task generation. */
```
kar-rahul-aws, line 164: 1. uxTaskGetTaskNumber() instead of vTaskGetTaskNumber . 
2 .We should also change `xTaskNumber` to `uxTasknumber` for naming convention, since its of type `UBaseType_t` . I can update this in a separate PR.

schilkp: > uxTaskGetTaskNumber() instead of vTaskGetTaskNumber.

Good catch. Fixed.

> We should also change xTaskNumber to uxTasknumber for naming convention, since its of type UBaseType_t . I can update this in a separate PR.

My original proposal was to rename this member to `uxTCBNumber` which matches the internal implementation. The member `uxTCBNumber` in a TCB is the unique ID for each TCB, while `uxTaskNumber` is the value accessed using the above mentioned functions.

It was my understanding from  some posts buried in my discussion (here)[https://forums.freertos.org/t/tracing-improvements/20097/7] that changing the name was not an option because it broken compatibility/kernel-aware debuggers.

If a rename is on the table, I would push for making this much less confusing :)

kar-rahul-aws: I saw this , thanks , yes it will be breaking change in some of the ports.

kar-rahul-aws (approved):

aggarg (approved):

PR #1083, opened 2024-06-05
Give queue set queues a unique type number.

schilkp:
Description
-----------

Changes the value of queueQUEUE_TYPE_SET to a unique value (5) to allow tracers to differentiate between normal queues and queue sets.

Test Steps
-----------
N/A

Checklist:
----------

Related Issue
-----------


schilkp:
On this note - A comment above the definition of the different queue types in `queue.h` states:

```c
/* For internal use only.  These definitions *must* match those in queue.c. */
```

Is this still correct? Where are the definitions in `queue.c` that need to stay in sync? As far as I can tell, `queue.c` also just uses these definitions? Or am I misunderstanding something?

Thanks!

kar-rahul-aws (approved):

shubnil (approved):

PR #1093, opened 2024-06-19
Remove run state assertion in prvCheckForRunStateChange

chinglee-iot:
Description
-----------
Address the problem in this [thread](https://forums.freertos.org/t/smp-porting-checklist-a53-4-as-reference/14499/29).

https://github.com/FreeRTOS/FreeRTOS-Kernel/blob/76eb44382178237197a2c4982f9d99e8d64c9599/tasks.c#L846-L853
In `prvCheckForRunStateChange()`, enabling interrupts should cause this core to immediately service the pending interrupt and yield. Upon the next scheduling of the task, the assertion `configASSERT(pxThisTCB->xTaskRunState != taskTASK_SCHEDULED_TO_YIELD);` may not be true, as other cores could have requested a yield for this task before it evaluates its run state within the assertion. To address this, the task re-evaluates its run state in critical section within a loop until it is eligible for execution, which is the current implementation. Consequently, this assertion should be removed to ensure correct behavior.

In this PR:
* Remove run state assertion in `prvCheckForRunStateChange()` as it doesn't account for the possibility that the task run state could be modified by another task after interrupt is enabled and checking the run state of a task should be performed in a critical section.

Test Steps
-----------
Forum user helps to test this on a Cortex A53 SMP port, with this PR. The main_full demo can run for 5 hours without problem.

Checklist:
----------

- [x] I have modified and/or added unit-tests to cover the code changes in this Pull Request. - https://github.com/FreeRTOS/FreeRTOS/pull/1231

Related Issue
-----------
[Forum thread](https://forums.freertos.org/t/smp-porting-checklist-a53-4-as-reference/14499/29).

### DIFF
--- a/tasks.c
+++ b/tasks.c
@@ -845,10 +845,11 @@ static void prvAddNewTaskToReadyList( TCB_t * pxNewTCB ) PRIVILEGED_FUNCTION;
 
             portENABLE_INTERRUPTS();
 
-            /* Enabling interrupts should cause this core to immediately
-             * service the pending interrupt and yield. If the run state is still
-             * yielding here then that is a problem. */
-            configASSERT( pxThisTCB->xTaskRunState != taskTASK_SCHEDULED_TO_YIELD );
+            /* Enabling interrupts should cause this core to immediately service
+             * the pending interrupt and yield. After servicing the pending interrupt,
+             * the task needs to re-evaluate its run state within this loop, as
+             * other cores may have requested this task to yield, potentially altering
+             * its run state. */
 
             portDISABLE_INTERRUPTS();
             portGET_TASK_LOCK();


aggarg (approved):

ActoryOu (approved):

Saiiijchan:
Excuse me, I am confused with this patch.

> Upon the next scheduling of the task, the assertion configASSERT(pxThisTCB->xTaskRunState != taskTASK_SCHEDULED_TO_YIELD); may not be true, as other cores could have requested a yield for this task before it evaluates its run state within the assertion.

Why not it may be true, if other cores request a yield, it will send a irq to this core firstly and change its task run state.

`    #define prvYieldCore( xCoreID )                                                          \
        else                                                                                 \
        {                                                                                    \
            /* Request other core to yield if it is not requested before. */                 \
            if( pxCurrentTCBs[ ( xCoreID ) ]->xTaskRunState != taskTASK_SCHEDULED_TO_YIELD ) \
            {                                                                                \
                portYIELD_CORE( xCoreID );                                                   \
                pxCurrentTCBs[ ( xCoreID ) ]->xTaskRunState = taskTASK_SCHEDULED_TO_YIELD;   \
            }                                                                                \
        }                                                                                    \
    } while( 0 )`

 Does it mean that the irq is not be responded if run state is taskTASK_SCHEDULED_TO_YIELD？

chinglee-iot:
@Saiiijchan 
I think the assumption for this configASSERT is that the core which is requested to yield will yield immediately without any latency. However, on some platforms, the core can still execute some instructions before it jumps to the interrupt handler for the yield request. The proper way to check the run state should be performed in a critical section and in a loop.

PR #1096, opened 2024-06-26
Update XMOS xcore.ai port to be compatible with v11.x

ACascarino:
### Update XMOS xcore.ai port

Description
-----------

The merge of the `smp` branch into `main` in version 11 broke an assumption that the xcore.ai port relied upon - the backwards-compatibility-preserving decision to conditionally define in tasks.c either pxCurrentTCB or pxCurrentTCBs depending solely on the value of configNUMBER_OF_CORES means that the pxCurrentTCBs symbol does not always exist. The assumption was made that this symbol would always exist, and for FreeRTOS instances with one core in use its 0th element would solely be populated. This change means that ports to SMP platforms need instead to be aware of the number of cores defined and change behaviour accordingly. This went unnoticed at XMOS at the time of release of v11, but we've now noticed the issue and I've (hopefully) created a fix.

This PR attempts to make the XMOS xcore.ai port agnostic to whether it is running on a single-core or SMP instance of FreeRTOS by simply introducing an additional layer of indirection to pxCurrentTCB(s) accesses. When the scheduler is started (which is the first time in the application that the port layer needs to interact with the TCB pointer(s)), we populate a global symbol with the address of the TCB pointer. It is this symbol, rather than pxCurrentTCB(s), which is then used in scheduler initialisation and on context switches.

This PR adds one instruction (plus compiler-defined register spill/restoration to protect the r5 clobber) to scheduler initialisation, and zero instructions to RTOS interrupt processing and context switching.

This PR also includes a fix to make exception behaviour entirely predictable; the previous implementation had assumed that the symbol _TrapHandler would always exist at 0x80080. This is not the case; for example, using the `--first` option in our toolchain to place arbitrary data at the top of memory shifts all other symbols, breaking this assumption and causing wildly undefined behaviour on exception. _DoException is a link-time visible symbol that presents the appropriate entry-point to the platform's exception handling routines, and its use here is preferable in all contexts.

Test Steps
-----------

As this is a community-supported port, there are no automated tests to highlight the issue, nor any to prove lack of regression. However, internal testing of FreeRTOS-based products has shown no regression in their functionality due to this change.

_As a query - partner-maintained ports seem to have a series of automated tests that they use to show lack of regression, but these, from a cursory look, seem entirely single-core focussed. Are there any partner-maintained SMP ports for which the test suite would exercise the additional SMP-related functionality?_

Checklist:
----------

- [ ] I have tested my changes. No regression in existing tests. _There are no existing tests, but implementation testing shows no regression in selected FreeRTOS-based products_
- [ ] I have modified and/or added unit-tests to cover the code changes in this Pull Request. _No unit tests exist_

Related Issue
-----------


AniruddhaKanhere:
Hello @ACascarino,

Thank you for taking the time to report and fix the issue. To me, in the first glance, it looks good.
I'll pass it along to our experts in the team for review.

Thanks,
Aniruddha

n9wxu (approved):

AniruddhaKanhere (dismissed):

portable/ThirdParty/xClang/XCOREAI/portasm.S
```diff
@@ -14,11 +14,7 @@ rest of the ISR callback functions. */
 .issue_mode dual
 .cc_top kexcept.function, kexcept
 kexcept:
-  ldc r11, 0x0008
-  shl r11, r11, 16
-  ldc r9, 0x0080
-  or  r11, r11, r9
-  bau r11 //_TrapHandler is at 0x00080080. TODO: Is it always? Why can't I access the symbol _TrapHandler?
+  bu _DoException
```
AniruddhaKanhere, line 17: I am not sure where this is branching to. Can you please point that out for me?

aggarg (approved):

portable/ThirdParty/xClang/XCOREAI/portasm.S
ACascarino: Hi @AniruddhaKanhere - this is a symbol that is generated by our toolchain on compilation. There are a couple of these which are automatically built into every xcore.ai executable to bring up the processor, set execution modes on each core, and provide routines for graceful exception handling (such as _DoException, which disables all events and interrupts and leaves the processor in a state in which an external debugger may connect and query the processor state at time of exception). Therefore, while it would not be expected to find this symbol defined in any application code, it will always be present in the final executable.

ACascarino:
Apologies - it turns out adding a new commit dismissed stale reviews. Sorry to ask you to review again!

AniruddhaKanhere (approved):

n9wxu (approved):

chinglee-iot:
@ACascarino 
>As a query - partner-maintained ports seem to have a series of automated tests that they use to show lack of regression, but these, from a cursory look, seem entirely single-core focussed. Are there any partner-maintained SMP ports for which the test suite would exercise the additional SMP-related functionality?

We have on target test cases which could exercise the SMP-related functionality. We are still in the progress to publish all the test cases. You could reference reference [this link](https://github.com/FreeRTOS/FreeRTOS/tree/main/FreeRTOS/Test/Target/tests/smp) for more information.

ACascarino:
> @ACascarino
> 
> > As a query - partner-maintained ports seem to have a series of automated tests that they use to show lack of regression, but these, from a cursory look, seem entirely single-core focussed. Are there any partner-maintained SMP ports for which the test suite would exercise the additional SMP-related functionality?
> 
> We have on target test cases which could exercise the SMP-related functionality. We are still in the progress to publish all the test cases. You could reference reference [this link](https://github.com/FreeRTOS/FreeRTOS/tree/main/FreeRTOS/Test/Target/tests/smp) for more information.

Ah great thank you - I managed to completely miss these!

PR #1105, opened 2024-07-28
Small optimisation of scheduler start on Cortex-M7

HazardyKnusperkeks:
Description
-----------
Setting some registers directly, since all meaningful bits are overwritten anyway.

Test Steps
-----------
_Does not apply._

Checklist:
----------

- [X] I have tested my changes. ~~No regression in existing tests.~~

Related Issue
-----------

### DIFF
--- a/portable/GCC/ARM_CM7/r0p1/port.c
+++ b/portable/GCC/ARM_CM7/r0p1/port.c
@@ -419,8 +419,7 @@ BaseType_t xPortStartScheduler( void )
 
     /* Make PendSV and SysTick the lowest priority interrupts, and make SVCall
      * the highest priority. */
-    portNVIC_SHPR3_REG |= portNVIC_PENDSV_PRI;
-    portNVIC_SHPR3_REG |= portNVIC_SYSTICK_PRI;
+    portNVIC_SHPR3_REG |= portNVIC_PENDSV_PRI | portNVIC_SYSTICK_PRI;
     portNVIC_SHPR2_REG = 0;
 
     /* Start the timer that generates the tick ISR.  Interrupts are disabled


aggarg:
The documentation does not say whether the reserved bits in SHPR3 are RAZ/WI - https://developer.arm.com/documentation/dui0646/c/Cortex-M7-Peripherals/System-control-block/System-Handler-Priority-Registers. So not sure we should make this assumption. Besides, this code is executed once at startup, so not much gain in terms of performance either.

portable/GCC/ARM_CM7/r0p1/port.c
```diff
@@ -828,9 +827,7 @@ static void vPortEnableVFP( void )
     __asm volatile
     (
         "   ldr.w r0, =0xE000ED88       \n" /* The FPU enable bits are in the CPACR. */
-        "   ldr r1, [r0]                \n"
-        "                               \n"
-        "   orr r1, r1, #( 0xf << 20 )  \n" /* Enable CP10 and CP11 coprocessors, then save back. */
+        "   mov.w r1, #( 0xf << 20 )    \n" /* Enable CP10 and CP11 coprocessors, and save. */
```
kstribrnAmzn, line 830: Before the code was loading the value of the CPACR and then setting bits 23->20. This new code looks to directly assign register R1 to `0x00F00000`. Wouldn't this be dropping the [access privileges for the coprocessor](https://developer.arm.com/documentation/dui0646/c/Cortex-M7-Peripherals/Floating-Point-Unit/Coprocessor-Access-Control-Register)?

HazardyKnusperkeks: I don't understand. Before it loaded CPACR into R1, set the bits, and store it again into CPACR. This skips the load, since all other bits are unused. (And according to my CM 7 programming manual from ST 0.)

HazardyKnusperkeks:
> The documentation does not say whether the reserved bits in SHPR3 are RAZ/WI - https://developer.arm.com/documentation/dui0646/c/Cortex-M7-Peripherals/System-control-block/System-Handler-Priority-Registers. So not sure we should make this assumption. Besides, this code is executed once at startup, so not much gain in terms of performance either.

The CM7 programming manual from ST says that they are. And as stated in the commit message, I'm just doing what was done with SHPR2 already.

portable/GCC/ARM_CM7/r0p1/port.c
kstribrnAmzn: You're right! I should really read the docs which I link better

kstribrnAmzn (dismissed):

aggarg:
I found more information as I started testing this change. The debugger shows that the lowest 8-bits of SHPR3 are used to set the priority of DebugMonitor - 

image

The following post from ST also seems to confirm the same - https://community.st.com/t5/stm32cubemx-mcus/bug-cubemx-allow-to-set-priority-for-debugmonitor-irqn-but/td-p/577779. Note that this is not the case with SHPR2 as all bits other than SVCall priority bits are reserved.

The debugger shows several bits in CPACR are used for other co-processors -

image

One of the possible explanations is that implementations are allowed to use these reserved bits. I am reaching out to ARM to confirm.

jefftenney:
I can't find any documentation indicating CP0 through CP7 are reserved.  Thus software should preserve them when changing CP10 and CP11.  I suspect Gaurav is right that implementations use them for implementation-specific coprocessors.

image

urutva:
@aggarg [Armv7-M architecture manual](https://developer.arm.com/documentation/ddi0403/latest/) defines SHPR3[7:0] as `Priority of system handler 12, DebugMonitor.` which is what is seen on the debugger and similar for `CPACR` as shown by @jefftenney in https://github.com/FreeRTOS/FreeRTOS-Kernel/pull/1105#issuecomment-2259055356. I believe the debugger is showing the register description as per Armv7-M architecture manual.

@jefftenney The Arm Cortex-M7 Devices Generic User Guide r1p2 shows the [CPACR[19:0] as reserved](https://developer.arm.com/documentation/dui0646/c/Cortex-M7-Peripherals/Floating-Point-Unit/Coprocessor-Access-Control-Register?lang=en). Same document shows [SHPR3[7:0] as reserved](https://developer.arm.com/documentation/dui0646/c/Cortex-M7-Peripherals/System-control-block/System-Handler-Priority-Registers?lang=en).

I'm checking internally if the MCU implementation (Cortex-M7) can mark the register bit-fields defined by the architecture as `reserved`. I'll provide an update soon.

HazardyKnusperkeks:
Fair enough, but one read modify write can be dropped.

(Wrote this yesterday, but apparently did not submit it...)

urutva:
Hi, the debug monitor priority field in SHPR3 is always implemented in v7-M and v8-M Mainline MCUs. Having said that, if the debug monitor isn't being used technically there's no problem in writing 0 to its priority field. Although its technically allowed from an architecture point of view, it may not be a good idea in this case since the Cortex-M7 port is a generic port for all MCUs based on Cortex-M7. For example, when someone starts using the debug monitor, they might be surprised why the scheduler startup code is resetting the priority to 0. Therefore, I'd recommend not to modify the current implementation which conforms to Armv7-M architecture manual. Thanks.

HazardyKnusperkeks:
In my current solution it is not changed.

aggarg:
The current update changes:
```
    portNVIC_SHPR3_REG |= portNVIC_PENDSV_PRI;
    portNVIC_SHPR3_REG |= portNVIC_SYSTICK_PRI;
```
to
```
    portNVIC_SHPR3_REG |= ( portNVIC_PENDSV_PRI | portNVIC_SYSTICK_PRI );
```

This should be okay as there is no functional change. I'd like to keep the code consistent across all the ports. Are you willing to make the same change to all other ports or do you want us to? Other option is to leave it as it is because this code is executed once on startup and it is not a big optimization.

aggarg:
@HazardyKnusperkeks What do you think about the above comment?

HazardyKnusperkeks:
I'll look into the other ports, but don't know when I find the time.

aggarg:
Thank you. I am closing this PR. Please re-open or create a new one when you are ready with the changes.

PR #1170, opened 2024-10-31
Fix RP2040 build without configASSERT defined

jackwilsdon:
Description
-----------
Fixes building the RP2040 port without `configASSERT` defined. It seems like no other ports use `configASSERT` in `portmacro.h`, so this isn't an issue anywhere else. Normally `FreeRTOS.h` defines `configASSERT`, but that can't be included in `portmacro.h`.

This may be a bit easier to review with whitespace changes ignored: https://github.com/FreeRTOS/FreeRTOS-Kernel/pull/1170/files?w=1

Test Steps
-----------
Build the RP2040 port without `configASSERT` defined.

Checklist:
----------

Related Issue
-----------

### DIFF
--- a/portable/ThirdParty/GCC/RP2040/include/portmacro.h
+++ b/portable/ThirdParty/GCC/RP2040/include/portmacro.h
@@ -215,12 +215,12 @@ static inline void vPortRecursiveLock( uint32_t ulLockNum,
 
     #ifdef configASSERT
         configASSERT( ulLockNum < portRTOS_SPINLOCK_COUNT );
-    #endif
+    #endif /* configASSERT */
     uint32_t ulCoreNum = get_core_num();
     uint32_t ulLockBit = 1u << ulLockNum;
     #ifdef configASSERT
         configASSERT( ulLockBit < 256u );
-    #endif
+    #endif /* configASSERT */
 
     if( uxAcquire )
     {
@@ -230,7 +230,7 @@ static inline void vPortRecursiveLock( uint32_t ulLockNum,
             {
                 #ifdef configASSERT
                     configASSERT( ucRecursionCountByLock[ ulLockNum ] != 255u );
-                #endif
+                #endif /* configASSERT */
                 ucRecursionCountByLock[ ulLockNum ]++;
                 return;
             }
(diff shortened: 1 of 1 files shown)


kar-rahul-aws (approved):

ActoryOu:
Hello @jackwilsdon,
Could you help share the compile error log when configASSERT is not defined without this PR?
The fix here is inside a inline function. It should be no issue If the caller also includes the FreeRTOS.h (before portmacro.h).

Thank you.

jackwilsdon:
Macros are evaluated when the inline function is defined, not where it is inlined to: https://godbolt.org/z/haGboYsnE

```
In file included from vendor/FreeRTOS-Kernel/include/portable.h:53,
                 from vendor/FreeRTOS-Kernel/include/FreeRTOS.h:107,
                 from src/main.c:1:
vendor/FreeRTOS-Kernel/portable/ThirdParty/GCC/RP2040/include/portmacro.h: In function 'vPortRecursiveLock':
vendor/FreeRTOS-Kernel/portable/ThirdParty/GCC/RP2040/include/portmacro.h:216:5: warning: implicit declaration of function 'configASSERT' [-Wimplicit-function-declaration]
  216 |     configASSERT( ulLockNum < portRTOS_SPINLOCK_COUNT );
      |     ^~~~~~~~~~~~
```

```
/usr/lib/gcc/arm-none-eabi/13.2.1/../../../arm-none-eabi/bin/ld: vendor/FreeRTOS-Kernel/tasks.c.obj: in function `vPortRecursiveLock':
vendor/FreeRTOS-Kernel/portable/ThirdParty/GCC/RP2040/include/portmacro.h:216:(.text.prvCheckForRunStateChange+0x30): undefined reference to `configASSERT'
```

ActoryOu:
Hi @jackwilsdon,
May I know if configASSERT definition in your FreeRTOSConfig.h?
If there is no definition, would it be easier to define it to nothing like below if you don't need it?
```
#define configASSERT( x )
```

Thank you.

jackwilsdon:
I'm already defining an empty `configASSERT` as a workaround - it'd be nice if it was consistent with other ports though.

ActoryOu:
Hi @jackwilsdon,
Thanks for your patient! We had a discussion about this topic.
We'll create a separate PR to move configASSERT definition above before including `portable.h`.
Then we don't need to handle this in the port layer anymore!
Will make a note here once the PR is created.

Thank you.

PR #1262, opened 2025-04-02
Fix race in POSIX port `vPortEndScheduler`

arctic-alpaca:
Description
-----------

Hi, after upgrading from V11.1.0 to V11.2.0, we saw the call to `vPortEndScheduler` sometimes return when using the POSIX port. We are only calling `vPortEndScheduler` from FreeRTOS tasks (threads in the POSIX port), so we don't expect the call to actually return.

There seems to be a race between the check whether the thread calling `vPortEndScheduler` is a FreeRTOS thread and the destruction of the thread key. If the thread key is destroyed before the check, the check fails even for FreeRTOS threads. 

Moving the check before the signal to end the scheduler should prevent the race.

Test Steps
-----------

Checklist:
----------

Related Issue
-----------

### DIFF
--- a/portable/ThirdParty/GCC/Posix/port.c
+++ b/portable/ThirdParty/GCC/Posix/port.c
@@ -48,8 +48,8 @@
 * stdio (printf() and friends) should be called from a single task
 * only or serialized with a FreeRTOS primitive such as a binary
 * semaphore or mutex.
-* 
-* Note: When using LLDB (the default debugger on macOS) with this port, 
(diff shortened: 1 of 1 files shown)


aggarg (approved):

kar-rahul-aws (approved):

ForsakenHarmony:
Is there any chance to get a `v11.2.1` or `v11.3.0` release with this fix?

PR #1281, opened 2025-06-05
Signing SBOMs in workflows

lefosg:
To provide transparency in the supply chain of the release, cosign (a client for [Sigstore](https://www.sigstore.dev/)) is used to sign the [SBOM](https://edu.chainguard.dev/open-source/sbom/sboms-and-attestations/). It is now common to provide signed artifacts and later on, before deployment, verify the SBOM's signature. This ensures integrity and authenticity (the SBOM has not been tampered with and, only authorized entities - the GitHub runner - produces/signs the SBOM)
-----------
Changes

Added two steps in the `auto-release.yml`, right after generating the SBOM, before commiting the changes.
-----------

It is easy to reproduce locally. Install `cosign`. Download the release. Run the two commands added in the `attest sbom` step. The first command (cosign sign-blob) should produce two files. As a sanity check, the second command verifies the signature, plus that the entity that signed is indeed the GitHub runner.

Checklist:
----------

Related Issue
-----------

### DIFF
--- a/.github/workflows/coverity_scan.yml
+++ b/.github/workflows/coverity_scan.yml
@@ -42,11 +42,22 @@ jobs:
           # ${{ env.stepName }}
           echo -e "::group::${{ env.bashInfo }} ${{ env.stepName }} ${{ env.bashEnd }}"
 
-          wget -nv -qO- https://scan.coverity.com/download/linux64 --post-data "token=${COVERITY_TOKEN}&project=FreeRTOS-Kernel" | tar -zx --one-top-level=cov_scan --strip-components 1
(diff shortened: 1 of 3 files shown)


ActoryOu:
Hi @lefosg,
Thank you for bringing this to us! I'll reach the team to review this.

kstribrnAmzn:
Thank you for the contribution @lefosg ! As a matter of security we are typically weary of externally owned Github actions - especially ones that themselves are helping implement security. I understand the need for signing our releases. Would you be open to the idea of us signing our entire release (including SBOM) or us signing our own SBOM (without cosign)?

PR #1287, opened 2025-07-01
Add Skip List implementation and unit test for improved scheduler determinism

OmniManAintGotNothingOnMe:
**Add Skip List implementation and unit test for improved scheduler determinism**

Description
-----------
This PR introduces a Skip List data structure to the FreeRTOS kernel, providing O(log n) insertion and search for sorted lists. It includes:
- skip_list.h and skip_list.c implementing the Skip List.
- test_skip_list.c as a unit test for the new data structure.
This enhancement is intended to improve temporal determinism for scheduler operations that previously relied on O(n) list traversal.

Test Steps
-----------
1. Build and run test_skip_list.c on a desktop environment to verify correct Skip List behavior.
2. (Optional) Integrate with FreeRTOS scheduler lists and verify no regression in task scheduling.

Checklist:
----------

Related Issue
-----------
N/A


test_skip_list.exe
aggarg, line 1: Why is this binary file included?

aggarg:
This appears to be a standalone implementation that is not utilized by the kernel. If you intend to use it in an application, you should be able to do so without issue. However, given that it is not integrated into the kernel's functionality, I don't believe it should be merged into the kernel codebase.

PR #1282, opened 2025-06-13
Fix/free_rtos

usman-pak1991:
Description
-----------

Test Steps
-----------

Checklist:
----------

- [ ] I have tested my changes. No regression in existing tests.
- [ ] I have modified and/or added unit-tests to cover the code changes in this Pull Request.

Related Issue
-----------

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #1302, opened 2025-07-31
Move windows headers to port.c

creiter64:
Move windows header files to port.c

Description
-----------
This prevents the inclusion of windows.h. into all header files using FreeRTOS.h and thus defining several macros conflicting with common definitions. This is a more minimal fix compared to https://github.com/FreeRTOS/FreeRTOS-Kernel/pull/1282

Test Steps
-----------
None, just a build issue.

Checklist:
----------

Related Issue
-----------

### DIFF
--- a/portable/MSVC-MingW/portmacro.h
+++ b/portable/MSVC-MingW/portmacro.h
@@ -29,17 +29,6 @@
 #ifndef PORTMACRO_H
 #define PORTMACRO_H
 
-#ifdef WIN32_LEAN_AND_MEAN
-    #include <winsock2.h>
-#else
-    #include <winsock.h>
-#endif
-
-#include <windows.h>
-#include <timeapi.h>
-#include <mmsystem.h>
-#include <winbase.h>
-
 #ifdef __cplusplus
 extern "C" {
 #endif
--- a/portable/MSVC-MingW/port.c
+++ b/portable/MSVC-MingW/port.c
@@ -33,6 +33,14 @@
 #include "FreeRTOS.h"
 #include "task.h"
 
+#ifdef WIN32_LEAN_AND_MEAN
+    #include <winsock2.h>
+#else
+    #include <winsock.h>
+#endif
+
+#include <timeapi.h>
+
 #ifdef __GNUC__
     #include "mmsystem.h"
 #else


creiter64:
I have removed some of the overly complicated task selection code. It would be good if someone knowledgeable could review it.

portable/MSVC-MingW/portmacro.h
```diff
@@ -146,36 +134,14 @@ void vPortExitCritical( void );
...
-
-    #endif /* __GNUC__ */
+    #define portGET_HIGHEST_PRIORITY( uxTopPriority, uxReadyPriorities ) \
+    do {                                                                 \
+        unsigned i = sizeof(UBaseType_t) * 8 - 1;                        \
+        while ((uxReadyPriorities & ((UBaseType_t)1 << i)) == 0) {       \
+            i--;                                                         \
+        }                                                                \
+        uxTopPriority = i;                                               \
+    } while( 0 )
```
ActoryOu, line 144: Why do you suggest to make this change? In my understanding the performance of assembly one should be better.

creiter64: Mostly for portability reasons, so that this can be compiled with every C99 compliant compiler.

billybednar: The kernel already has portable task selection logic that is used when `configUSE_PORT_OPTIMISED_TASK_SELECTION` is disabled. The macros here are used to provide a fast but non-portable alternative to that. With this change, the port-optimized version becomes slower than the generic version.

aggarg:
> This prevents the inclusion of windows.h. into all header files using FreeRTOS.h and thus defining several macros conflicting with common definitions.

Would you please share the conflicting definitions?

creiter64:
>Would you please share the conflicting definitions?

In my case, `windows.h` included `winbase.h` which in turn contains a macro defined as `BAUD_9600`, which conflicted with an enum in my code.

aggarg:
> In my case, windows.h included winbase.h which in turn contains a macro defined as BAUD_9600, which conflicted with an enum in my code.

Is it not possible to change the name of this enum?

creiter64:
>Is it not possible to change the name of this enum?

No. In addition I'd argue that it's bad practice to force the user to know all the ways `windows.h` messes with their defines on a cross-platform OS like FreeRTOS.

aggarg:
> In addition I'd argue that it's bad practice to force the user to know all the ways windows.h messes with their defines on a cross-platform OS like FreeRTOS.

If you want to run your application on Windows, using a name that is used by Windows OS can lead to potential conflicts. For example, what if a future Windows update moves this define to `intrin.h`, which could break your application.

I assume that you have constraints preventing you from renaming the enum, but I would recommended renaming it for robustness.

creiter64:
I'm not using anything from the Windows API so I was suprised that FreeRTOS was publicly including it. The windows build is for unit testing of an embedded firmware. I try to keep that standard C compliant as possible.

aggarg (approved):

PR #1311, opened 2025-08-15
Fix risk of Win32 timer setup call getting ignored

kschwan:
Description
-----------
If a user configures the `configASSERT` macro to expand to nothing, a call to the Win32 API `SetWaitableTimer()` doesn't get compiled.

This can happen if, for example, `configASSERT` set defined as `assert` (from assert.h) which expands to nothing when `NDEBUG` is set (common for "release" builds).

This PR has a patch to fix the issue.

Test Steps
-----------
Using the Win32 port, define `configASSERT` as follows:
```
#define configASSERT( x ) ( (void)0 )
```
then the scheduler doesn't "tick" and calls to, fx., `vTaskDelay()` won't return.

Checklist:
----------


kschwan:
I'm not sure how to run tests - please let me know if I should do anything in that regard.

xuelix (approved):

PR #1317, opened 2025-09-08
fix[RL78 Port] incorrect register image for pvParameters in FAR mode (#1316)

sauleThsQuin:
### Title
fix[RL78 Port] incorrect register image for pvParameters in FAR mode (#1316)

### Description
In the RL78 FAR data model, pxPortInitialiseStack() did not correctly
initialize the register image in the task stack for the parameter (pvParameters).

The saved A:DE registers were filled with dummy values instead of the actual pointer value.

As a result, when the first context restore occurred, the compiler's function prologue read a corrupted parameter from the register image.

This only affected FAR builds; NEAR model was not impacted.

This patch updates the RL78 FAR path so that the register image matches the RL78 calling convention:
DE = low 16 bits of pvParameters
A = high 8 bits of pvParameters (via AX register image, X = 0)

With this change, tasks in FAR model receive the correct parameter at entry, while NEAR model remains unchanged.

### Test Steps
- Build FreeRTOS for RL78 with FAR data model.
- Create a task with xTaskCreate(task, "name", stack, pvParameters, prio, NULL).
- Inspect pvParameters inside the task:

**Before fix:** value corrupted (dummy registers A:DE = 0x1111/0xDEDE).

**After fix:** value matches the pointer passed to xTaskCreate.

Confirm no regression with NEAR model build.

### Checklist
- [x]  I have tested my changes. No regression in existing tests.

- [ ]  I have modified and/or added unit-tests to cover the code changes in this Pull Request.

### Related Issue
 #1316

### DIFF
--- a/portable/IAR/RL78/port.c
+++ b/portable/IAR/RL78/port.c
@@ -95,10 +95,10 @@ StackType_t * pxPortInitialiseStack( StackType_t * pxTopOfStack,
     uint32_t * pulLocal;
 
     /* With large code and large data sizeof( StackType_t ) == 2, and
-    * sizeof( StackType_t * ) == 4.  With small code and small data
-    * sizeof( StackType_t ) == 2 and sizeof( StackType_t * ) == 2. */
+     * sizeof( StackType_t * ) == 4.  With small code and small data
+     * sizeof( StackType_t ) == 2 and sizeof( StackType_t * ) == 2. */
 
-    #if __DATA_MODEL__ == __DATA_MODEL_FAR__
+#if __DATA_MODEL__ == __DATA_MODEL_FAR__
     {
         /* Far pointer parameters are passed using the A:DE registers (24-bit).
          * Although they are stored in memory as a 32-bit value.  Hence decrement
@@ -124,13 +124,44 @@ StackType_t * pxPortInitialiseStack( StackType_t * pxTopOfStack,
         *pulLocal = ( ( ( uint32_t ) pxCode ) | ( portPSW << 24UL ) );
         pxTopOfStack--;
 
-        /* An initial value for the AX register. */
-        *pxTopOfStack = ( StackType_t ) 0x1111;
+        /* In FAR data model, the compiler expects the task parameter (pvParameters)
+         * to be passed in the A:DE registers (24-bit). Therefore the parameter is
+         * split into:
+         *   - DE = low 16 bits
+         *   - A  = high 8 bits (loaded through the AX register image, X = 0)
+         * This ensures that the task entry function receives pvParameters in the
+         * format required by the C calling convention. */
+        {
+            uint32_t p = ( uint32_t ) pvParameters;
+            /* low 16 bits */
+            uint16_t de_init = (uint16_t)( p & 0xFFFFU );
+            /* A = high byte, X = 0 */   
+            uint16_t ax_init = (uint16_t)( ((p >> 16) & 0xFFU) << 8 );  
+
+            /* AX register image */
+            *pxTopOfStack = ( StackType_t ) ax_init;   
+            pxTopOfStack--;
+
+            /* HL register image (dummy) */
+            *pxTopOfStack = ( StackType_t ) 0x2222;    
+            pxTopOfStack--;
+
+            /* CS:ES register image */
+            *pxTopOfStack = ( StackType_t ) 0x0F00;    
+            pxTopOfStack--;
+
+            /* DE register image */
+            *pxTopOfStack = ( StackType_t ) de_init;   
+            pxTopOfStack--;
+        }
+
+        /* BC remains a dummy value (not used for parameter passing). */
+        *pxTopOfStack = ( StackType_t ) 0xBCBC;
         pxTopOfStack--;
     }
-    #else /* if __DATA_MODEL__ == __DATA_MODEL_FAR__ */
+#else /* if __DATA_MODEL__ == __DATA_MODEL_FAR__ */
     {
-        /* The return address, leaving space for the first two bytes of the
+       /* The return address, leaving space for the first two bytes of the
          * 32-bit value.  See the comments above the prvTaskExitError() prototype
          * at the top of this file. */
         pxTopOfStack--;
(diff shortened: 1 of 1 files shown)


aggarg:
@KeitaKashima What do you think about this change?

ghost:
@aggarg san
> @KeitaKashima What do you think about this change?

My team will check the code. Could you please have some time to check it in about one or two weeks?

aggarg:
@KeitaKashima Of course. Take your time. Thank you for your help with the review!

ghost:
@aggarg 
I am sorry that my team needs more time than I expected.

@TakeoTakahashi2020 and @ShunichiroNakamura of my team members, will check and answer the result.

aggarg:
@KeitaKashima No worries at all! Please take your time.

portable/IAR/RL78/port.c
```diff
@@ -95,10 +95,10 @@ StackType_t * pxPortInitialiseStack( StackType_t * pxTopOfStack,
     uint32_t * pulLocal;
 
     /* With large code and large data sizeof( StackType_t ) == 2, and
-    * sizeof( StackType_t * ) == 4.  With small code and small data
-    * sizeof( StackType_t ) == 2 and sizeof( StackType_t * ) == 2. */
+     * sizeof( StackType_t * ) == 4.  With small code and small data
+     * sizeof( StackType_t ) == 2 and sizeof( StackType_t * ) == 2. */
 
-    #if __DATA_MODEL__ == __DATA_MODEL_FAR__
+#if __DATA_MODEL__ == __DATA_MODEL_FAR__
```
ShunichiroNakamura, line 101: Regarding the line 109 (*pulLocal = ( uint32_t ) pvParameters;).  
When using `__DATA_MODEL_FAR__`, is saving the `prvParameters` to the stack to maintain compatibility with IAR compiler V1 and V2?  
If yes, that’s fine, because I understand saving to the A:DE registers is dummy operation in the V1, and saving to the stack is dummy operation in the V2. According to the IAR compiler's calling convention, it seems that the V1 passes the `prvParameters` through the stack whereas the V2 passes it through the A:DE registers.  
If you prefer to refine the code, I propose switching the location where the `prvParameters` is saved by using the predefined macro `__CC_V1__` or `__CC_V2__` (or else).

portable/IAR/RL78/port.c
```diff
@@ -95,10 +95,10 @@ StackType_t * pxPortInitialiseStack( StackType_t * pxTopOfStack,
     uint32_t * pulLocal;
 
     /* With large code and large data sizeof( StackType_t ) == 2, and
```
ShunichiroNakamura, line 97: Strictly speaking, sizeof( StackType_t * ) is affected by `__DATA_MODEL__`, not `__CODE_MODEL__`. So, I propose updating comments (*1):

* "With large code and large data" -> "With large data"
* "With small code and small data" -> "With small data"

Alternatively, I propose clearly explaining effects of `__CODE_MODEL__` and `__DATA_MODEL__` for users (*2). This means adding a comment that the size of argument `pxCode` of pxPortInitialiseStack() is 2 or 4 bytes depending on `__CODE_MODEL__`.  

*1: I guess the reason for mentioning both of code and data was that `__CODE_MODEL__`/`__DATA_MODEL__` combinations were used only [near/near and far/far](https://github.com/FreeRTOS/FreeRTOS-Kernel/blob/dde270bfbcc335a7019e1ca841c921af31c7f261/portable/IAR/RL78/portmacro.h#L50) in the test environments.

*2: The sizes of arguments for pxPortInitialiseStack() are shown below:

| Code model | Data model | StackType_t *pxTopOfStack | TaskFunction_t pxCode | void *pvParameters |
| ---- | ---- | ---- | ---- | ---- |
| **near** | **near** | 2bytes | 2bytes | 2bytes |
| near | far | 4bytes | 2bytes | 4bytes |
| far | near | 2bytes | 4bytes | 2bytes |
| **far** | **far** | 4bytes | 4bytes | 4bytes |

ShunichiroNakamura (commented):
@TakeoTakahashi2020 and I checked the code. Thank  you.

aggarg:
@ShunichiroNakamura and @TakeoTakahashi2020, thank you for reviewing the code.

@sauleThsQuin Would you please address the comments?

portable/IAR/RL78/port.c
sauleThsQuin: Thank you for the review and clarification. I was not aware of the V1 vs V2 calling convention difference when I first submitted the patch — my intent was only to fix the issue I had in FAR model. The stack write remained as legacy from the original code. Based on your feedback I’ve updated the patch to explicitly branch on __CC_V2__ so that only the relevant write is kept. This way the fix still solves the FAR bug and is now aligned with both V1 and V2.

ShunichiroNakamura: @sauleThsQuin  
Thank you. But, could you update #if defined(__CC_V2__) to #if __CALLING_CONVENTION__ == __CC_V2__?  

When building 5f740f1098a49c2a29e640ae0faf907a2c9408e4 with IAR EW 5.20.1, it seems that both __CC_V1__ and __CC_V2__ are always defined, so the condition #if defined(__CC_V2__) was not evaluated correctly.  


@aggarg  
Due to my request, the amount of conditional compilation has increased, which has reduced code readability. I’d appreciate any advice you might have from coding style etc.

portable/IAR/RL78/port.c
ShunichiroNakamura: Thank you for the update. No concerns from my side.

portable/IAR/RL78/port.c
sauleThsQuin: Thanks for the feedback. I updated the condition to #if CALLING_CONVENTION == CC_V2 as suggested.

sauleThsQuin: No problem. I’ve fixed it as shown.

ShunichiroNakamura: Thank you. I confirmed it.

ShunichiroNakamura:
@aggarg 
We have confirmed that the changes in portable/IAR/RL78/port.c are no problem.

aggarg:
@ShunichiroNakamura Thank you so much! 

@sauleThsQuin The diff shows changes in a lot of unrelated files. Would you please fix them so that we can merge the PR?

sauleThsQuin:
Thanks for the reviews and feedback.
I appreciate the guidance and I'm glad the fix has been merged.

Issue #1321, opened 2025-09-26
[Feature Request] Comply to reuse spec


rzr:
**Is your feature request related to a problem? Please describe.**

Complying to RUSE spec, can help to create SBOM and other compliance programs

Please confirm that project should be declared as:

```
SPDX-FileCopyrightText = "2021 Amazon.com, Inc. or its affiliates.  All Rights Reserved."
SPDX-License-Identifier = "MIT AND Apache-2.0"
```

**Describe the solution you'd like**

Implement https://reuse.software/ in pipelines, I can share hints if needed

**Additional context**

Also it would make sense to add all licenses in project

For instance this file brings more licenses (previously GPL-2.0):

- FreeRTOS-Kernel/portable/ThirdParty/GCC/Xtensa_ESP32/include/portmacro.h MIT
- FreeRTOS-Kernel/portable/ThirdParty/GCC/Xtensa_ESP32/port.c MIT
- FreeRTOS-Kernel/portable/ThirdParty/GCC/Xtensa_ESP32/portmux_impl.h Apache-2.0
- FreeRTOS-Kernel/portable/ThirdParty/GCC/Xtensa_ESP32/portmux_impl.inc.h Apache-2.0

PR #1322, opened 2025-09-26
fix(legal): Workaround reuse tool

rzr:
I noticed while using https://reuse.software/
this false positive error:

    reuse.extract - ERROR - Could not parse 'MIT\n','
    reuse.extract - ERROR - \
      '.../FreeRTOS-Kernel/.github/scripts/kernel_checker.py' \
      holds an SPDX expression that cannot be parsed, skipping the file

Relate-to: https://github.com/FreeRTOS/FreeRTOS-Kernel/issues/1321

Description
-----------

Test Steps
-----------

Checklist:
----------

Related Issue
-----------

### DIFF
--- a/.github/scripts/kernel_checker.py
+++ b/.github/scripts/kernel_checker.py
@@ -125,7 +125,7 @@ KERNEL_HEADER = [
     ' * FreeRTOS Kernel <DEVELOPMENT BRANCH>\n',
     ' * Copyright (C) 2021 Amazon.com, Inc. or its affiliates. All Rights Reserved.\n',
     ' *\n',
-    ' * SPDX-License-Identifier: MIT\n',
+    ' * SPDX', '-License-Identifier: MIT\n',
     ' *\n',
     ' * Permission is hereby granted, free of charge, to any person obtaining a copy of\n',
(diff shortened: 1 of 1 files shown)


aggarg:
This line does not need to be parsed by the tool as it is not license. Seems more like a tool limitation.

rzr:
> This line does not need to be parsed by the tool as it is not license. Seems more like a tool limitation.

yes i can confirm , that's why i committed as a workaround
Feel free to ignore if you think it is polluting the code, but it helped me to scan some repository that pull freertos as submodule dependency.

xuelix:
The reuse tool needs to be updated in this case.
Close the PR.

rzr:
> The reuse tool needs to be updated in this case. Close the PR.

ok fair enough, I will try to forward the issue at the source

meanwhile please consider:

https://github.com/FreeRTOS/FreeRTOS-Kernel/issues/1321

rzr:
Is this an acceptable option:

https://github.com/rzr/FreeRTOS-Kernel/commit/a7a7bdb75c738c67edd9bdc5efbc97785d653c8b


Relate-to: https://reuse.software/faq/#exclude-lines

Issue #1337, opened 2025-11-27
[BUG] Incorrect argument to  portRESET_READY_PRIORITY macro in vTaskPriorityDisinheritAfterTimeout()


wirelinker:
**Describe the bug**
I found in `vTaskPriorityDisinheritAfterTimeout(),` tasks.c Line 6909,
after a `uxListRemove()`, there is a  `portRESET_READY_PRIORITY` macro.

The argument be passed to the macro is `pxTCB->uxPriority`, which
should be `uxPriorityUsedOnEntry`.

Because the `pxTCB` was removed from `pxReadyTasksLists[ uxPriorityUsedOnEntry ]`.
Then the  `portRESET_READY_PRIORITY()` will be called if the  `pxReadyTasksLists[uxPriorityUsedOnEntry]` became empty after the remove.
And `pxTCB->uxPriority` had already been changed to another higher/lower task priority.
That is : `uxPriorityUsedOnEntry != pxTCB->uxPriority`.

Thank you.

**Target**
Ports with optimized task selection implementation.

**Host**
- Host OS: Ubuntu
- Version: 24.04

**To Reproduce**
I have no idea how to reproduce. It only effects customized scheduler ports. I am using pi-pico 
as target which is not the case.

**Expected behavior**
Should reset ready priority with the priority of the actually empty ready list.

**Additional context**

kstribrnAmzn:
Wow...this is a great find! I'll raise a PR shortly to update this. I have a way to test this and I can post some results on Monday as I left the test board in the office.

kstribrnAmzn:
PR - https://github.com/FreeRTOS/FreeRTOS-Kernel/pull/1338

wirelinker:
I was wondering about if I misunderstood the code.
Because this bug didn't cause any abnormal behavior on pico.
I don't even know how to reproduce the bug.
So I checked all the appearances of ` portRESET_READY_PRIORITY()` in the file, then do the post.
I am glad to see this is helpful.

kstribrnAmzn:
Sorry to reverse my previous stance here but there was an error in my thinking. I was thinking about this priority disinheritance bug from the perspective of a single task in isolation. When doing this it will appear bugged, but once I thought about it in the context of multiple tasks then the code is correct. Let me see if I can explain and diagram this below.

[...]

PR #1338, opened 2025-11-28
[Fix] Reset correct priority on disinherit timeout

kstribrnAmzn:
Description
-----------
This commit corrects vTaskPriorityDisinheritAfterTimeout to reset the previously inherited priority when the task disiheriting timeout was the only task at that priority. Without this change the ready list for the inherited priority will remain set when no task is ready at that priority. This can have consequences later as the ready priority flags are assumed to be accurate.

Thanks @wirelinker for bringing this to the teams attention!

Test Steps
-----------
Testing this on Monday when I have a dev board handy.

Automated UTs are passing.

Checklist:
----------

  - Note: Bug fix with UT already correct

Related Issue
-----------
https://github.com/FreeRTOS/FreeRTOS-Kernel/issues/1337


kstribrnAmzn:
Closing as this will break correct behavior. See explanation on https://github.com/FreeRTOS/FreeRTOS-Kernel/issues/1337.

kstribrnAmzn:
Reopening after further discussion on https://github.com/FreeRTOS/FreeRTOS-Kernel/issues/1337

jasonpcarroll:
Honestly - we should really rework the variable naming a bit here ... also its a bit confusing as to why we set pxTCB to pxMutexHolder directly... esp since both are used in the function.. makes it hard to understand what is happening.

jasonpcarroll:
Perhaps we should also add some UTs that ensure portRESET_READY_PRIORITY is called with the correct priorities for functions that call it ? I.e. just stub it with our own def that just asserts it is called with the correct priorities when the system is in a known state. It worries me that before/after this change no UTs were added and none failed either - how will we catch things like this in the future?

kstribrnAmzn:
@jasonpcarroll UTs are easy to add but will need to be done in a separate PR since they live in FreeRTOS/FreeRTOS. They do not fail because they do not verify the `portRESET_READY_PRIORITY` macro call. I'll see if I can get these created shortly.

aggarg (approved):

kstribrnAmzn:
I have the UTs ready in https://github.com/FreeRTOS/FreeRTOS/pull/1382. Merging this in now so I can update the submodule so that the UTs pass.

PR #1345, opened 2025-12-20
Win32: Guard timeapi.h include for MinGW compatibility

### LINKED ISSUE #1344: [BUG]Win32 port: timeapi.h include breaks MinGW builds (GCC)
The Win32 simulator port unconditionally includes <timeapi.h>.
This header is available when building with MSVC, but is not
consistently available when building with MinGW-w64 GCC.

As a result, the Win32 port fails to compile with MinGW unless
manual changes are applied.

OS: Windows 11 25h2
Compiler: MinGW-w64 GCC
FreeRTOS-Kernel: v11.x / main
Port: Win32 simulator

fatal error: timeapi.h: No such file or directory

nordync:
Win32 port: fix MinGW build failure caused by timeapi.h

Description
-----------
The Win32 simulator port unconditionally includes <timeapi.h>,
which is available when building with MSVC but not consistently
available when building with MinGW-w64 GCC.

This change guards the include using _MSC_VER and relies on
<mmsystem.h> for MinGW builds, restoring build compatibility
without affecting MSVC builds.

Test Steps
-----------
1. Build FreeRTOS Win32 simulator using MSVC (no regression)
2. Build FreeRTOS Win32 simulator using MinGW-w64 GCC
3. Confirm successful compilation and execution

Checklist:
----------

Related Issue
-----------
Fixes #1344

### DIFF
--- a/include/task.h
+++ b/include/task.h
@@ -897,7 +897,8 @@ void vTaskDelay( const TickType_t xTicksToDelay ) PRIVILEGED_FUNCTION;
  *
  * @return Value which can be used to check whether the task was actually delayed.
  * Will be pdTRUE if the task way delayed and pdFALSE otherwise.  A task will not
- * be delayed if the next expected wake time is in the past.
+ * be delayed if the next expected wake time is in the past. This prevents periodic
+ * tasks from accumulating delays and allows them to resume their regular timing pattern.
  *
  * Example usage:
  * @code{c}


aggarg:
We already have [this code block](https://github.com/FreeRTOS/FreeRTOS-Kernel/blob/main/portable/MSVC-MingW/port.c#L42-L48) which includes `mmsystem.h`. Does it mean that we need to move `#include <timeapi.h>` to the [#else block](https://github.com/FreeRTOS/FreeRTOS-Kernel/blob/main/portable/MSVC-MingW/port.c#L47)?

nordync:
> We already have [this code block](https://github.com/FreeRTOS/FreeRTOS-Kernel/blob/main/portable/MSVC-MingW/port.c#L42-L48) which includes `mmsystem.h`. Does it mean that we need to move `#include <timeapi.h>` to the [#else block](https://github.com/FreeRTOS/FreeRTOS-Kernel/blob/main/portable/MSVC-MingW/port.c#L47)?

Thanks for the feedback.
The previous GCC block is no longer necessary because we’ve reorganized the includes based on compiler:

_MSC_VER (MSVC) uses <timeapi.h> and links winmm.lib.

Other compilers (GCC/MinGW) use <mmsystem.h>.

This ensures cross-compiler compatibility and avoids duplicate includes. The final change removes the old GCC block entirely.

aggarg (approved):

moninom1 (approved):

PR #1361, opened 2026-01-29
Fix undefined behavior in pointer comparison

Tedlion:
Replace direct pointer comparison with integer comparison in prvInsertBlockIntoFreeList() to avoid undefined behavior.

According to the C standard, take C11 standard for instance, there is  following description in Section 6.5.8 Relational operators: "When two pointers are compared, the result depends on the relative locations in the address space of the objects pointed to. If two pointers to object types both point to the same object, or both point one past the last element of the same array object, they compare equal. If the objects pointed to are members of the same aggregate object, pointers to structure members declared later compare greater than pointers to members declared earlier in the structure, and pointers to array elements with larger subscript values compare greater than pointers to elements of the same array with lower subscript values. All pointers to members of the same union object compare equal. If the expression P points to an element of an array object and the expression Q points to the last element of the same array object, the pointer expression Q+1 compares greater than P. In all other cases, the behavior is undefined."

So the comparison "heapPROTECT_BLOCK_POINTER( pxIterator->pxNextFreeBlock ) < pxBlockToInsert" is UB, and a explicit cast will fix that.

Description
-----------

Test Steps
-----------

Checklist:
----------

Related Issue
-----------

### DIFF
--- a/portable/MemMang/heap_5.c
+++ b/portable/MemMang/heap_5.c
@@ -489,7 +489,8 @@ static void prvInsertBlockIntoFreeList( BlockLink_t * pxBlockToInsert ) /* PRIVI
 
     /* Iterate through the list until a block is found that has a higher address
      * than the block being inserted. */
-    for( pxIterator = &xStart; heapPROTECT_BLOCK_POINTER( pxIterator->pxNextFreeBlock ) < pxBlockToInsert; pxIterator = heapPROTECT_BLOCK_POINTER( pxIterator->pxNextFreeBlock ) )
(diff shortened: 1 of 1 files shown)


archigup:
This comparison does not seem like UB to me. Both pointers are from the same memory object (the backing array for the heap). Though using a static array for backing an allocator technically violates strict aliasing, so you may want to compile with a `-fno-strict-aliasing` or equivalent compiler flag.

Tedlion:
"_pointers to array elements with larger subscript values compare greater than pointers to elements of the same array with lower subscript values_". Well, it does not expilicitly require the pointer to be the same type of the array elements. So maybe I took the rule too strictly?

Issue #1349, opened 2025-12-25
[BUG] xTaskGetTickCount strange behavior causes xTaskDelayUntil to always fail


ntd:
I think that my loop starts with a last wake up time (`last`) greater than the current time (`tick`, from `xTaskGetTickCount`) and #1339 does the rest, so `xTaskDelayUntil` becomes a no-op and fails always.

- Development board: ESP32-S3, dual core
- Instruction Set Architecture: XTENSA32 LX7?
- IDE and version: esp-idf 5.5.1
- Toolchain and version: xtensa-esp-elf esp-14.2.0_20241119
- Host OS: Archlinux

**To Reproduce**
```c
#include "freertos/FreeRTOS.h"
#include <stdio.h>

static void loop(void *data)
{
        TickType_t last = xTaskGetTickCount();
        for (int i = 0; ; ++i) {
                if (xTaskDelayUntil(&last, 5) == pdFALSE) {
                        printf("Iteration %d too slow (last=%lu, tick=%lu)\n", i, last, xTaskGetTickCount());
                }
        }
}

void app_main(void)
{
        TaskHandle_t handle;
        xTaskCreate(loop, "TEST", 2048, NULL, 1, &handle);
        /* The following two lines trigger the issue */
        vTaskSuspend(handle);
        vTaskResume(handle);
}
```

With the above code, I get a busy loop that starts with:
```
Iteration 1 too slow (last=11, tick=2)
Iteration 2 too slow (last=16, tick=3)
Iteration 3 too slow (last=21, tick=3)
Iteration 4 too slow (last=26, tick=4)
```

Without `vTaskSuspend/vTaskResume` everything works as expected.

No idea how `last` can be greater than `tick` or if this is an issue specific to the dual core nature of the ESP32-S3 (i.e. the above code works on other architectures).

kstribrnAmzn:
In your `app_main` function, where are you starting the scheduler by calling [vTaskStartScheduler](https://www.freertos.org/Documentation/02-Kernel/04-API-references/04-RTOS-kernel-control/03-vTaskStartScheduler)? You'll need to call this function before you can use `xTaskDelayUntil` OR `xTaskGetTickCount`.

kstribrnAmzn:
You'll want to update your code to the following:

```
void app_main(void)
{
        TaskHandle_t handle;
        xTaskCreate(loop, "TEST", 2048, NULL, 1, &handle);
        vTaskStartScheduler();
[...]

Issue #1339, opened 2025-12-01
[BUG] xTaskDelayUntil updates pxPreviousWakeTime incorrectly in case of deadline miss


INOMO-RKr:
**Describe the bug**
If the targeted wake-up time is already in the past, i.e. `xTaskDelayUntil( &previousWakeUpTime, xTimeIncrement ) == pdFALSE` then `previousWakeUpTime` is incremented by the specied delay ticks, and goes out of sync with the actual ticks.

**Target**
- All

**To Reproduce**
Call the function in a way that the specified deadline is missed.

**Expected behavior**
After a call to `xTaskDelayUnitl(&previousWakeUpTime, xTimeIncrement )` the variable `previousWakeUpTime` holds the current ticks.
according to the functions documentation: 
```
/**
 * @param pxPreviousWakeTime Pointer to a variable that holds the time at which the
 * task was last unblocked.  The variable must be initialised with the current time
 * prior to its first use (see the example below).  Following this the variable is
 * automatically updated within xTaskDelayUntil ().
*/
```

OR
The documentation clearly describes the current behavior.

**Additional context**
***Relevant section in source code***
https://github.com/FreeRTOS/FreeRTOS-Kernel/blob/151fb04ad113632695a571e8cb93146120f13e67/tasks.c#L1231-L1265
```
            /* Generate the tick time at which the task wants to wake. */
            xTimeToWake = *pxPreviousWakeTime + xTimeIncrement;
/*...
Lines omitted for clarity 
...*/
            *pxPreviousWakeTime = xTimeToWake; 
```
it is clear that the update of `*pxPreviousWakeTime` is the same irregardless of if the deadline is missed or not.

***Example***
This causes in the following example the `previousWakeTime` and `xTaskGetTickCount()` to diverge.

```
void exampleTaskFunction(void)
{
    TickType_t previousWakeUpTime = xTaskGetTicks();
    const TickType_t cDelayTicks = 5;
    while(true)
    {
        vTaskDelay(10*cDelayTicks) /*Simple way to force a deadline miss*/
       if(!xTaskDelayUntil(&previousWakeUpTime, cDelayTicks) )
      { /*Always true*/
        assert(xTaskGetTicks() != previousWakeUpTime);
      }
}
````

kstribrnAmzn:
[...]

PR #1368, opened 2026-02-17
New API xTaskPeriodicDelay (#1349)

ntd:
New function to be used for periodic tasks to ensure a constant execution frequency. It is intended to supersede xTaskDelayUntil to overcome its shortcomings, that is:

- avoid run away of pxPreviousWakeTime (#1339)
- catch up any skipped period immediately (and update pxPreviousWakeTime accordingly), notify the caller of the number of periods skipped (by returning them) and wait until the next period (it could be less than xTimeIncrement if we are close to the next period)
- notify the caller when not enough ticks have been elapsed (by returning 0) and handle the situation gracefully (by properly waiting until the next wake time)

Description
-----------

Test Steps
-----------
I tested that function on my ESP-IDF setup, directly running the resulting binary on an ESP32-S3 evaluation board. I would be more than happy to provide a unit test but I simply don't know how to do it.

This is the code I used:
```C
static void loop(void *data)
{
        TickType_t rv;
        TickType_t last = xTaskGetTickCount();
        for (int i = 0; i < 10; ++i) {
                rv = xTaskPeriodicDelay(&last, 5);
                printf("Iteration %d: last=%lu, tick=%lu, rv=%lu\n",
                       i, last, xTaskGetTickCount(), rv);
        }
        vTaskDelete(NULL);
}

void app_main(void)
{
        TaskHandle_t handle;
        xTaskCreate(loop, "LOOP", 2048, NULL, 1, &handle);
        vTaskSuspend(handle);
        vTaskResume(handle);
        vTaskDelay(3);
        vTaskSuspend(handle);
        vTaskResume(handle);
        vTaskDelay(6);
        vTaskSuspend(handle);
        vTaskDelay(12);
        vTaskResume(handle);
}

#if 0

/* This is an approximation of what ESP-IDF is doing behind the scene */
static void main_task(void *data)
{
        printf("main_task: Calling app_main()\n");
        app_main();
        printf("main_task: Returned from app_main()\n");
        vTaskDelete(NULL);
}

void main(void)
{
        TaskHandle_t handle;
        xTaskCreate(main_task, "MAIN", 2048, NULL, 1, &handle);
        vTaskStartScheduler();
}
#endif
```
and here are the results:
```
I (301) main_task: Calling app_main()
Iteration 0: last=1, tick=1, rv=0
Iteration 1: last=1, tick=4, rv=0
Iteration 2: last=1, tick=6, rv=0
I (406) main_task: Returned from app_main()
Iteration 3: last=6, tick=22, rv=1
Iteration 4: last=21, tick=26, rv=3
Iteration 5: last=26, tick=31, rv=1
Iteration 6: last=31, tick=36, rv=1
Iteration 7: last=36, tick=41, rv=1
Iteration 8: last=41, tick=46, rv=1
Iteration 9: last=46, tick=51, rv=1
```

Checklist:
----------

Related Issue
-----------
#1349 

### DIFF
--- a/include/task.h
+++ b/include/task.h
@@ -896,6 +896,43 @@ void vTaskDelay( const TickType_t xTicksToDelay ) PRIVILEGED_FUNCTION;
 /**
  * task. h
  * @code{c}
+ * TickType_t xTaskPeriodicDelay( TickType_t *pxPreviousWakeTime, const TickType_t xTimeIncrement );
+ * @endcode
+ *
+ * INCLUDE_xTaskDelayUntil must be defined as 1 for this function to be available.
+ * See the configuration section for more information.
+ *
+ * Periodic task delay to ensure a constant execution frequency.
+ *
+ * This function is similar to xTaskDelayUntil () with a few important differences:
+ * - pxPreviousWakeTime contains the last past wake time, so it never runs away
+ * - if you suspend the task, when you resume it pxPreviousWakeTime will instantly
+ *   catch up all skipped increments
+ * - it returns the number of increments added to pxPreviosWakeTime
+ *
+ * @param pxPreviousWakeTime Pointer to a variable that holds the time at which the
+ * task was last unblocked.  The variable must be initialised with the current time
+ * prior to its first use.  Following this the variable is automatically updated.
+ *
+ * @param xTimeIncrement The cycle time period.  The task will be unblocked at
+ * time *pxPreviousWakeTime + xTimeIncrement.  Passing the same xTimeIncrement
+ * parameter value will cause the task to execute with a fixed interval.
+ *
+ * @return Number of times xTimeIncrement has been added to pxPreviousWakeTime.
+ * It is 0 on the first call or if not enough ticks have been elapsed since the
+ * last call, 1 in normal circumstances or more than 1 if some period has been
+ * skipped for some reason (e.g. when the caller task is suspended for more than
+ * xTimeIncrement ticks).
+ *
+ * \defgroup xTaskPeriodicDelay xTaskPeriodicDelay
+ * \ingroup TaskCtrl
+ */
+TickType_t xTaskPeriodicDelay( TickType_t * const pxPreviousWakeTime,
+                               const TickType_t xTimeIncrement ) PRIVILEGED_FUNCTION;
+
+/**
+ * task. h
+ * @code{c}
  * BaseType_t xTaskDelayUntil( TickType_t *pxPreviousWakeTime, const TickType_t xTimeIncrement );
  * @endcode
  *
--- a/tasks.c
+++ b/tasks.c
@@ -2362,6 +2362,62 @@ STATIC void prvInitialiseNewTask( TaskFunction_t pxTaskCode,
 
 #if ( INCLUDE_xTaskDelayUntil == 1 )
 
+    TickType_t xTaskPeriodicDelay( TickType_t * const pxPreviousWakeTime,
+                                   const TickType_t xTimeIncrement )
+    {
+        TickType_t xIncrements, xTicksIncrements, xTicksToWait;
+
+        traceENTER_xTaskPeriodicDelay( pxPreviousWakeTime, xTimeIncrement );
+
+        configASSERT( pxPreviousWakeTime );
+        configASSERT( ( xTimeIncrement > 0U ) );
+
+        vTaskSuspendAll();
+        {
+            /* As long as everything is the same type, this plays well with overflows */
+            const TickType_t xTicksElapsed = xTickCount - *pxPreviousWakeTime;
+
+            configASSERT( uxSchedulerSuspended == 1U );
+
+            /* Number of increments to catch up: it could be 0 if
+             * not enough ticks have elapsed, 1 in the common case or
+             * more than 1 if the task has not been resumed in time */
+            xIncrements = xTicksElapsed / xTimeIncrement;
+            xTicksIncrements = xIncrements * xTimeIncrement;
+
+            /* Update to the last wake time */
+            *pxPreviousWakeTime += xTicksIncrements;
+
+            /* Ticks to the next wake time */
+            xTicksToWait = xTimeIncrement - ( xTicksElapsed - xTicksIncrements );
+
+            if( xTicksToWait > 0 )
+            {
+                prvAddCurrentTaskToDelayedList( xTicksToWait, pdFALSE );
+            }
+            else
+            {
+                mtCOVERAGE_TEST_MARKER();
+            }
+        }
+
+        /* Force a reschedule if xTaskResumeAll has not already done so, we may
+         * have put ourselves to sleep. */
+        if( xTaskResumeAll() == pdFALSE )
+        {
+            taskYIELD_WITHIN_API();
+        }
+        else
+        {
+            mtCOVERAGE_TEST_MARKER();
+        }
+
+        traceRETURN_xTaskPeriodicDelay( xIncrements );
+
+        return xIncrements;
+    }
+
+
     BaseType_t xTaskDelayUntil( TickType_t * const pxPreviousWakeTime,
                                 const TickType_t xTimeIncrement )
     {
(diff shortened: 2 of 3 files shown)


archigup:
Is this not solved by the timer task?

ntd:
> Is this not solved by the timer task?

Not sure what to answer here. This PR just implements `xTaskPeriodicDelay` as a "saner" alternative to `xTaskDelayUntil`. You can check #1349 for the whole story.

kstribrnAmzn:
@archigup this could be implemented with a signalling mechanism + a software timer task however that is more setup than simply calling this function. For that reason and because it avoids the runaway problem of `xTaskDelayUntil` I'm actually a fan of this function.

include/task.h
```diff
@@ -862,6 +862,43 @@ void vTaskDelete( TaskHandle_t xTaskToDelete ) PRIVILEGED_FUNCTION;
...
+ * @return Number of times xTimeIncrement has been added to pxPreviousWakeTime.
+ * It is 0 on the first call or if not enough ticks have been elapsed since the
+ * last call, 1 in normal circumstances or more than 1 if some period has been
+ * skipped for some reason (e.g. when the caller task is suspended for more than
+ * xTimeIncrement ticks).
+ *
+ * \defgroup xTaskPeriodicDelay xTaskPeriodicDelay
+ * \ingroup TaskCtrl
+ */
+TickType_t xTaskPeriodicDelay( TickType_t * const pxPreviousWakeTime,
```
kstribrnAmzn, line 930:
```suggestion
UBaseType_t xTaskPeriodicDelay( TickType_t * const pxPreviousWakeTime,
```

Let's use the unsigned based type since this is a simple count

tasks.c
```diff
@@ -2374,6 +2374,64 @@ static void prvInitialiseNewTask( TaskFunction_t pxTaskCode,
 
 #if ( INCLUDE_xTaskDelayUntil == 1 )
 
+    TickType_t xTaskPeriodicDelay( TickType_t * const pxPreviousWakeTime,
+                                   const TickType_t xTimeIncrement )
+    {
+        BaseType_t xAlreadyYielded;
```
kstribrnAmzn:
```suggestion
        BaseType_t xAlreadyYielded = pdFALSE;
```

Assigning a default value here may avoid issues with static analysis checks.

tasks.c
```diff
@@ -2374,6 +2374,64 @@ static void prvInitialiseNewTask( TaskFunction_t pxTaskCode,
...
+        TickType_t xIncrements, xTicksIncrements, xTicksToWait;
+
+        traceENTER_xTaskPeriodicDelay( pxPreviousWakeTime, xTimeIncrement );
+
+        configASSERT( pxPreviousWakeTime );
+        configASSERT( xTimeIncrement > 0 );
```
kstribrnAmzn:
```suggestion
        configASSERT( ( xTimeIncrement > 0U ) );
```

tasks.c
```diff
@@ -2374,6 +2374,64 @@ static void prvInitialiseNewTask( TaskFunction_t pxTaskCode,
...
+            /* This plays well with overflows */
+            const TickType_t xTicksElapsed = xTickCount - *pxPreviousWakeTime;
+
+            configASSERT( uxSchedulerSuspended == 1U );
+
+            /* Number of increments to catch up: it could be 0 if
+             * not enough ticks have elapsed, 1 in the common case or
+             * more than 1 if the task has not been resumed in time */
+            xIncrements = xTicksElapsed / xTimeIncrement;
+            xTicksIncrements = xIncrements * xTimeIncrement;
```
kstribrnAmzn, line 2386: Probably need a check for overflow here. Don't want `xIncrements * xTimeIncrement` to be exactly one more than the `xTicksIncrements` maximum possible value.

include/task.h
```diff
@@ -862,6 +862,43 @@ void vTaskDelete( TaskHandle_t xTaskToDelete ) PRIVILEGED_FUNCTION;
...
+ *   catch up all skipped increments
+ * - it returns the number of increments added to pxPreviosWakeTime
+ *
+ * @param pxPreviousWakeTime Pointer to a variable that holds the time at which the
+ * task was last unblocked.  The variable must be initialised with the current time
+ * prior to its first use.  Following this the variable is automatically updated.
+ *
+ * @param xTimeIncrement The cycle time period.  The task will be unblocked at
+ * time *pxPreviousWakeTime + xTimeIncrement.  Passing the same xTimeIncrement
+ * parameter value will cause the task to execute with a fixed interface period.
```
kstribrnAmzn:
```suggestion
 * parameter value will cause the task to execute with a fixed interval.
```

kstribrnAmzn:
I'm not particularly worried about the failing UTs as those will need to be added.

tasks.c
ntd: `xTicksIncrements` cannot overflow. If you substitute `xIncrements` you get:
```
xTicksIncrements = xTicksElapsed / xTimeIncrement * xTimeIncrement;
```
so `xTicksIncrements` would never exceed `xTicksElapsed`.

include/task.h
ntd: My problem is this code:
```
/* This plays well with overflow */
const TickType_t xTicksElapsed = xTickCount - *pxPreviousWakeTime;
```
needs everything to be of the same type to work properly, so `xTicksElapsed` must be `TickType_t`. Then `xIncrements` (that is, what `xTaskPeriodicDelay` returns) is calculated as:
```
xIncrements = xTicksElapsed / xTimeIncrement;
```
and if I declare it as `UBaseType_t` I get the following error on my host:
```
error: conversion from ‘TickType_t’ {aka ‘long unsigned int’} to ‘UBaseType_t’ {aka ‘unsigned char’} may change value [[-Werror=conversion](https://gcc.gnu.org/onlinedocs/gcc-16.1.0/gcc/Warning-Options.html#index-Wconversion)]
```
I can cast it to remove the error, but in that case I think the result can easily overflow if I suspend a task for too much time.

tasks.c
ntd: As I use it only in one place, I just dropped `xAlreadyYielded` and used the function return value directly.

tasks.c
ntd: :+1: Done.

include/task.h
ntd: :+1: Done.

ntd:
@kstribrnAmzn Many thanks for the review. I force-pushed before realizing you rebased, so I force-pushed again... sorry for the mess.

Some issues you raised are still open though. In the specific, I don't know how to solve the `UBaseType_t` return value in the proper way.

include/task.h
kstribrnAmzn: You're right. I forgot about the mismatch between `UBaseType_t` and `TickType_t` bit width. Overflow is obviously a problem.

tasks.c
kstribrnAmzn: You're right. Closing this out.

tasks.c
```diff
@@ -2362,6 +2362,62 @@ STATIC void prvInitialiseNewTask( TaskFunction_t pxTaskCode,
...
+            xTicksToWait = xTimeIncrement - ( xTicksElapsed - xTicksIncrements );
+
+            if( xTicksToWait > 0 )
+            {
+                prvAddCurrentTaskToDelayedList( xTicksToWait, pdFALSE );
+            }
+            else
+            {
+                mtCOVERAGE_TEST_MARKER();
+            }
```
kstribrnAmzn:
```suggestion
            prvAddCurrentTaskToDelayedList( xTicksToWait, pdFALSE );
```

This one is quite subtle but this conditional will always evaluate to `true` so let's remove the conditional.

An example:
* `xTimeIncrement` == 4
* `xTicksElapsed` == 20
* Therefore `xIncrements` == (20/4) == 5
* Therefore `xTicksIncrements` = 20

Now you get to `xTicksToWait = xTimeIncrement - ( xTicksElapsed - xTicksIncrements )` which works out to be `4 - (20-20)` which evaluates to a non-zero value, entering the if conditional. The same would happen for all other values (19, 18, 17, 16, 15, ...).

kstribrnAmzn:
I've applied the simplification from https://github.com/FreeRTOS/FreeRTOS-Kernel/pull/1368#discussion_r3692093346. Otherwise this PR looks great! I'm going to run the checks now. The only one which should fail is the unit tests since you've introduced a new API. That's not a worry. This is easy enough to fix with a PR to FreeRTOS/FreeRTOS.

I'll have this merge in by the end of the day.

kstribrnAmzn (approved):

AniruddhaKanhere (approved):

PR #1369, opened 2026-02-19
added example project for porting of freertos to stm32u545re mcu

yaswanthksas:
Added STM32U545RE ported version of FreeRTOS bring-up example (Cortex-M33)

Description
------------ 
1. This example demonstrates two tasks are running one after other. 
2. sucessful porting of freeRTOS to stm32u545re mcu. 
3. changed systick handler, pendsv handler, svc handler configurable to cortex m33. 

Test Steps
-----------
1. two tasks are running one after other with round robin fashion and by changing priority of one of the task can make premptive priority fashion. 
Architecture view: 
image


yaswanthksas:
how long it will take to merge in the main branch

cookpate:
This project should be in its own repository. The FreeRTOS org already a reference for another STM32U5 series MCU, also using a Cortex-M33, here: https://github.com/FreeRTOS/iot-reference-stm32u5. 

The reference platform uses Git Submodules to maintain up-to-date dependencies. This PR contains bare copies of the FreeRTOS Kernel and STM drivers and headers, which probably cannot be accepted as-is.

It might be possible to add the stm32u545re to the reference's Projects folder if it can use close to the same setup steps. The main difference should be the choice of project and MCU the user builds and flashes to.

PR #1370, opened 2026-02-19
RISC-V port enhancements

heshamelmatary:
RISC-V port enhancements

Description
-----------
Non-breaking, backward-compatible enhancements to the RISC-V port mostly to do with types in order to make it more portable across architectures (eg: CHERI extensions that enforce memory safety and pointer provenance by separating pointer vs integer types) and environments (eg: POSIX).

Test Steps
-----------
Normal RISC-V demos should work, tested with `main_blinky`. 

Checklist:
----------

Related Issue
-----------

### DIFF
--- a/portable/GCC/RISC-V/port.c
+++ b/portable/GCC/RISC-V/port.c
@@ -93,6 +93,9 @@ const size_t uxTimerIncrementsForOneTick = ( size_t ) ( ( configCPU_CLOCK_HZ ) /
 UBaseType_t const ullMachineTimerCompareRegisterBase = configMTIMECMP_BASE_ADDRESS;
 volatile uint64_t * pullMachineTimerCompareRegister = NULL;
 
+volatile uint32_t * pulTimeHigh = ( volatile uint32_t * const ) ( ( configMTIME_BASE_ADDRESS ) + 4UL ); /* 8-byte type so high 32-bit word is 4 bytes up. */
+volatile uint32_t * pulTimeLow = ( volatile uint32_t * const ) ( configMTIME_BASE_ADDRESS );
+
 /* Holds the critical nesting value - deliberately non-zero at start up to
  * ensure interrupts are not accidentally enabled before the scheduler starts. */
 size_t xCriticalNesting = ( size_t ) 0xaaaaaaaa;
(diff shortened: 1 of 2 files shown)


cookpate:
Seems reasonable enough.

cookpate (approved):

IClementI:
With all respect, @cookpate, this changes do not seems correct .

The variables:

```c
volatile uint32_t * pulTimeHigh = ( volatile uint32_t * const ) ( ( configMTIME_BASE_ADDRESS ) + 4UL );
volatile uint32_t * pulTimeLow  = ( volatile uint32_t * const ) ( configMTIME_BASE_ADDRESS );
````

should only be defined when `configMTIME_BASE_ADDRESS != 0`, as they are not meaningful otherwise. Defining them unconditionally introduces unnecessary and potentially misleading code when no `mtime` is present.

Additionally, any modification made here should also be applied to:

```c
portable/IAR/RISC-V/port.c
```

since it contains nearly identical logic and should remain consistent with the GCC implementation.

Extract from portable/IAR/RISC-V/port.c:
```c
#if ( configMTIME_BASE_ADDRESS != 0 ) && ( configMTIMECMP_BASE_ADDRESS != 0 )

    void vPortSetupTimerInterrupt( void )
    {
        uint32_t ulCurrentTimeHigh, ulCurrentTimeLow;
        volatile uint32_t * const pulTimeHigh = ( uint32_t * ) ( ( configMTIME_BASE_ADDRESS ) + 4UL ); /* 8-byte type so high 32-bit word is 4 bytes up. */
        volatile uint32_t * const pulTimeLow = ( uint32_t * ) ( configMTIME_BASE_ADDRESS );
        volatile uint32_t ulHartId;
```

**I am not a maintainer, so please take my comment with caution. Sorry if I made any mistake.**

```
```

heshamelmatary:
> should only be defined when `configMTIME_BASE_ADDRESS != 0`, as they are not meaningful otherwise. Defining them unconditionally introduces unnecessary and potentially misleading code when no `mtime` is present

`configMTIME_BASE_ADDRESS` must always be defined (to 0 or non-zero value), otherwise FreeRTOS [won't compile](https://github.com/FreeRTOS/FreeRTOS-Kernel/blob/129b09fb1da6fac94c0472e9934299c9d76b0ee8/portable/GCC/RISC-V/portmacro.h#L197). If it's defined to 0, the two variables will be unused, and, depending on the compiler/flags, will likely be optimised away. So this change won't affect any existing assumptions/correctness IMO. Still, guarding them with additional `configMTIME_BASE_ADDRESS != 0` makes sense if that addresses your concern?

IClementI:
Hey, thanks for your quick reply. 

> Still, guarding them with additional configMTIME_BASE_ADDRESS != 0  
> makes sense if that addresses your concern?

Yes, that is exactly my concern, your solution should be good with that.

Would it also make sense to apply the same change in `portable/IAR/RISC-V/port.c`?  
These files are almost copies of each other.

**PS: I am not a maintainer, so please take my comment with caution. Sorry if I made any mistake.**

PR #1396, opened 2026-04-04
stream_buffer: require batching buffers to exceed the trigger level

officialasishkumar:
Description
-----------
Batching stream buffers are documented to unblock receivers only after the buffered byte count exceeds the trigger level. Both `xStreamBufferSend()` and `xStreamBufferSendFromISR()` currently notify as soon as the count reaches that level, which wakes a blocked receiver too early and can make `xStreamBufferReceive()` return `0` bytes while the buffer still holds exactly trigger-level data.

This change routes both send paths through a shared helper so batching buffers require a strict greater-than check, while stream and message buffers keep their existing trigger semantics.

Test Steps
-----------
1. Built the existing `FreeRTOS/Test/CMock/stream_buffer/api` harness against the updated `FreeRTOS-Kernel` working tree.
2. Ran focused local batching-buffer regression cases against that harness for both the task send path and the ISR send path.

Checklist:
----------

Related Issue
-----------
#1375

### DIFF
--- a/stream_buffer.c
+++ b/stream_buffer.c
@@ -257,6 +257,15 @@ typedef struct StreamBufferDef_t
 static size_t prvBytesInBuffer( const StreamBuffer_t * const pxStreamBuffer ) PRIVILEGED_FUNCTION;
 
 /*
+ * Returns pdTRUE when the amount of buffered data should unblock a task that
+ * is waiting to receive data. Stream batching buffers require the buffered
+ * data to exceed the trigger level, whereas stream and message buffers unblock
+ * when the trigger level is reached.
+ */
+static BaseType_t prvBytesInBufferMeetTriggerLevel( const StreamBuffer_t * const pxStreamBuffer,
+                                                    size_t xBytesInBuffer ) PRIVILEGED_FUNCTION;
+
+/*
  * Add xCount bytes from pucData into the pxStreamBuffer's data storage area.
  * This function does not update the buffer's xHead pointer, so multiple writes
  * may be chained together "atomically". This is useful for Message Buffers where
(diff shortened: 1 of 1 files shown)


kstribrnAmzn (approved):

kstribrnAmzn:
Changes look good though a rebase is needed. Please update your PR to contain the latest commits from the main branch.

I did leave a conceptual comment on https://github.com/FreeRTOS/FreeRTOS-Kernel/issues/1375. We can greatly simplify code changes by having the batching buffer trigger when data _equals_ the threshold rather than _exceeding_ the threshold. This is probably a breaking change though so your fix here is best.

kstribrnAmzn:
I've raised https://github.com/FreeRTOS/FreeRTOS-Kernel/pull/1406 which is the counterpart to your PR. Let me know your thoughts. I'm in favor of simplify the number of conditionals required by better aligning existing stream buffer behavior - though your change is the correct one if we want to keep the existing threshold behavior.

officialasishkumar:
Rebased onto current `main` in 4df89dcf5.

kstribrnAmzn:
Accepting your PR since this preserves the behavior that the user would expect per documentation. In the future, as in when the next major Kernel release is done, I'll back out this change and replace it with https://github.com/FreeRTOS/FreeRTOS-Kernel/pull/1406 to reduce code.

kstribrnAmzn:
The only failing check (WIN/MSVC) was failing due to an ARM gitlab repo. This PR does not impact this. Approving.

kstribrnAmzn (approved):

archigup (approved):

PR #1397, opened 2026-04-08
Provide alternative approach for consuming freeRTOS by cmake projects

eMKa94:
Description
-----------
Provide possibility to consume freeRTOS sources as CMake Interface libraries without compromising existing solution.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -16,6 +16,7 @@ cmake_minimum_required(VERSION 3.15)
 # implementation (e.g. when only static allocation is used).
 
 # `freertos_config` target defines the path to FreeRTOSConfig.h and optionally other freertos based config files
+if(NOT CMAKE_INTERFACE_LIBRARY)
 if(NOT TARGET freertos_config )
     if (NOT DEFINED FREERTOS_CONFIG_FILE_DIRECTORY )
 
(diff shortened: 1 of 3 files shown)


archigup:
Hi, could you describe the use case you are looking for? I'm concerned about adding more options to the cmake adding maintenance overhead.

eMKa94:
> Hi, could you describe the use case you are looking for? I'm concerned about adding more options to the cmake adding maintenance overhead.

I've described it a bit [here](https://forums.freertos.org/t/freertos-as-a-cmake-interface-library/23239).

It is not only my personal issue... Current FreeRTOS internal CMake is not quite friendly for projects that define multiple different targets ( Like maintaining an ecosystem with many different devices, which is quite common in many companies... ). Reconfiguring the whole CMake project just to set a different port, heap strategy, or simply different compile options is a huge overkill.

My simple solution provides a way to reuse kernel files and select port/heap files per target as many times as the user wants within a single CMake configure operation....

In other words... 

add_subdirectory(FreeRTOS) does not need to have HEAP / PORT CMake variables.
It just provides interface libraries for kernel, particular ports and heap strategies. 

And any target that need freeRTOS simply link what it needs and have a correct setup with a single CMake Configure operation.

I've decided to share it since a lot of people were saying that it will be cool to have it in the official branch...

PR #1409, opened 2026-04-25
fix: TOCTOU race condition in vTaskListTasks() + add .editorconfig

srpatcha:
### 1. Add .editorconfig for consistent code formatting
Adds a standard `.editorconfig` file to ensure consistent formatting across editors.

### 2. Fix TOCTOU race condition in `vTaskListTasks()` (`tasks.c`)
`vTaskListTasks()` reads the volatile `uxCurrentNumberOfTasks` variable twice — once to set `uxArraySize` and again for the `pvPortMalloc()` call. If a task is created between the two reads, the allocated buffer is too small for the data written by `uxTaskGetSystemState()`, causing a buffer overflow.

**Fix:** Use the local `uxArraySize` variable (already captured) for both the size check and the malloc call.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,21 @@
+root = true
+
+[*]
+end_of_line = lf
+insert_final_newline = true
+trim_trailing_whitespace = true
+charset = utf-8
+
+[*.{c,h,cpp,hpp}]
+indent_style = space
+indent_size = 4
+
+[*.{yml,yaml}]
+indent_style = space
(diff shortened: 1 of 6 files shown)


archigup:
Please separate out the bug fix from the other changes

srpatcha:
Closing per @archigup's request to separate the bug fix from other changes. Reopening as a focused PR with only the TOCTOU fix.

PR #1415, opened 2026-05-12
Add affinity to IDLE tasks in SMP systems

rus084:
Add affinity to IDLE tasks in SMP systems

Description
-----------
currently in FreeRTOS idle threads created without affinity

that means it not possible to get idle percentage of specific core in SMP system via vTaskGetRunTimeStatistics. all IDLE threads will have almost same % time

also vApplicationIdleHook may be called from any cpu core, that may create unexpected conditions.

Test Steps
-----------
set configIDLE_AFFINITY to 1

Checklist:
----------

Related Issue
-----------

### DIFF
--- a/examples/template_configuration/FreeRTOSConfig.h
+++ b/examples/template_configuration/FreeRTOSConfig.h
@@ -558,6 +558,13 @@
  * tskNO_AFFINITY if left undefined. */
 #define configTIMER_SERVICE_TASK_CORE_AFFINITY    tskNO_AFFINITY
 
+/* When using SMP (i.e. configNUMBER_OF_CORES is greater than one), set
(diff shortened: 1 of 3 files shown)


aggarg:
Please add a #define with description here as well - https://github.com/FreeRTOS/FreeRTOS-Kernel/blob/main/examples/template_configuration/FreeRTOSConfig.h#L541.

rus084:
> Please add a #define with description here as well - https://github.com/FreeRTOS/FreeRTOS-Kernel/blob/main/examples/template_configuration/FreeRTOSConfig.h#L541.

done

aggarg (approved):

felicityzhao9 (approved):

PR #1420, opened 2026-05-28
Remove ACL permissions for deleted resource

kstribrnAmzn:
Description
-----------
Setting the `configENABLE_ACL_OBJECT_DELETION_CLEANUP` flag will revoke ACL permissions across
all tasks. Previously the task deleting the object would need to also revoke permissions.

Test Steps
-----------
To be completed

Checklist:
----------

Related Issue
-----------


kstribrnAmzn:
Discarding this until we have further requests for ACL cleanup upon object deletion with MPUv2 wrappers. I don't want to introduce this code as iterating over the task list will slow down the code.

PR #1424, opened 2026-06-08
Add uxTaskCallForEachTask, and refactor uxTaskGetSystemState to use it.

DRNadler:
Get task statistics for all tasks with less memory use, by doing it one task at a time.

Description
-----------
Add uxTaskCallForEachTask, which is the iteration portion of last released uxTaskGetSystemState. uxTaskCallForEachTask calls a user-provided callback once for each task with a filled-in TaskStatus_t.

Refactor uxTaskGetSystemState so that it uses uxTaskCallForEachTask to fill in the pre-allocated array of TaskStatus_t, preserving the existing uxTaskGetSystemState API for current users.

Test Steps
-----------
1. Use uxTaskCallForEachTask to observe task statistics for each task in the system.
2. Compare existing uxTaskGetSystemState output to new version.

Checklist:
----------

Related Issue
-----------
Discussed on forum but no Github Issues.

### DIFF
--- a/.github/SECURITY.md
+++ b/.github/SECURITY.md
@@ -1,3 +1,13 @@
+## Threat model
+
+Before reporting an issue, please review the FreeRTOS kernel threat model. It
+describes the security assumptions the kernel makes, which threats are in scope,
+and the protections the kernel does and does not provide. Understanding these
+boundaries helps determine whether an observed behavior is a security
+vulnerability or expected, documented behavior.
+
+- [FreeRTOS Kernel Threat Model](https://www.freertos.org/Security/02-Kernel-threat-model)
+
 ## Reporting a Vulnerability
 
 If you discover a potential security issue in this project we ask that you notify AWS/Amazon Security 


aggarg:
@DRNadler Thank you for your contribution. Could you please review my suggestion: [aggarg_suggestion.patch](https://github.com/user-attachments/files/29178139/aggarg_suggestion.patch). 

This approach is very similar to yours, with the key difference being that it preserves the existing flow of `uxTaskGetSystemState` and avoids introducing an additional `TaskStatus_t` variable on the stack, ensuring we don't break existing applications.

Please note that I have not yet compiled or tested this patch, as I wanted to get your feedback first.

DRNadler:
@aggarg - I can't seem to apply your patch; could you send me the modified task.c so I can diff and test? Thanks!

aggarg:
@DRNadler Of course! Here it is: [tasks.c](https://github.com/user-attachments/files/29440107/tasks.c). You can diff it with the mainline version of tasks.c.

PR #1431, opened 2026-06-16
fix: TOCTOU race condition in vTaskListTasks()

srpatcha:
## Description

vTaskListTasks() and vTaskList() read the volatile uxCurrentNumberOfTasks twice: once for the zero-check and once for pvPortMalloc(). If a task is created between the two reads, the allocated buffer is undersized, causing a potential buffer overflow in uxTaskGetSystemState().

## Fix

Use the local variable uxArraySize (already assigned from uxCurrentNumberOfTasks) for the pvPortMalloc() call instead of re-reading the volatile.

## Changes

- tasks.c: 2 lines changed (line 7359 and 7528)

Split from #1409 per reviewer request to separate the bug fix from .editorconfig changes.


AniruddhaKanhere (approved):
Thank you for your contribution @srpatcha!
This looks good to me. Approved!

gmtt (approved):

PR #1433, opened 2026-06-19
Optional PRIVILEGED_DATA wrapper + Partner-Supported-Ports submodule update

ianstcdns:
Support moving PRIVILEGED_DATA via an optional MPU wrapper:
- User can define portMOVE_PRIVILEGED_DATA as a section attribute to move the PRIVILEGED_DATA section arbitrarily.
- Allows critical data to be placed in port-specific location for improved performance, notably for cache-coherent SMP.
- Does not require enabling the full MPU wrappers.

Increment Partner-Supported-Ports submodule to the latest commit
- Xtensa Port v3.13 Updates (PR #31)

Passed regression testing with 100+ Xtensa configurations (SMP and single-core) 

Checklist:
----------
- [ x ] I have modified and/or added unit-tests to cover the code changes in this Pull Request.

### DIFF
--- a/include/mpu_wrappers.h
+++ b/include/mpu_wrappers.h
@@ -280,6 +280,12 @@
     #define PRIVILEGED_DATA         __attribute__( ( section( "privileged_data" ) ) )
     #define FREERTOS_SYSTEM_CALL    __attribute__( ( section( "freertos_system_calls" ) ) )
 
+#elif ( defined portMOVE_PRIVILEGED_DATA )
+
+    #define PRIVILEGED_FUNCTION
+    #define PRIVILEGED_DATA         portMOVE_PRIVILEGED_DATA
+    #define FREERTOS_SYSTEM_CALL
+
 #else /* portUSING_MPU_WRAPPERS */
 
     #define PRIVILEGED_FUNCTION
--- a/include/mpu_wrappers.h
+++ b/include/mpu_wrappers.h
@@ -284,6 +284,12 @@
     #define PRIVILEGED_DATA         __attribute__( ( section( "privileged_data" ) ) )
     #define FREERTOS_SYSTEM_CALL    __attribute__( ( section( "freertos_system_calls" ) ) )
 
+#elif ( defined portMOVE_PRIVILEGED_DATA )
+
+    #define PRIVILEGED_FUNCTION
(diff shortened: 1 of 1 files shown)


ianstcdns:
@AniruddhaKanhere any chance you can review, or point this PR to the appropriate reviewer(s)?  Thanks in advance.

aggarg:
Instead of adding a new config, how about changing [this block](https://github.com/FreeRTOS/FreeRTOS-Kernel/blob/main/include/mpu_wrappers.h#L289-L291) to the following:
```c
#ifndef PRIVILEGED_FUNCTION
    #define PRIVILEGED_FUNCTION
#endif

#ifndef PRIVILEGED_DATA
    #define PRIVILEGED_DATA
#endif

#ifndef FREERTOS_SYSTEM_CALL
    #define FREERTOS_SYSTEM_CALL
#endif
```

This makes it more generic as ports can override whichever macro they need in portmacro.h.

ianstcdns:
@aggarg thanks for your suggestion -- it is cleaner and more flexible than my original proposal.  I should be able to make this work, although I will need to rearrange some include file ordering within the Xtensa port.  I will do some testing and submit a PR to the submodule before updating this PR.

Separately, I'm assuming that updating the pointer to the Partner-Supported-Ports submodule through a PR like this is acceptable, but please LMK if there's a preferred procedure.

aggarg:
> Separately, I'm assuming that updating the pointer to the Partner-Supported-Ports submodule through a PR like this is acceptable, but please LMK if there's a preferred procedure.

Yes, this is totally acceptable.

ianstcdns:
@aggarg updated PR with the changes you proposed, and bumped Partner-Supported-Ports reference.  Ready for review again.  Thanks.

aggarg (approved):

aggarg:
@ianstcdns Would you please rebase on top of main?

ianstcdns:
@aggarg PR rebased onto main; no other changes.  Would it also be helpful if I squash the commits, or is that typically done when merging the PR?  Thanks again.

ianstcdns:
@AniruddhaKanhere @patrzhan any chance you could review?  This is a follow-up PR to https://github.com/FreeRTOS/FreeRTOS-Kernel-Partner-Supported-Ports/pull/32 ... thanks in advance

kstribrnAmzn (approved):

PR #1439, opened 2026-06-29
Add uxTaskCallForEachTask, and refactor uxTaskGetSystemState to use it.

DRNadler:
Get task statistics for all tasks with less memory use, by doing it one task at a time.
Provide a public iterator over all tasks for diagnostics.

@aggarg - This is a better refactoring and does not increase stack space use (from #1424):
- The public iterator function is cleaner (passes task handle and state plus context to callback).
- Does not replicate task location code as in your suggested patch.
- Based on the latest development head for easier merge.
- Retested in my application.

Hope you approve!

Description
-----------
Add uxTaskCallForEachTask, which is the iteration portion of last released uxTaskGetSystemState. uxTaskCallForEachTask calls a user-provided callback once for each task (with the task handle and state, plus a context).

Refactor uxTaskGetSystemState so that it uses uxTaskCallForEachTask to fill in the pre-allocated array of TaskStatus_t, preserving the existing uxTaskGetSystemState API for current users.

Test Steps
-----------
1. Use uxTaskCallForEachTask to observe task statistics for each task in the system.
2. Compare existing uxTaskGetSystemState output to new version.

Checklist:
----------

Related Issue
-----------
Discussed on forum but no Github Issues.

### DIFF
--- a/include/task.h
+++ b/include/task.h
@@ -2231,7 +2231,7 @@ char * pcTaskGetName( TaskHandle_t xTaskToQuery ) PRIVILEGED_FUNCTION;
  * boot. pulTotalRunTime can be set to NULL to omit the total run time
  * information.
  *
- * @return The number of TaskStatus_t snapshots provided to the callback.
+ * @return The number tasks provided to the callback.
  */
     UBaseType_t uxTaskCallForEachTask( void (* pxCallbackFunction)( TaskHandle_t xTask,
                                                                     eTaskState eState,
--- a/include/task.h
+++ b/include/task.h
@@ -184,6 +184,13 @@ typedef struct xTASK_STATUS
     #endif
 } TaskStatus_t;
 
+/* Callback type used by uxTaskCallForEachTask(). The callback receives one
+ * task handle and state at a time, plus an opaque caller-supplied context
+ * pointer. The callback may call vTaskGetInfo() if it needs a TaskStatus_t. */
+typedef void (* TaskStatusCallbackFunction_t)( TaskHandle_t xTask,
+                                               eTaskState eState,
+                                               void * pvCallbackContext );
+
 /* Possible return values for eTaskConfirmSleepModeStatus(). */
 typedef enum
 {
@@ -2211,12 +2218,6 @@ char * pcTaskGetName( TaskHandle_t xTaskToQuery ) PRIVILEGED_FUNCTION;
                                       const UBaseType_t uxArraySize,
                                       configRUN_TIME_COUNTER_TYPE * const pulTotalRunTime ) PRIVILEGED_FUNCTION;
 
-/* Callback type used by uxTaskCallForEachTask(). The callback receives one
- * task handle and state at a time, plus an opaque caller-supplied context
- * pointer. The callback may call vTaskGetInfo() if it needs a TaskStatus_t. */
-    typedef void (* TaskStatusCallbackFunction_t)( TaskHandle_t xTask,
-                                                   eTaskState eState,
-                                                   void * pvCallbackContext );
 
 /**
  * For each task, call pxCallbackFunction with the task's handle and state,
--- a/include/task.h
+++ b/include/task.h
@@ -2211,6 +2211,13 @@ char * pcTaskGetName( TaskHandle_t xTaskToQuery ) PRIVILEGED_FUNCTION;
                                       const UBaseType_t uxArraySize,
                                       configRUN_TIME_COUNTER_TYPE * const pulTotalRunTime ) PRIVILEGED_FUNCTION;
 
+/* Callback type used by uxTaskCallForEachTask(). The callback receives one
+ * task handle and state at a time, plus an opaque caller-supplied context
+ * pointer. The callback may call vTaskGetInfo() if it needs a TaskStatus_t. */
+    typedef void (* TaskStatusCallbackFunction_t)( TaskHandle_t xTask,
+                                                   eTaskState eState,
+                                                   void * pvCallbackContext );
+
 /**
  * For each task, call pxCallbackFunction with the task's handle and state,
  * and the provided context.
@@ -2219,6 +2226,9 @@ char * pcTaskGetName( TaskHandle_t xTaskToQuery ) PRIVILEGED_FUNCTION;
  * the scheduler for an extended period. The callback runs while the
  * scheduler is suspended, so it must return quickly and must not perform
  * blocking operations.
+ * 
+ * NOTE: This API is privileged-only (it invokes a user callback from
+ * privileged context).
  *
  * @param pxCallbackFunction Callback to invoke once for each task (passing
  * the task's handle, state, and the pvCallbackContext).
@@ -2233,9 +2243,7 @@ char * pcTaskGetName( TaskHandle_t xTaskToQuery ) PRIVILEGED_FUNCTION;
  *
  * @return The number tasks provided to the callback.
  */
-    UBaseType_t uxTaskCallForEachTask( void ( * pxCallbackFunction )( TaskHandle_t xTask,
-                                                                      eTaskState eState,
-                                                                      void * pvCallbackContext ),
+    UBaseType_t uxTaskCallForEachTask( TaskStatusCallbackFunction_t pxCallbackFunction,
                                        void * pvCallbackContext,
                                        configRUN_TIME_COUNTER_TYPE * const pulTotalRunTime ) PRIVILEGED_FUNCTION;
 
--- a/include/task.h
+++ b/include/task.h
@@ -2233,13 +2233,13 @@ char * pcTaskGetName( TaskHandle_t xTaskToQuery ) PRIVILEGED_FUNCTION;
  *
  * @return The number of TaskStatus_t snapshots provided to the callback.
  */
-    UBaseType_t uxTaskCallForEachTask( void (* pxCallbackFunction)( TaskHandle_t xTask,
-                                                                    eTaskState eState,
-                                                                    void * pvCallbackContext ),
+    UBaseType_t uxTaskCallForEachTask( void ( * pxCallbackFunction )( TaskHandle_t xTask,
+                                                                      eTaskState eState,
+                                                                      void * pvCallbackContext ),
                                        void * pvCallbackContext,
                                        configRUN_TIME_COUNTER_TYPE * const pulTotalRunTime ) PRIVILEGED_FUNCTION;
 
-#endif
+#endif /* if ( configUSE_TRACE_FACILITY == 1 ) */
 
 /**
  * task. h
--- a/include/task.h
+++ b/include/task.h
@@ -185,9 +185,10 @@ typedef struct xTASK_STATUS
 } TaskStatus_t;
 
 #if ( configUSE_TRACE_FACILITY == 1 )
-    /* Callback type used by uxTaskCallForEachTask(). The callback receives one
-     * task handle and state at a time, plus an opaque caller-supplied context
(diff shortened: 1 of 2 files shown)


rawalexe:
Hello @DRNadler,
Thank you for your contribution can you also please add some unit tests

DRNadler:
@rawalexe - Can you give me a hint about how I would add a unit test for this, as it must run in a FreeRTOS application? Also, why am I receiving messages like:
`[FreeRTOS/FreeRTOS-Kernel] PR run failed: CMock Unit Tests, Attempt #2 - Add `
`[FreeRTOS/FreeRTOS-Kernel] PR run failed: CI Checks, Attempt #2 - Add uxTaskCallForEachTask, and refactor uxTaskGetSystemState to use it. (4df5099) `
Thanks!

DRNadler:
@rawalexe @aggarg - Let's try again:

1. I've applied the formatting patch for the ancient uncrustify that failed format check.
2. Your mock-generation tool **lost a typedef** introduced in task.h, which subsequently caused a compile failure. I changed the header to avoid using a typedef, making it much less readable, aaarrrgggg....

Regarding a unit test: This code is a refactoring. The added public interface to iterate over tasks is used in `uxTaskGetSystemState`, hence all new/refactored code will be covered by the existing unit test. Is there actually a unit test for `uxTaskGetSystemState`? The configuration file used for unit tests does not enable this code...

Hopefully this is finally OK,
Thanks!
Best Regards, Dave

DRNadler (commented):
fix formating 'correction' -  missing `void` in `tasks.h`

rawalexe:
Hello @DRNadler,
Do not worry about the formatting ones, i can fix those for you, I had  a lot of PRs to go though last week and do not remember the context around this one request, but your explanation makes sense. Give me some time and will review the PR. Thank you for your contribution

DRNadler:
Thanks @rawalexe - I just fixed an initializer that wasn't C89 (good lord, still?) compatible.
Hope it is finally OK!

DRNadler:
@rawalexe - cmock test failed due to an unrelated change. 
Appears to be a unit-test bug, nothing I can do anything about...
ChatGPT wrote this summary:

# FreeRTOS-Kernel CI Failure Report

## Summary

GitHub Actions run **29527926691** failed during the **“Run Unit Tests for coverage”** step.

The sanitizer test run passed. The coverage run failed because one mutex unit test triggered a `configASSERT()` in `queue.c`.

## Root Cause

The failing test was:

`test_macro_xSemaphoreTake_blocking_mutex_inherit_timeout_high_prio_waiting`

The test deliberately inserted a fake task into the mutex queue’s `xTasksWaitingToReceive` list, then called `vSemaphoreDelete()` without removing that task.

During deletion, `vQueueDelete()` asserted that the receive-waiting list must be empty:

```c
configASSERT( pxQueue );
configASSERT( listLIST_IS_EMPTY( &( pxQueue->xTasksWaitingToSend ) ) );
configASSERT( listLIST_IS_EMPTY( &( pxQueue->xTasksWaitingToReceive ) ) );
```

The assertion occurred at `queue.c:2265`.

## Secondary Failure

Because the mutex test failed, its coverage file was not generated. This prevented creation of `queue_semaphore.info`.

The final `lcov` merge then failed because that file did not exist, causing the workflow to exit with code 2.

Therefore, the `lcov` error was a consequence of the unit-test failure, not the original cause.

## Relation to PR #1439

PR #1439 changes only:

- `include/task.h`
- `tasks.c`

It does not modify `queue.c` or the failing mutex test.

The failure therefore appears to be an existing unit-test cleanup defect exposed by the current queue-deletion assertion, rather than a defect in the task-iterator changes.

## Recommended Fix

Update the failing unit test so that the fake task is removed from `xTasksWaitingToReceive` before calling:

```c
vSemaphoreDelete( xSemaphore );
```

Using `lcov --ignore-errors empty` would only mask the secondary coverage error and would not fix the failed test.

rawalexe:
I can take over the rest, Don't worry about it. I apologize for the testing infra sometimes it can be painful. I will update thisPR as I find time. :) Thank you for the contribution really appreciate it

include/task.h
```diff
@@ -2210,7 +2210,36 @@ char * pcTaskGetName( TaskHandle_t xTaskToQuery ) PRIVILEGED_FUNCTION;
...
+ * each callback invocation.
+ *
+ * @param pulTotalRunTime If configGENERATE_RUN_TIME_STATS is set to 1 in
+ * FreeRTOSConfig.h then *pulTotalRunTime is set to the total run time since
+ * boot. pulTotalRunTime can be set to NULL to omit the total run time
+ * information.
+ *
+ * @return The number of TaskStatus_t snapshots provided to the callback.
+ */
+    UBaseType_t uxTaskCallForEachTask( void ( * pxCallbackFunction )( TaskHandle_t xTask,
```
rawalexe: **[Blocking]** The callback type is declared here as a spelled-out function pointer, but the definition in `tasks.c` uses the file-local typedef `TaskStatusCallbackFunction_t`. Since that typedef is not exposed in any public header, users cannot declare conforming callbacks without re-typing the signature by hand, and the header/definition declarations differ textually.

Suggest moving the typedef into `task.h` (above this declaration) and using it in both the declaration and the `tasks.c` definition:

```c
typedef void (* TaskStatusCallbackFunction_t)( TaskHandle_t xTask,
                                               eTaskState eState,
                                               void * pvCallbackContext );

UBaseType_t uxTaskCallForEachTask( TaskStatusCallbackFunction_t pxCallbackFunction,
                                   void * pvCallbackContext,
                                   configRUN_TIME_COUNTER_TYPE * const pulTotalRunTime ) PRIVILEGED_FUNCTION;
```

tasks.c
```diff
@@ -4457,11 +4441,108 @@ char * pcTaskGetName( TaskHandle_t xTaskToQuery )
...
+        {
+            /* Visit each task in the Suspended state. */
+            uxTask = ( UBaseType_t ) ( uxTask + prvForEachTaskInList( &xSuspendedTaskList, eSuspended, pxCallbackFunction, pvCallbackContext ) );
+        }
+        #endif
+
+        return uxTask;
+    }
+
+    UBaseType_t uxTaskCallForEachTask( TaskStatusCallbackFunction_t pxCallbackFunction,
```
rawalexe, line 4517: **[Blocking]** This new public API has no MPU wrapper plumbing. `uxTaskGetSystemState` has full support (`MPU_uxTaskGetSystemState` in `mpu_wrappers.h`, `mpu_prototypes.h`, syscall number, and `mpu_wrappers_v2.c` implementation), but `uxTaskCallForEachTask` has none — unprivileged callers on MPU-enabled ports will fault or fail to link.

Either add the corresponding MPU wrapper set, or explicitly document this API as privileged-only. (Worth noting a raw function-pointer callback invoked from kernel context is difficult to wrap safely for unprivileged callers, so documenting it as privileged-only may be the more defensible choice.)

include/task.h
```diff
@@ -2210,7 +2210,36 @@ char * pcTaskGetName( TaskHandle_t xTaskToQuery ) PRIVILEGED_FUNCTION;
...
+ *
+ * @param pvCallbackContext Opaque caller-provided context passed through to
+ * each callback invocation.
+ *
+ * @param pulTotalRunTime If configGENERATE_RUN_TIME_STATS is set to 1 in
+ * FreeRTOSConfig.h then *pulTotalRunTime is set to the total run time since
+ * boot. pulTotalRunTime can be set to NULL to omit the total run time
+ * information.
+ *
+ * @return The number of TaskStatus_t snapshots provided to the callback.
```
rawalexe: The callback receives `TaskHandle_t` + `eTaskState`, not `TaskStatus_t` snapshots — this reads like a copy-paste from `uxTaskGetSystemState`. Suggest: "The number of tasks for which the callback was invoked."

tasks.c
```diff
@@ -4457,11 +4441,108 @@ char * pcTaskGetName( TaskHandle_t xTaskToQuery )
...
+                    *pulTotalRunTime = ( configRUN_TIME_COUNTER_TYPE ) portGET_RUN_TIME_COUNTER_VALUE();
+                #endif
+            #else
+                *pulTotalRunTime = 0;
+            #endif /* if ( configGENERATE_RUN_TIME_STATS == 1 ) */
+        }
+    }
+
+/* For each task, call the provided callback function (passing the provided context). */
+    STATIC UBaseType_t prvCallForEachTask( TaskStatusCallbackFunction_t pxCallbackFunction,
```
rawalexe, line 4484: `prvCallForEachTask` is called both from `uxTaskCallForEachTask` (which wraps it in `vTaskSuspendAll()`) and from `uxTaskGetSystemState` (which already holds the scheduler suspended). Safe because suspension nests, but suggest a brief comment here stating the caller must hold the scheduler suspended, so future callers don't assume this function provides its own atomicity.

tasks.c
```diff
@@ -4457,11 +4441,108 @@ char * pcTaskGetName( TaskHandle_t xTaskToQuery )
...
+        return uxTask;
+    }
+
+    UBaseType_t uxTaskCallForEachTask( TaskStatusCallbackFunction_t pxCallbackFunction,
+                                       void * pvCallbackContext,
+                                       configRUN_TIME_COUNTER_TYPE * const pulTotalRunTime )
+    {
+        UBaseType_t uxTask;
+
+        configASSERT( pxCallbackFunction != NULL );
```
rawalexe, line 4523: `configASSERT` compiles out in release builds, so a NULL `pxCallbackFunction` would be dereferenced inside the loop with the scheduler suspended. Suggest an early `return 0;` guard after the assert, consistent with defensive checks in other public APIs.

rawalexe (changes requested):
Thanks for the contribution — the refactor of `uxTaskGetSystemState()` on top of the new iterator looks behaviorally equivalent (list order, `eRunning` detection via `vTaskGetInfo`, run-time counter handling all preserved). Two blocking items around the new public API surface, plus a few smaller suggestions inline.

rawalexe:
Please look into the comments and let me know your thought. We can discuss on how to work these out

DRNadler:
Appreciate the help @rawalexe ! To your points:

1. typedef for callback: This is how I originally coded it, but unfortunately your cmock generator created code that used the typedef, but did not include task.h - so compilation failed. If you can fix the test case, I will move the typedef back to where I originally had it. Agreed, much more readable!
2. MPU wrapper plumbing: Per your suggestion, I will document this as privileged only.
3. `uxTaskCallForEachTask` old comment corrected (already pushed).
4. I will add comment about keeping scheduler suspended.
5. I forgot configASSERT compiles out - in my current project I force it in release build and write failure to a log but that's not standard! Will add guard.

Thanks again!

rawalexe:
Thank you for the quick turn around, I will fix other CI failures as I find time

rawalexe:
we might need to update the test so that your path of ifdef is covered for this, possibly. Let me do a bit digging and get back

tasks.c
kstribrnAmzn: This looks to be resolved

tasks.c
kstribrnAmzn: I don't think we want a wrapper here. Adding a wrapper would allow unprivileged tasks the ability to execute a callback for each task - potentially giving system-wide visibility.

kstribrnAmzn:
Added a couple of commits to take care of formatting and unit test issues. You change should be good to go now. When the checks pass I'll approve.

kstribrnAmzn (approved):

rawalexe (approved):

PR #1441, opened 2026-06-30
Fix vPortFreeSecureContext to read xSecureContext at correct offset

aggarg:
Description
-----------
When MPU is enabled, the first item in the TCB is not the top of the stack but the stored context location. As a result, `xSecureContext` is located at a negative offset from that position rather than at offset 0. The current implementation unconditionally reads `xSecureContext` at offset 0, which returns an incorrect value when MPU is enabled.

This commit updates `vPortFreeSecureContext` to read `xSecureContext` at the correct offset based on the port configuration:
- CM33/CM35P/CM52/CM55/CM85/STAR_MC3: -20 (or -36 with PAC enabled)
- CM23: -20 (no PAC support)
- Without MPU: 0 (`xSecureContext` remains at the top of stack)

Test Steps
-----------
Tested using [CORTEX_MPU_M33F_Nordic_NRF9160_SES](https://github.com/FreeRTOS/FreeRTOS/tree/main/FreeRTOS/Demo/CORTEX_MPU_M33F_Nordic_NRF9160_SES) demo.

Checklist:
----------

Related Issue
-----------
NA

### DIFF
--- a/portable/ARMv8M/non_secure/portable/GCC/ARM_CM23/portasm.c
+++ b/portable/ARMv8M/non_secure/portable/GCC/ARM_CM23/portasm.c
@@ -48,6 +48,12 @@
 #endif
 
 #if ( configENABLE_MPU == 1 )
+    #define SECURE_CONTEXT_OFFSET   -20
+#else
+    #define SECURE_CONTEXT_OFFSET   0
+#endif
+
+#if ( configENABLE_MPU == 1 )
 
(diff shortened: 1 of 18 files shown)


kstribrnAmzn (approved):

kstribrnAmzn:
Worked with you to look at the compiled assembly and verify the offset change on both IAR and Keil. Approving.

cookpate (approved):

PR #1445, opened 2026-07-04
docs: clarify MPU region parameter macros

### LINKED ISSUE #1384: [DOC] API deviation for CM33 port is not documented
**Describe the issue**
[The documentation for xTaskCreateRestrictedStatic](https://www.freertos.org/Documentation/02-Kernel/04-API-references/13-FreeRTOS-MPU-specific/02-xTaskCreateRestrictedStatic) states that the `portMPU_REGION_*` macros are to be used. Most ARM ports follow this API. However, the CM33 port (and maybe other ARMv8 as well) do not follow this convention, but instead use the `tskMPU_REGION_*` macros.

Using the `portMPU_REGION_*_SHAREABLE` attributes clashes with the access attributes. Shareability cannot be configured (see also https://github.com/FreeRTOS/FreeRTOS-Kernel/issues/1383), and using the corresponding macros in the call to `xTaskCreateRestrictedStatic` can override the access attributes.

This deviation could be classified as a "bug", but the different attributes are encoded in different registers, and therefore cannot - as in other ARM ports, be expressed as a bitwise OR of the single flags in a single register. So I think the documented API cannot be implemented for this port.
The documentation should
* state that different ports use different APIs
* list both API variants
* either list which ports use which API, or explain how to find out which API a specific port uses

**Reference**
https://www.freertos.org/Documentation/02-Kernel/04-API-references/13-FreeRTOS-MPU-specific/02-xTaskCreateRestrictedStatic

**Screenshot**

Image

**Browser**
- Browser: Firefox
- Version: any

See also: https://forums.freertos.org/t/shareable-mpu-regions-for-dma-on-cortex-m33/24918

Old-Ding:
## Summary

- Clarify that `MemoryRegion_t.ulParameters` uses port-specific MPU region parameter macros.
- Keep the existing `portMPU_REGION_*` examples for Cortex-M3/4 MPU ports.
- Document that ARMv8-M MPU ports such as CM23, CM33, CM52, CM55, CM85, and STAR_MC3 use the `tskMPU_REGION_*` values from `task.h`, which the port translates into MPU register settings.

Fixes FreeRTOS/FreeRTOS-Kernel#1384

## Impact

This is a documentation-only change. It does not change any API, ABI, macro value, structure layout, configuration option, or port behavior.

## Validation

- `git diff --check -- include/task.h`
- `git diff --stat` and `git diff --word-diff -- include/task.h`
- `rg -n "Some ports, including the Cortex-M3/4 MPU ports|ARMv8-M MPU ports|MemoryRegion_t\.ulParameters|tskMPU_REGION|portMPU_REGION|xTaskCreateRestrictedStatic|vTaskAllocateMPURegions" include/task.h`
- Static source check: ARMv8-M ports consume `tskMPU_REGION_*` from `MemoryRegion_t.ulParameters`.
- Static source check: Cortex-M3/4 MPU ports still consume `portMPU_REGION_*` from `MemoryRegion_t.ulParameters`.

Doxygen output was not generated locally because `doxygen` is not installed in this environment.

### DIFF
--- a/include/task.h
+++ b/include/task.h
@@ -548,6 +548,14 @@ typedef enum
  * The function parameters define the memory regions and associated access
  * permissions allocated to the task.
  *
+ * The parameter macros used in MemoryRegion_t.ulParameters are port specific.
+ * Some ports, including the Cortex-M3/4 MPU ports, use the portMPU_REGION_*
+ * values shown below.
+ * ARMv8-M MPU ports, such as CM23, CM33, CM52, CM55, CM85 and STAR_MC3, use
+ * the tskMPU_REGION_* values defined in this header; the port translates them
+ * into MPU register settings. Check the selected port's headers before selecting
+ * the region parameter macros.
+ *
  * See xTaskCreateRestrictedStatic() for a version that does not use any
  * dynamic memory allocation.
  *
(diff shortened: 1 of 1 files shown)


rawalexe (approved):

rawalexe (approved):
Good clarification for a common source of confusion between `portMPU_REGION_*` and `tskMPU_REGION_*` macros across MPU port variants. No logic changes, comment-only. LGTM.

cookpate (approved):

PR #1473, opened 2026-08-20
Mask MPU region base addresses when writing to MPU_RBAR to fix potential kernel region setting override

CocoDico78:
Description
-----------
While the docs are extensively mentioning that all regions *should* be aligned to power of two for correct MPU configuration, it is not explicit enough that *not aligning* (or not sanitizing) a user-defined region or a stack region may override MPU settings defined for higher priority kernel regions.

For example, for the ARM_CM3_MPU port, in port.c at lines 1318 (`https://github.com/FreeRTOS/FreeRTOS-Kernel/blob/ce221a8bb468e462ca6b435cef66a9636e00baf4/portable/GCC/ARM_CM3_MPU/port.c#L1318`):

```c
xMPUSettings->xRegion[ 0 ].ulRegionBaseAddress =
    ( ( uint32_t ) pxBottomOfStack ) |
    ( portMPU_REGION_VALID ) |
    ( portSTACK_REGION ); /* Region number. */
```

will escalate the actual region priority from 3 (value of `portSTACK_REGION`) to 7 (value of `portPRIVILEGED_RAM_REGION`) if `pxBottomOfStack ` is maliciously or inadvertently shifted by 4 bytes (for example passing `0x20000004` instead of `0x20000000`).

Same issue for user-defined regions which has the same bitwise calculation without masking.

This is in particular dangerous for implementers still using MPU wrappers v1 (which allows restricted tasks to create other restricted tasks, allowing to pass arbitrary misaligned stack pointers), and dangerous for implementers using MPU wrappers v2 who are not aware and might blindly trust user defined pointers thinking that the higher priority MPU regions will guard them.
Base addresses should be aligned to a power of two for proper MPU configuration. Masking ensures that a misaligned base address cannot modify less significant bits in the attribute reserved for other use, such as the region bits. Failing to mask the address may allow a malicious user to pass in misaligned addresses in a user-defined region or as stack buffer which could in turn override the settings for higher-priority kernel-defined regions.

In this pull-request, we enforce such masking on the ARM_CM3/4/4F_MPU ports. Port ARM_CM0 is untouched.

Test Steps
-----------
- Start a restricted task and pass a stack buffer starting at `0x20000004` in a user-defined region
- Observe that the previously set `portPRIVILEGED_RAM_REGION` settings are no longer overriden by the user-defined region

Checklist:
----------

Related Issue
-----------
Follow up of a [topic on the forum](https://forums.freertos.org/t/user-defined-misaligned-mpu-regions-can-override-kernel-defined-regions-on-arm-cm3-mpu/25130).

### DIFF
--- a/portable/GCC/ARM_CM3_MPU/port.c
+++ b/portable/GCC/ARM_CM3_MPU/port.c
@@ -95,6 +95,7 @@ typedef void ( * portISR_t )( void );
 #define portMPU_REGION_ENABLE                     ( 0x01UL )
 #define portPERIPHERALS_START_ADDRESS             0x40000000UL
 #define portPERIPHERALS_END_ADDRESS               0x5FFFFFFFUL
+#define portMPU_RBAR_ADDRESS_MASK                 0xFFFFFFE0
 
 /* Constants required to access and manipulate the SysTick and other FreeRTOS
  * interrupts. */
@@ -1137,7 +1138,7 @@ static void prvSetupMPU( void )
     if( portMPU_TYPE_REG == portEXPECTED_MPU_TYPE_VALUE )
     {
         /* First setup the unprivileged flash for unprivileged read only access. */
-        portMPU_REGION_BASE_ADDRESS_REG = ( ( uint32_t ) __FLASH_segment_start__ ) | /* Base address. */
+        portMPU_REGION_BASE_ADDRESS_REG = ( ( ( uint32_t ) __FLASH_segment_start__ ) & portMPU_RBAR_ADDRESS_MASK ) | /* Base address. */
                                           ( portMPU_REGION_VALID ) |
                                           ( portUNPRIVILEGED_FLASH_REGION );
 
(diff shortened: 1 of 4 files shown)


aggarg:
@CocoDico78 I would have expected the change to clear the VALID and REGION bits in the RBAR register. Would something like the following work instead?
```c
#define RBAR_ADDRESS_MASK 0xFFFFFFE0
xMPUSettings->xRegion[ 0 ].ulRegionBaseAddress =
            ( ( ( uint32_t ) __SRAM_segment_start__ ) & RBAR_ADDRESS_MASK ) | /* Base address. */
            ( portMPU_REGION_VALID ) |
            ( portSTACK_REGION );  
```

CocoDico78:
@aggarg 
> Would something like the following work instead?
> 
> ```c
> #define RBAR_ADDRESS_MASK 0xFFFFFFE0
> ```

Yes, this static mask would absolutely protect the bits [0-4] and thus solve the main issue here. This static mask would however still not prevent users from passing improperly aligned addresses: say the the user is configuring a region of size 64; bit 5 could still be incorrectly set if the user passes a base address not aligned with 64 but aligned with 32.

You are suggesting a simpler, static mask, while I'm presenting a solution with a dynamic mask that guarantees that the written ADDR is always correctly aligned. What are your thoughts? I'm fine either way.

EDIT: the simplicity argument compels me. First, let's solve the VALID and REGION masking. The discussion on ensuring alignment in the ADDR field is independent.

aggarg (dismissed):

aggarg:
@CocoDico78 Similar change is needed in CM4_MPU ports as well, right? Can you do that too?

CocoDico78:
@aggarg i've implemented similar change on ports ARM_CM3/4/4F_MPU ports.
I left port ARM_CM0 untouched because it already defines `#define portMPU_RBAR_ADDRESS_BITMASK ( 0xFFFFFF00UL )`, which troubles me because the bit mask [should also be](https://support.arm.com/documentation/dui0662/b/Cortex-M0--Peripherals/Memory-Protection-Unit/MPU-Region-Base-Address-Register) `0xFFFFFFE0UL` so I'd rather discuss before changing.

PR #1475, opened 2026-08-20
Allow start scheduler and free context SVC from privileged code only

kstribrnAmzn:
Description
-----------
Restrict further functions to privileged code.

Thanks @aggarg for the help developing this!

Test Steps
-----------
Tested on ARMv8m port. Demo worked.

Checklist:
----------

Related Issue
-----------


rawalexe (dismissed):

portable/ARMv8M/non_secure/port.c
```diff
@@ -1170,39 +1182,54 @@ void vPortSVCHandler_C( uint32_t * pulCallerStackAddress ) /* PRIVILEGED_FUNCTIO
...
-
-            /* R0 contains TCB being freed and R1 contains the secure
-             * context handle to be freed. */
-            ulR0 = pulCallerStackAddress[ 0 ];
-            ulR1 = pulCallerStackAddress[ 1 ];
-
-            /* Free the secure context. */
-            SecureContext_FreeContext( ( SecureContextHandle_t ) ulR1, ( void * ) ulR0 );
+            #if( configENABLE_MPU == 1 )
+            if( ( ulPC >= ( uint32_t ) __privileged_functions_start__ ) &&
```
aadityasg, line 1186: The `__privileged_functions_start__`/`__end__` declarations added to vPortSVCHandler_C are guarded by:

`#if ( ( configENABLE_MPU == 1 ) && ( configENABLE_TRUSTZONE == 1 ) )`

  But the `portSVC_START_SCHEDULER` case uses those symbols under a weaker guard:

```
  #if( configENABLE_MPU == 1 )
  if( ( ulPC >= ( uint32_t ) __privileged_functions_start__ ) && ... )
```

  The `FREE_SECURE_CONTEXT` usage is safe (it lives inside `#if configENABLE_TRUSTZONE == 1)`, but `START_SCHEDULER` runs in *all* configs. So for a build with `configENABLE_MPU == 1` and `configENABLE_TRUSTZONE == 0` — i.e. every *_NTZ port (CM23_NTZ, CM33_NTZ,
  CM35P_NTZ, CM52_NTZ, CM55_NTZ, CM85_NTZ, STAR_MC3_NTZ) — the symbols are used but never declared in that function scope → **compile error**.

  I (Claude) verified against main that these symbols have no file-scope declaration; the only other declarations are function-local externs inside prvSetupMPU() and vSystemCallExit(), which do not leak into vPortSVCHandler_C. So this is a genuine break, replicated across 6–7 NTZ ports.

kstribrnAmzn: This is a great catch! I've fixed this with the latest commit.

aadityasg (approved):

PR #1485, opened 2026-08-28
portable: Fix malformed comment in IAR ARM_CM33 port

berringtoni36-commits:
## Description

The IAR ARM_CM33 non-secure port contains a URL outside the assembly comment delimiters in `portasm.s`. IAR reports an "Invalid syntax" error because it parses the bare URL as assembly code.

Move the opening `/*` before the URL so the whole explanatory text is a valid comment.

Fixes #1480

## Testing

- `git diff --check`
- Verified the changed line is now enclosed by `/* ... */`; target IAR toolchain is not available on macOS.

Issue #1480, opened 2026-08-25
[BUG] IAR ARMCM33 Non-Secure portasm.s stray URL outside comment delimiters causes "Invalid syntax" assembler error in portasm.s


murali78260:
**Describe the bug**
Building the FreeRTOS-Kernel with the IAR toolchain targeting the ARM CM33 TrustZone Non-Secure port fails immediately with an assembler __"Invalid syntax"__ error at `**portable/IAR/ARM_CM33/non_secure/portasm.s**`.

The root cause is a stray URL **(`http://files.iar.com/ftp/pub/box/bxarm-9.60.3.deb`) that was accidentally placed outside the `/* */` comment delimiters on that line. The IAR assembler sees the bare URL text as an invalid assembly token and aborts.**

Corrupted line as it exists in the file:
    ldmia r2!, {r0, r3, lr}             _http://files.iar.com/ftp/pub/box/bxarm-9.60.3.deb/* Read from stack - r0 = xSecureContext, r3 = PSPLIM and LR restored. */_

The `/*` needs to be moved before the URL so the entire text is inside the comment:
    ldmia r2!, {r0, r3, lr}             _**/* http://files.iar.com/ftp/pub/box/bxarm-9.60.3.deb Read from stack - r0 = xSecureContext, r3 = PSPLIM and LR restored. */**_

**Target**
- Development board: PNEV7642FAMA(NXP)
- Instruction Set Architecture: ARM CM33 TZ
- IDE and version:  IAR ARM 9.70.2
- Toolchain and version: ewarm-9.70.2

**Host**
- Host OS: Windows
- Version:  Windows 11

**To Reproduce**
- Clone FreeRTOS-Kernel `main` branch.
- Open or create an IAR EWARM project targeting an ARM CM33 TrustZone device (e.g. NXP PNEV7642FAMA).
- Add `portable/IAR/ARM_CM33/non_secure/portasm.s` to the project.
- Build the project.
- The assembler fails on line 466 with: `Error[0]: Invalid syntax portasm.s 466`

**Expected behavior**
The file should assemble without errors. The URL on line 466 must be enclosed inside the `/* */` comment delimiters so the assembler does not encounter bare non-instruction text.

**Screenshots**

<img width="1874" height="122" alt="Image" src="https://github.com/user-attachments/assets/fea029f1-f823-4da6-9d25-b853de124374" />

[...]

PR #1363, opened 2026-02-03
Add Arm China STAR-MC3 port support

cpussw01:
Add Arm China STAR-MC3 port support 

Description
-----------
Changes:

Kernel part no changes, copied for ARM_CM52 , only chang portARCH_NAME and update License Info:

portable/ARMv8M/non_secure/portable/GCC/STAR_MC3/
portable/ARMv8M/non_secure/portable/IAR/STAR_MC3/
portable/GCC/STAR_MC3/
portable/GCC/STAR_MC3_NTZ/
portable/IAR/STAR_MC3/
portable/IAR/STAR_MC3_NTZ/

Common part changes, include CmakeList/copy_files.py/Docs:

CMakeLists.txt
examples/template_configuration/FreeRTOSConfig.h
portable/ARMv8M/ReadMe.txt
portable/ARMv8M/copy_files.py
portable/ARMv8M/non_secure/ReadMe.txt
portable/ARMv8M/secure/ReadMe.txt
portable/CMakeLists.txt
portable/ThirdParty/GCC/ARM_TFM/README.md

Test Steps
-----------
Using STAR_MC3 port, build a FreeRTOS_Blinky demo, running passed on in-house STAR-MC3 MPS3 board

Checklist:
----------

- [x] I have tested my changes. No regression in existing tests.

Related Issue
-----------
 #1359

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #1334, opened 2025-11-13
Add ARM Cortex-M52 port support FreeRTOS#1305

cpussw01:
Fixes #1305
- Add ARM_CM52 and ARM_CM52_NTZ port implementations for GCC and IAR compilers

Description
-----------
Changes:
CMakeLists.txt
portable/ARMv8M/ReadMe.txt
portable/ARMv8M/copy_files.py
portable/ARMv8M/non_secure/ReadMe.txt
portable/ARMv8M/secure/ReadMe.txt
portable/CMakeLists.txt

Add ports for Cortex-M52 ,there are copied form correspond CM55 dir, only portmacro.h CPU_NAME changed 
 portable/ARMv8M/non_secure/portable/GCC/ARM_CM52/
 portable/ARMv8M/non_secure/portable/IAR/ARM_CM52/
 portable/GCC/ARM_CM52/
 portable/GCC/ARM_CM52_NTZ/
 portable/IAR/ARM_CM52/
 portable/IAR/ARM_CM52_NTZ/

Test Steps
-----------
Reference FreeRTOS/Demo/ThirdParty/Partner-Supported-Demos/CORTEX_M85_MPU_PXN_FVP_GCC/  Cortex-M85 demo 
Create an Cortex-M52 demo , compile + build and test on Cortex-M52 device 

Checklist:
----------

- [x] I have tested my changes. No regression in existing tests.
- [ ] I have modified and/or added unit-tests to cover the code changes in this Pull Request.

Related Issue
#1305 
-----------

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #1492, opened 2026-09-14
portable: remove stray installer URL from IAR TrustZone non-secure ports

sekior11:
## Problem

A stray IAR installer download URL sits **outside** the comment delimiters in the
`restore_special_regs` section of the IAR TrustZone non-secure ports:

```asm
        ldmia r2!, {r0, r3, lr}             http://files.iar.com/ftp/pub/box/bxarm-9.60.3.deb/* Read from stack - r0 = xSecureContext, r3 = PSPLIM and LR restored. */
```

Because `/*` appears *after* the URL, the assembler reads the bare URL as an operand
of the preceding `ldmia` and aborts:

```
Error: garbage following instruction -- `ldmia r2!,{r0,r3,lr} http:'
```

The line sits **outside** the `#if ( configENABLE_PAC == 1 )` block, which ends on the
preceding line, so it is not limited to PAC-enabled builds. It is, however, inside the
`#else` branch of `#if ( configENABLE_MPU == 1 )`: the breakage is on **every non-MPU
build**, independently of `configENABLE_PAC`, while MPU builds never compile the line.

*(Corrected — an earlier revision of this description said "breaks every configuration
of these ports". That was too strong; the guard stack for line 476 is
`#else of #if ( configENABLE_MPU == 1 )` and nothing else.)*

## Scope

Issue #1480 reports this for `portable/IAR/ARM_CM33/non_secure/portasm.s`.

A tree-wide scan of `main` (`8be86d4`) finds the stray URL in **7** files, all at
line 476. This PR fixes all seven:

- `portable/ARMv8M/non_secure/portable/IAR/ARM_CM33/portasm.s`
- `portable/IAR/ARM_CM33/non_secure/portasm.s`
- `portable/IAR/ARM_CM35P/non_secure/portasm.s`
- `portable/IAR/ARM_CM52/non_secure/portasm.s`
- `portable/IAR/ARM_CM55/non_secure/portasm.s`
- `portable/IAR/ARM_CM85/non_secure/portasm.s`
- `portable/IAR/ARM_STAR_MC3/non_secure/portasm.s`

The first revision covered only the last six, leaving
`portable/IAR/ARM_CM33/non_secure/portasm.s` to #1485. Per @jefftenney's review
comment that split no longer makes sense: `portable/ARMv8M/copy_files.py` derives the
non-secure IAR ports from `portable/ARMv8M/non_secure/portable/IAR/ARM_CM33` — the whole
directory for ARM_CM33 itself, file by file for the derivatives — so running the script
after the first commit produces the ARM_CM33 fix as well. It is included here and
**#1485 becomes redundant**.

Note there are two distinct CM33 assembly files in the tree, both affected, and they
are easy to confuse:

| path | role |
|---|---|
| `portable/IAR/ARM_CM33/non_secure/portasm.s` | generated copy |
| `portable/ARMv8M/non_secure/portable/IAR/ARM_CM33/portasm.s` | source of truth |

## Fix

The stray URL is deleted, and the comment keeps its original text and column alignment:

```diff
-        ldmia r2!, {r0, r3, lr}             http://files.iar.com/ftp/pub/box/bxarm-9.60.3.deb/* Read from stack - r0 = xSecureContext, r3 = PSPLIM and LR restored. */
+        ldmia r2!, {r0, r3, lr}             /* Read from stack - r0 = xSecureContext, r3 = PSPLIM and LR restored. */
```

This matches the unaffected sibling port `portable/IAR/ARM_CM23/non_secure/portasm.s`
(line 444), which carries the same comment text without any URL. Optionally the URL
could be kept *inside* the comment instead, as #1485 does — either form assembles; I
removed it because it is a meaningless installer link.

## Origin

Commit 78e0cc7 ("ARMv8.1-M: Add task dedicated PAC key support", #1195) introduced the
URL into CM33, CM35P, CM55, CM85 and the ARMv8M non-secure port. The CM52 (#1334) and
STAR-MC3 (#1363) ports were later created from an already-affected `portasm.s` and
inherited it.

## Verification

Assembled with `arm-none-eabi-gcc 13.3.1 20240614` (GNU Tools for STM32 13.3.rel1),
`-march=armv8.1-m.main -mthumb -mfpu=fpv5-d16`. IAR's own directives
(`PUBLIC` / `EXTERN` / `SECTION` / `THUMB` / `END`) were translated 1:1 to GNU
equivalents so that the only difference between the "before" and "after" inputs is the
one comment line, making the result attributable to it. The translated pair still
diffs in exactly that single line.

Baseline (known-good sibling, proves the harness accepts a healthy file):

```
portable/IAR/ARM_CM23/non_secure/portasm.s   ->  assembles OK
```

All seven files in this PR, over `configENABLE_MPU` / `configENABLE_PAC`:

| file | MPU=0, PAC=0 | MPU=1 |
|---|---|---|
| `portable/IAR/ARM_CM33/non_secure/portasm.s` | error → OK | not compiled either way |
| `portable/IAR/ARM_CM35P/non_secure/portasm.s` | error → OK | not compiled either way |
| `portable/IAR/ARM_CM52/non_secure/portasm.s` | error → OK | not compiled either way |
| `portable/IAR/ARM_CM55/non_secure/portasm.s` | error → OK | not compiled either way |
| `portable/IAR/ARM_CM85/non_secure/portasm.s` | error → OK | not compiled either way |
| `portable/IAR/ARM_STAR_MC3/non_secure/portasm.s` | error → OK | not compiled either way |
| `portable/ARMv8M/non_secure/portable/IAR/ARM_CM33/portasm.s` | error → OK | not compiled either way |

```
RESULT: 7/7 files fail before the change and assemble cleanly after it
        (configENABLE_MPU=0); 0 stray URLs remaining in the tree.
```

Caveats, stated plainly:

- This is **GNU as, not IAR `iccarm`** (no IAR licence available here). Both reject a
  bare token after an instruction, and the reporter of #1480 reproduced the failure on
  IAR ARM 9.70.2, so the defect and the fix are cross-confirmed — but this is not an
  IAR build.
- The PAC-enabled non-MPU combination could not be exercised: GNU as rejects the
  `PAC_KEY_*` `mrs`/`msr` operands on this target, before it ever reaches line 476.

## Note on CI

`.github/workflows/kernel-demos.yml` only builds the `CORTEX_MPS2_QEMU_IAR_GCC` demo,
which is compiled with GCC. The IAR ports are not built by CI at all, which is how this
slipped through. Adding an assembler-only syntax check for these ports would be a
separate, larger change — happy to look at it if you think it is worth doing.

Relates to #1480.


TomasGalbickaNXP (approved):

jefftenney:
Hi @sekior11, would you mind including that last file with these changes?  PR #1485 uses a different fix, but your fix is preferable.  You can simply execute portable/ARMv8M/copy_files.py and then commit the results.  It will be just that one file.  Then maintainers can simply close #1485 without merging it.

sekior11:
@jefftenney Done — added `portable/IAR/ARM_CM33/non_secure/portasm.s` in 73567c5, so the PR is now +7/-7 and all seven files report zero hits for `files.iar.com`.

You were right about the script. I checked what it actually does before committing, since I wanted to be sure it would not drag in unrelated churn: `copy_files.py` sources ARM_CM33 from `non_secure/portable/IAR/ARM_CM33` as a whole directory, while the derivative ports take `portasm.s` from that same place file by file. The other two files it copies into the directory, `mpu_wrappers_v2_asm.S` and `portmacro.h`, already have identical blob SHAs between source and destination, so re-running it changes only this one file. The blob I pushed has the same SHA as the source of truth (`7cb05a1`), as expected.

Re-verifying also turned up a mistake of mine that I have corrected. The original description said the stray URL "breaks every configuration of these ports". That is too strong: the line is outside the `configENABLE_PAC` block, but it sits inside the `#else` of `#if ( configENABLE_MPU == 1 )`, so it breaks **non-MPU** builds and MPU builds never compile it. The description is updated and the correction is recorded in the new commit message.

Thanks for the pointer — this makes #1485 unnecessary.

sekior11:
Hi @FreeRTOS/pr-bar-raisers — context for this PR: 7 IAR port files, one line each, removing a stray IAR
installer URL that the assembler parses as an operand of the preceding ldmia
(Error: garbage following instruction). It affects all non-MPU builds of the TrustZone
non-secure ports. CI has no IAR coverage, so I verified with a GNU-as equivalent harness
(7/7 fail before, 7/7 clean after) and documented that caveat in the PR body.
@jefftenney asked me to fold in the ARM_CM33 file — done in 73567c5. One approval is in;
per the repo process a CODEOWNER review is still needed.

